PR #1109, opened 2016-01-27
Made PagedList class refactoring

skoshelev:


AndreiMaz:
The last ctor implementation should not be changed. It's for already paged collections

PR #1121, opened 2016-01-28
Replaced RemoveByPattern logic to extension method

skoshelev:


skoshelev:
Хорошо

skoshelev:
Я все исправил

m1dst:
This is broken!

"Collection was modified; enumeration operation may not execute."

You can't iterate through a collection and delete things from it at the same time.

AndreiMaz:
Hi James,

Thanks a lot for reporting! Fixed

PR #1130, opened 2016-01-28
Fix for Replaced RemoveByPattern logic to extension method

m1dst:


AndreiMaz:
Hi James,

Thanks a lot! We've fixed a bit other way.as well

Issue #1092, opened 2016-01-21
All delete functionality should use "POST" (XSRF validation)


AndreiMaz:
1. Forum post or topic delete.
2. RemoveExternalAssociation

We should implement it similar to https://github.com/nopSolutions/nopCommerce/commit/ee71f7e068b86a202117309fda80f7b83c7f7322 and https://github.com/nopSolutions/nopCommerce/commit/0cf1dd5e7145da74d149a367f7493cb2a7e421f9

odises:
Please assign it to me. I will do it.

AndreiMaz:
odeses, thanks a lot for interest on contribution. Please create a new pull request as described at http://docs.nopcommerce.com/display/nc/Working+with+source+code+and+contributions

odises:
@AndreiMaz Ok, thanks, I read it. In the second paragraph "branch description and naming" I think this is "till January 2016" not 1016. I will create the pull request soon.

odises:
I checked all of the forms and realized that we send antiForgeryToken from client-side. I mean all of the views contains **Html.AntiForgeryToken()**
But no validation before related actions.
I should say antiForgery validation is necessary not just for delete action but for all the actions with post verb. I fixed them all, and made a pull request.

odises:
Can I have a question? This issue mean all delete functionality in public area? Can you make it more clear for me?

AndreiMaz:
Sure. Please see the issue description:

1. Forum post or topic delete.
2. RemoveExternalAssociation


We should implement it similar to https://github.com/nopSolutions/nopCommerce/commit/ee71f7e068b86a202117309fda80f7b83c7f7322 and https://github.com/nopSolutions/nopCommerce/commit/0cf1dd5e7145da74d149a367f7493cb2a7e421f9

PR #1131, opened 2016-01-28
Issue #1092 XSRF validation added before all actions with POST verb

odises:


odises:
Yeah, sorry. You'r right. I made a mistake.

PR #1146, opened 2016-02-01
Removed switch from IsStaticResource method

skoshelev:


skoshelev:
The formatting did not change, changed only the number of rows and their order.

Issue #1143, opened 2016-01-31
When we create a user in admin, ensure that email is entered if "registered" customer role is checked


AndreiMaz:
Otherwise, it'll be possible to created registered customer with "empty" email

manojkulkarni30:
Please assign it to me. I will do it.

AndreiMaz:
Sure. please create a pull request

PR #1150, opened 2016-02-01
#1143 Ensured that email address is not empty if registered role is checked

manojkulkarni30:
Ensured that email address is not empty if registered role is checked to avoid customers registered with empty email address


AndreiMaz:
Thanks. We we should use "CommonHelper.IsValidEmail" to validate entered email in this case.

manojkulkarni30:
Thank you for the feedback. I have updated the code. Please take a look.

AndreiMaz:
Thanks. But there's no need to separately compare with null or empty. I mean the following block:

if (String.IsNullOrWhiteSpace(model.Email))
                {
                    ModelState.AddModelError("", "Email is required for customer to be in 'Registered' role");
                    ErrorNotification("Email is required for customer to be in 'Registered' role", false);
                }


One check with CommonHelper.IsValidEmail(model.Email) is enough.

It can all be placed in the same "if" condition validation:

if (newCustomerRoles.Count > 0 && newCustomerRoles.FirstOrDefault(c => c.SystemName == SystemCustomerRoleNames.Registered) != null & !CommonHelper.IsValidEmail(model.Email))


P.S. Could you please clarify a new pull request with one commit with all these changes?

manojkulkarni30:
Thank you for the feedback. I just created a new pull request with all the changes. Thank you for all your help and suggestions.

PR #1157, opened 2016-02-02
Feature google analytics: allow store owner to choose widgetzone

rudgr:


rudgr:
ok cool! :)

PS: I got sample widgetzone code from nopcommerce livepersonchat plugin (http://www.nopcommerce.com/p/981/liveperson-chat-widget.aspx). I guess the confusion started when  livepersonchat plugin was removed from main nopcommerce source solution? " "Admin.ContentManagement.Widgets.ChooseZone" seems a useful general resource string from which multiple plugins could benefit.

PR #1173, opened 2016-02-07
AutoMapper usage changed from static to dynamic.

ArturAlekseev:
AutoMapper initialization is in AutoMapperRegistrar.cs.
It is registered for autofac in DependencyRegistrar.cs.
AutoMapperStartupTask.cs should be removed (its content is moved to AutoMapperRegistrar).


AndreiMaz:
Thanks! But we should not mix tasks in pull requests (you have two distinct commits to two tasks). A new branch should be created for each new task

Issue #1142, opened 2016-01-30
Search functionality in Admin Area


AndreiMaz:
It's would be great if next version of NopCommerce has the feature that let Admin search function in Admin Area. Example if Admin want to "add product" so just type in textbox word "Add Product" (similar to function name) then it will redirect to page add product.

Please take a look at cPanel Admin Page URL below.

http://postimg.org/image/uq7uw3vc5/

Source: http://www.nopcommerce.com/boards/t/40320/search-functionality-in-admin-area-page.aspx
![aaa](https://cloud.githubusercontent.com/assets/5019111/12697034/9c80f67c-c789-11e5-8d5b-06eaabd392c8.png)

PR #1180, opened 2016-02-08
#1142 Search functionality in Admin Area

odises:
- Search functionality added to admin area


AndreiMaz:
Looks very interesting. A couple of notes:
1. Localize "Search" text
2. Is it possible to disable (ctrl +s) hotkey? I think it should be left as default browser hotkey for saving
3. It's better to move the search box to the top (header). Now it confuses a bit when rendered near other search boxes

Issue #1141, opened 2016-01-30
Fix AutoMapper warnings


AndreiMaz:
Mapper.CreateMap<TSource, TDestination>()' is obsolete: 'The static API will be removed in version 5.0. Use a MapperConfiguration instance and store statically as needed. Use CreateMapper to create a mapper instance

ArturAlekseev:
Maybe this way:
https://github.com/nopSolutions/nopCommerce/pull/1173

PR #1182, opened 2016-02-08
AutoMapper: change usage to dynamic

ArturAlekseev:
Related to [https://github.com/nopSolutions/nopCommerce/issues/1141](url)


AndreiMaz:
Thanks! But constructor of AutoMapperRegistrar in invoked multiple times for each HTTP request (in admin area). It's not acceptable. It should be run only once (on application startup). It's better to make this class static (there's no need to register it in DependencyRegistrar.cs)

ArturAlekseev:
Maybe i'm not very good in C#.
Static class can contain only methods (no objects). And we must share objects of IMapper and MapperConfiguration.
If to make static class that return these objects from its methods then still methods are executed for each request.
I'v made the same solution but class is registered in dependecyregistrar with .SingleInstance(); so it runs only once.

AndreiMaz:
Hi Artur,

Please have a look at the latest commit (how it's already done)

PR #1185, opened 2016-02-08
AutoMapper: change usage to dynamic

ArturAlekseev:
This time init class renamed to AutoMapperInit.
It is registered in DependecyRegistrar with SingleInstance(). So its initialization is executed only once.


AndreiMaz:
Thanks. Already done

PR #1241, opened 2016-02-25
Added dynamic detection of plugin view locations

cybernexus:
I originally wrote this code inside a custom view engine but it seems like it would work and play nicely with the core framework / default view engine. Upon installation of a plugin w/ restart it will add that plugins folder location with standard conventions to the list of locations.

### DIFF
--- a/src/Presentation/Nop.Web.Framework/Controllers/ThemeablePluginViewLocationAttribute.cs
+++ b/src/Presentation/Nop.Web.Framework/Controllers/ThemeablePluginViewLocationAttribute.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace Nop.Web.Framework.Controllers
+{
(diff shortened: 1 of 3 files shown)


cybernexus:
Found a bug in the initial implementation caused by two different plugins having the same controller and action names. Now there is only one path added to location arrays, which contains a tokenized string that is replaced at run time based on the controller context and a class attribute.

This change also gives developers the ability to dynamically change the default location of their views for a particular controller. 

The only drawback to this implementation is that it causes a default view location of "{DynamicPluginPath}/{1}/{0}.cshtml" to be added. I'm not sure if this would cause an issue or not, I haven't found anything in my tests so far to indicate an issue.

cybernexus:
Thought this would work but I ended up finding a rather large problem that I does not happen when used from my plugins implementation. Cancelling Pull request.

PR #1243, opened 2016-02-25
Top Menu first time load speed up

ArturAlekseev:
1 function added to category service + it is used in CatalogController
Else is EntityFramework added into Services.dll to use SqlQuery.
Just the idea how and what to do.
For example if there are 20000 categories. Then all of them are loaded from db. Then all loaded are parsed in cshtml file for includeintopmenu flag (and parsed again and again with each request). It's long. We should load from db only needed categories.

p.s. nearly same problem is exist for category navigation. All categories are loaded (and cached) first time (and it's around 5-7 seconds for 20000 categories). Then cache is used for else (these requests are faster < 1 second). As I see cache is cleared even if application pool is not recycled. It makes all categories reload after a while.

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/CategoryService.cs
+++ b/src/Libraries/Nop.Services/Catalog/CategoryService.cs
@@ -11,6 +11,8 @@ using Nop.Services.Customers;
 using Nop.Services.Events;
 using Nop.Services.Security;
 using Nop.Services.Stores;
+using Nop.Data;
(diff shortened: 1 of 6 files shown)


ArturAlekseev:
Forgot to add check for published = true. So this example will load even hidden categories.

RomanovM:
Thanks for your contribution. But right now we don't plan to do it.

PR #1248, opened 2016-02-27
Select role in registration form

cesaremarasco:
Hi all,
In this release it is possible enable role selection during customer registration. Administration must enable first this feauture in customer settings. In attach a small document of specification
[Specification.docx](https://github.com/nopSolutions/nopCommerce/files/149308/Specification.docx)


RomanovM:
Thanks for your contribution. But right now we don't plan to do it out of the box.

Issue #988, opened 2016-01-19
Support SSL on NewsFeeds


mariannk:
On Nop.Web/Controller/BlogController Modiy ListRss method to set the protocol as either http or https depending on the how the site is being run. 

Replace 
string blogPostUrl = Url.RouteUrl("BlogPost", new { SeName = blogPost.GetSeName(blogPost.LanguageId, ensureTwoPublishedLanguages: false) }, "http"); 

for: 
string blogPostUrl = Url.RouteUrl("BlogPost", new { SeName = blogPost.GetSeName(blogPost.LanguageId, ensureTwoPublishedLanguages: false) }, _webHelper.IsCurrentConnectionSecured() ? "https" : "http");

It should also be done for other RSS methods (e.g. NewsController, etc)

PR #1258, opened 2016-02-29
Add check for secure Connection #988

schulz3000:
Implemented like suggested


AndreiMaz:
Thanks. But it should also be done for other RSS methods (e.g. NewsController, etc)

schulz3000:
done

PR #1273, opened 2016-03-04
Replace hard coded value with Int32.MaxValue

SamPlusPlus:


AndreiMaz:
Thanks a lot. But please replace it with "int.MaxValue" (not Int32.MaxValue) to keep it consistent with other classes

SamPlusPlus:
@AndreiMaz changed.

PR #1288, opened 2016-03-12
Category navigation speed up

ArturAlekseev:
Load only categories that are needed in tree generation. Especially useful when there are a lot of categories in database.


src/Presentation/Nop.Web/Controllers/CatalogController.cs
```diff
@@ -649,12 +649,12 @@ public ActionResult CategoryNavigation(int currentCategoryId, int currentProduct
                 _workContext.WorkingLanguage.Id,
```
arieldeil: If you are not caching the result at all then you should remove "var cacheKey" because it has no use anymore

MaxMommersteeg:
Could someone inform me about the status of this PR? Having +- 1000 categories in my database en the category page is very slow. Would this code be of any use? Will it be added to the original source?

RomanovM:
Thanks for your contribution. But right now we don't plan to do it.

PR #1291, opened 2016-03-15
Fixed messed up project reference

sethmoore:
Fixed some references that prevented the solution from building properly.


SamPlusPlus:
+1

PR #1295, opened 2016-03-16
encapsulate Streams with using

schulz3000:
small refactoring to make use of "using" for correct disposing


AndreiMaz:
Thanks. But the "fix" in EmailSender breaks the functionality (attach downloads to message templates). The stream should not be closed.

PR #1306, opened 2016-03-20
Allow setting colors to predefined product attribute values

rsheptolut:
Before that you needed to pick colors for each product by hand. Predefining colors is useful for example for paint or textile: there are vendor-defined palettes from which a customer should be able to choose a color.


RomanovM:

Issue #1355, opened 2016-04-20
"Buy Button" should be re-enabled after canceling the order contains stock managed product


odises:
- Assume **Stock quantity** is set to **10** for a product, and **Minimum stock availablity** is **0**, and also **Low stock activity** is set to **Disable buy button**.
- A user buys all the **10**.
- **Reserved quantity** automatically will be set to **10**.
- And then **Add to Cart** button becomes disabled.
- Admin cancels the order.
- Reserved quantity returns to **0**.
- But the product's **Add to Cart** button is still unavailble. (event after cleaning the cache or restarting the application)

odises:
After reporting this issue I noticed that there is an advanced setting for this to republish product (PublishBackProductWhenCancellingOrders). But I believe it's hard to find this setting for users. So I made a pull request and put this in Catalog setting.

AndreiMaz:
thanks.

PR #1363, opened 2016-04-21
#1355 Buy Button should be re-enabled after cancelling orders

odises:
It turns out that there is a setting _(PublishBackProductWhenCancellingOrders)_ that only appears in **All Settings (Advanced)** section. I believe that it's not very easy to find for users, so I put this setting in **Settings/Catalog Settings** section.


RomanovM:
Thanks for your contribution. But we are planning to leave it in the advanced settings

RomanovM:
Related task https://github.com/nopSolutions/nopCommerce/issues/1355

PR #1366, opened 2016-04-21
Remove extra parentheses

SamPlusPlus:


SamPlusPlus:
Actually this is done everywhere in the checkout controller. This cleanup should happen everywhere. Better yet this could actually be refactored into one method.

Issue #159, opened 2016-01-15
Use output caching


mariannk:
Currently we use model caching mostly. Should we use output caching as well?

In this case we should also create caching keys based on appropriate workflow (current store, language, etc)

We not forget about Azure support (multiple instances). So we should use our caching managers (and not a default ASP.NET caching infrastructure for output caching)

keremdemirer:
Will we integrate output caching with current ModelCacheEventConsumers? 

Microsoft has a library for caching to redis. [link](https://azure.microsoft.com/en-us/documentation/articles/cache-asp.net-output-cache-provider/). It's parallel to one we use for redis session. Supports azure out of the box.

[This ](http://antix.co.uk/Blog/IfModifiedAttribute) looks promising if we will code our own. 

Another thing we should consider is customer related parts of the pages are cached properly. Header links, recently viewed products, shopping cart, acl related entities etc... In our custom implementation we moved these parts to separate ajax requests and cache the common parts of the page. Also enabled  `catalogsettings.ignoreacl`. We are very satisfied with the performance but this is not the best implementation. There is a library called MvcDonutCaching presumes (by its name:)) enabling donut hole caching. I have not tested it, and it seems inactive. But we can analyze their code.

Enabling this feature will decrease page load times more than %30 in my experiences, which is huge for any website.

sc0ttc0:
I think it would definitely be worth looking at output caching .. even if there were some caveats around initial implementation e.g.
- guest users only
- you have nothing in the cart 
- you cannot used recently viewed products.

- in that sense definitely worth making the cart on each page and recently viewed an ajax load.  (often wondered why these are not just plugins that ship with nop - they would lend themselves to widgets - performance ??)  You could fudge it a bit and just implement ajax that returns the cart html as is so that it isn't in the initial page load?

Alternative thought - some way to flag that the user has no basket items (cookie?) and if recent items is off you could use an Nginx or similar on the front end to cache pages.

PR #1368, opened 2016-04-21
output cache attribute implementation.

keremdemirer:
https://github.com/nopSolutions/nopCommerce/issues/159

I only added attribute to Home/Index action. 
ps. logic taken from Orchard Project.


keremdemirer:
found some bugs in the code and closing this request.

Issue #1379, opened 2016-05-03
How can I limit payment method when I choose specific shipping method?


kenny2061:
Hi, I come from Taiwan.
In my country, we have some different payment method.

For example, I can send product from convenience store A to B AND collection money(product price and shipping fee), so I want to show or hide the payment method when I choose specific shopping method.

Similarly, If I choose "pickup at store", that maybe mean "hey, I will go to your store buy this product and pay by cash at your store, please keep this product for me", so it is no need to select shipping and payment method.

So, I want to know how can I control the shipping and payment method display on the checkout process.

PR #1383, opened 2016-05-05
Shipping plugin should be tell checkout to limited to show payment method

kenny2061:
I post #1379 issue and this is my first time contribute project on github.
I think shipping plugin sometimes would be relation other payment method.
In my case, I need to send the product to customer and collect money by carrier, so if customer select this shipping method, checkout should be hide other payment method.


RomanovM:
Thanks for your contribution. But right now we don't plan to implement it. BTW, we already have a HidePaymentMethod() method https://github.com/nopSolutions/nopCommerce/blob/develop/src/Libraries/Nop.Services/Payments/IPaymentMethod.cs#L33  which can be used to hide the appropriate payment under the necessary conditions

PR #1430, opened 2016-05-25
Option to wrap DatePickerDropDowns selects with span tags for styling

lfurzewaddock:
Useful for Bootstrap and other grid layout based CSS frameworks.


AndreiMaz:
@lfurzewaddock Thanks a lot!

PR #1431, opened 2016-05-26
Updated PayPal Payment Processor to use TLS 1.2

ChrisRowtcliff:
Added code to force the connection to PayPal uses TLS 1.2 which fixes the error:
'Invalid HTTP response The request was aborted: Could not create SSL/TLS secure channel.' when trying to access the PayPal API.

See https://www.paypal-knowledge.com/infocenter/index?page=content&widgetview=true&id=FAQ1914&viewlocale=en_US


AndreiMaz:
Thanks. But it's already fixed in the latest version

Issue #1420, opened 2016-05-17
 "My Product Reviews" in "my account" (contribution)


AndreiMaz:
Another minor enhancement I'd like to see would be to add a new navigation item in the customer account for "My Product Reviews." 

It is important now because it is possible for customers to earn Reward Points for giving product reviews. Customers need an easy way to see a list of which products they have already reviewed.

This view would display a list of all reviews created by this customer. The list should include columns with Product name, Date Reviewed, and Current Approval Status (Approved/Disapproved). Maybe also include paging for list..some admins have many products.

Contirbution (see whether it's good and could be merged): https://github.com/nopSolutions/nopCommerce/pull/1440

DaveHogan:
Is this up-for-grabs @AndreiMaz

AndreiMaz:
@DaveHogan hi. yeah. this work item is not assigned yet

DaveHogan:
@AndreiMaz - assign to me. I'll be happy to pick this up.

AndreiMaz:
@DaveHogan please create a fork. If it's good, then we'll merge it. thanks

PR #1440, opened 2016-05-27
Inital concept of a my reviews section

DaveHogan:
Suggested implementation for [Issue 1420](https://github.com/nopSolutions/nopCommerce/issues/1420)

First time I've worked against the nopCommernce codebase, so criticism and guidance is certainly welcome.

Few points
 * Created as a separate controller, wasn't sure if that is that describable?
 * Paging hard coded to 20.
 * ApprovalStatus is only a Boolean Yes No - not (Approved/Disapproved/Pending)
therefore I've assumed true = Approved, false = Pending.
 * No way to turn off reviews? Should this be a global store setting? Or a customer preference?


AndreiMaz:
Thanks a lot, Dave. Merged.

PR #1441, opened 2016-05-29
Allows you to set up any currency as the main exchange rates, not fixed to euro

chunlizh:
Allows you to set up any currency as the main exchange rates, not fixed to euro


AndreiMaz:
thanks. but we don't need AliPay available out of the box. It's available as third-party plugin on our extensions page

PR #1461, opened 2016-06-06
Nivo Slider widget remove inline jQuery

lfurzewaddock:
Performance optimisation and code organisation.


RomanovM:
Thanks for your contribution. But this should be done for all such scripts. We opened new issue https://github.com/nopSolutions/nopCommerce/issues/2682

PR #1471, opened 2016-06-07
Add miniprofiler for sql queries

schulz3000:
With Miniprofiler for EF6 we can detect a lot of duplicated and slow queries more easily


src/Presentation/Nop.Web/Global.asax.cs
```diff
@@ -46,6 +47,9 @@ protected void Application_Start()
             //disable "X-AspNetMvc-Version" header name
             MvcHandler.DisableMvcResponseHeader = true;
 
+            //MiniProfiler for EF must be initialized before first Database call. Doesn't matter if we later start Profiler or not.
+            MiniProfilerEF6.Initialize();
```
AndreiMaz: thanks. but this code should be put into
if (databaseInstalled)
            {
                if (EngineContext.Current.Resolve<StoreInformationSettings>().DisplayMiniProfilerInPublicStore) {

schulz3000: That's not possible.
You have to initialize EF6 Support for Miniprofiler before first databasecall.
At the point you mentioned database was queried before.

AndreiMaz: That's bad. In this case MiniProfiler will always be used. Even in production environments where it's not required. Hence it can affect performance

RomanovM:
Thanks for your contribution. But right now we don't plan to implement it out of the box.

PR #1479, opened 2016-06-09
Performance. Add support for "async" for javascript files

agametov:
Add support "async" attribute when are not bundled and bundling is disabled, added new bool parameter **isAsync**


AndreiMaz:
thanks a lot, Rustam!

PR #1490, opened 2016-06-14
Separated social links out into separate model and view to allow them…

MarkSizer:
… to be shown in places other than just the footer. As per a customer request I have received.


AndreiMaz:
thanks a lot!

PR #1545, opened 2016-06-27
Added assigned to store discounts

shurub3l:
It is difficult to add discounts to all products in a store, even with the "restricted to store" discount requirment rule plugin, especially if you have many products and categories, that change often. When you add new products or categories (depeneding on the setting of the discount - assigned to products or assigned to categories) you have to remember to add the store discount to that product/category too.

So I added assigned to store discount type, similar to assigned to products, categories, or manufacturers, which applies discount to all products in a certain store.


RomanovM:

PR #1559, opened 2016-07-07
Refactoring LINQ queries

LexZhukov:
Refactoring LINQ queries


AndreiMaz:
Thanks a lot for contribution. But we would prefer to stay with the current approach.

1. GetAllUrlRecordsCached - method "GetBySlug" and "GetBySlugCached" - ".ThenByDescending(ur => ur.Id);" should be ".ThenBy(ur => ur.Id);"

Issue #1456, opened 2016-06-02
Double click at confirm button during checkout


AndreiMaz:
Not sure whether anyone notice that, when accidentally click twice at the confirm button, it will direct you to a empty cart screen, but I was expecting a payment screen.

This may be a minor bug, but might affect the user experience.

Source: http://www.nopcommerce.com/boards/t/42164/double-click-at-confirm-button-during-checkout.aspx

also http://stackoverflow.com/questions/8757963/avoiding-duplicate-form-submission-in-asp-net-mvc-by-clicking-submit-twice

Xzelsius:
A similar bug exists for customer registration.
The same script should work there too.

PR #1572, opened 2016-07-14
#1456 Prevent double click on confirm.

Xzelsius:
Added a script to prevent double click on checkout confirm.

### DIFF
--- a/src/Presentation/Nop.Web/Views/Checkout/Confirm.cshtml
+++ b/src/Presentation/Nop.Web/Views/Checkout/Confirm.cshtml
@@ -73,6 +73,14 @@
                     <div class="buttons">
                         <input type="submit" name="nextstep" value="@T("Checkout.ConfirmButton")" class="button-1 confirm-order-next-step-button" />
                     </div>
+					<script>
+                        $(document).on('submit', 'form', function() {
+                            var button = $(this).find('input[type="submit"]');
+                            setTimeout(function() {
+                                button.attr('disabled', 'disabled');
+                            }, 0);
+                        });
+                    </script>
                 }
                 if (Model.Warnings.Count > 0)
                 {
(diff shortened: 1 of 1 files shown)


src/Presentation/Nop.Web/Views/Checkout/Confirm.cshtml
```diff
@@ -73,6 +73,14 @@
                     <div class="buttons">
                         <input type="submit" name="nextstep" value="@T("Checkout.ConfirmButton")" class="button-1 confirm-order-next-step-button" />
                     </div>
+					<script>
+                        $(document).on('submit', 'form', function() {
+                            var button = $(this).find('input[type="submit"]');
+                            setTimeout(function() {
```
AndreiMaz: But what will happen if some error happens? It won't be enabled back. So it can be clicked again. For example, payment error or "MinimumOrderPlacementInterval" setting requirement is not met

Xzelsius: **IsMinimumOrderPlacementIntervalValid** in ConfirmOrder() throws an exception and to display that exception the page reloads. -> The initial state of the button stays enabled.

And on the other side, OpcConfirm is not affected by this.

AndreiMaz: sure. you're right. It should be good. For multi-step checkout.

But it misses one-page checkout page. This is what I meant in the previous comment (one-page checkout)

I would also use a certain button ID selector (some validation). This way not other submit buttons are disabled on the page.

Xzelsius:
So i changed the selector to use the order form and the next-step button.

--

One-page-checkout isn't affected by this bug. Since the ConfirmOrder.save() script in public.onepagecheckout.js allready checks if there's an ajax call running.

AndreiMaz:
Merged! Thanks a lot for this contribution!

PR #1574, opened 2016-07-14
Return the object in InsertCategory, InsertProduct, InsertManufacturer methods

erkerkan:
The object will usually be needed right after method call, so probably it's better to return it.


AndreiMaz:
But why should it be done? The already passed entity (as parameter) contain this object

PR #1596, opened 2016-07-22
Feature 1566 discountrule hasorderedxtimes

orrTechnologies:
Plugin will allow you to use promotions based on how many times a customer has ordered IE Free shipping for first order only.


src/NopCommerce.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
```
odises: Why did you change the sln file?!!

orrTechnologies: Ugh must of missed it in my commit

AndreiMaz:
Thanks a lot. But we do not add new plugins to the core. Could you please create a new plguin and upload it to http://www.nopcommerce.com/extensions-and-themes.aspx? as described at http://www.nopcommerce.com/contribute.aspx

PR #1627, opened 2016-07-29
Email validator does not accept 'cs_test@cs_test.com' as a valid email address.

ilich:
Hello,

I have a customer who wants to use 'cs_test@cs_test.com' as his email address. It seems to be a valid email address.

I have updated the code to use Microsoft email validator to validate emails. https://msdn.microsoft.com/en-us/library/01escwtf(v=vs.110).aspx

Please let me know if you have any question.

Sincerely,
Ilya


AndreiMaz:
But it's not a valid email - http://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it

ilich:
Unfortunately we had a client who reportd this problem. I think they had underscore only in the username, not in the domain name. It might be just a test email account they use for testing. We didn't go deep into details at that time.

I think you can reject the pul request based on the link from stack overflow.

AndreiMaz:
thanks anyway for your contribution!

Issue #107, opened 2016-01-15
Discounts by vendor


mariannk:
It would be great when using the multi-vendor feature in nopCommerce if a store owner could set whether or not Vendors could create their own discounts. The discounts should only be applied to the creating vendors products thoug

Source: http://www.nopcommerce.com/boards/t/35617/feature-request-allow-vendors-to-setup-their-own-discounts.aspx

---

Ability to assign a discount based on a vendor. 

Should vendors be allow to create discount out of the box? I think no. But they should be allowed to apply discounts on a product details page (discounts tab)
See http://www.nopcommerce.com/boards/t/32312/discount-feature-request-add-discount-per-manufacturer-or-vendor.aspx

---

Contribution: https://github.com/nopSolutions/nopCommerce/pull/1643

SilhouetteBS:
Please add this feature. You can create a discount based on just about everything else, but vendors are left out.

AndreiMaz:
Also see https://github.com/nopSolutions/nopCommerce/issues/1972

AndreiMaz:
After some consideration we've decided not to implement this functionality out of the box and leaving it for customization

PR #1643, opened 2016-08-03
added vendor discount management code

hitinder:
Made code changes to allow vendors to view only their own discounts.
Administrator can view all discount types, and filter discounts based on vendor name.
Administrator can create new discount and assign vendor to it.


AndreiMaz:
Hi,

So how is this property used (when assigned to a vendor)?

saYRam:
Excuse me. Is the Vendor Discount implemented?

RomanovM:
Related task https://github.com/nopSolutions/nopCommerce/issues/107

RomanovM:

PR #1655, opened 2016-08-05
Simplify the mapping of Security Settings.

m1dst:
I used AutoMapper to simplify the code and remove the need for manual mapping.

Other settings code could easily be reworked too but I didn't want to waste time in case you rejected the entire principle.  I was making changes to Security Settings anyway.

I've also looked at the ability to do the same for the others where SaveSettingOverridablePerStore is called but I struggled with Reflection.  Got very close but gave up.


AndreiMaz:
Thanks a lot, James! We'll have a look - https://github.com/nopSolutions/nopCommerce/issues/1656

AndreiMaz:
Thanks a lot! Done. But with other refactoring

PR #1657, opened 2016-08-05
Enable TLS 1.2 support.

ilich:
Most of the API providers require TLS 1.2 to call their methods. The change enables TLS 1.2 in nopCommerce itself instead of adding this support to every single plugin.


AndreiMaz:
Thanks a lot! Done with some minor refactoring (removed duplicated code from plugins)

Issue #295, opened 2016-01-18
Use more secure password hashing algorithm. Other security enhacements


mariannk:
With the recent password hash leak by LinkedIn/eHarmony, a lot of people are talking about password hashing algorithms. NopCommerce uses SHA-1 with a 5 character salt that is unique to each customer. That is much better than what LinkedIn was doing (MD5 with no salt) but it looks like even SHA-1 with salts can be brute forced in a reasonable time frame. (days) 

Is there any thought of offering Bcrypt or PBKDF2? I know PBKDF2 is built in to the .net framework (Stackoverflow switched from Bcrypt to PBKDF2). The benefit of these is that it takes .3 seconds to generate a hash versus milliseconds for MD5/SHA-1. That difference isn't a big deal when you're logging in but it greatly increases the time need to crack a hash. It goes from days to thousands of years. 

I was going to ask how one would handle switching hashing methods when customers are already registered and then saw that each customer has a PasswordFormatId. Awesome! 

Forum source: http://www.nopcommerce.com/boards/t/17505/password-hashing-algorithms.aspx 

[AM] It's already supported. Just set to CustomerSettings.HashedPasswordFormat to any new supported value - https://msdn.microsoft.com/ru-ru/library/wet69s13(v=vs.110).aspx

---

There has been some talk in the past about improving hash and encryption methods used in nopCommerce.  Over the past few days, I have worked on a solution that will work with the next version of nop, 3.90.
You can see the commit here:
https://github.com/f1ana/nopCommerce/commit/85ee625f2059519e7a5e0251d9d7484af2d356a9

I made the focus of this on new installations of nop since there was a pretty hard dependency on everything using sha1/3des out of the gate.  More so, this would likely break plugins that have existing data encrypted or hashed, so too many unknowns in my opinion to backport this.

Here's a recap of the changes:
- Stronger default hashes
- Stronger default encryption key
- Select hash and encryption at install time
- Add more unit tests for hash and encryption
- NopDropdownListFor support on installation page (kind of)

The reasons why I made these changes:
SHA1 is on the way out and we should be using some form of SHA2, either 256 or 512.
Unless I'm mistaken, 3DES is only used in 112 bit mode with nop since the keysize reports itself to be 128 bits.
There was a hard dependency in the system for 3DES.  The encryption service now mostly works seamlessly with any algorithm that inherits from the SymmetricAlgorithm class.
[...]

PR #1664, opened 2016-08-10
Feature/encryption improvements

f1ana:
- Stronger default hashes - DONE
- Stronger default encryption key
- Add more unit tests for hash and encryption
- NopDropdownListFor support on installation page (kind of)

If we want to allow changes in production, then we have to store hash algorith for each record in "CustomerPassword" table. Because existing records wwould be broken


f1ana:
Hello,

Just checking in on the status of this pull request.  If any other changes need to be made prior to accepting, please let me know.

-Chris

AndreiMaz:
Thanks a lot! We'll consider it a bit later. This work item (https://github.com/nopSolutions/nopCommerce/issues/295) is still open

src/Libraries/Nop.Core/Domain/Security/HashFormat.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nop.Core.Domain.Security {
+    public enum HashFormat {
+        Sha1 = 0,
```
ibcooley: I'm not a maintainer here, but why add support for SHA1?  It has been broken in practice, in some cases, and is not being used by governments any more.   I'd suggest removing this from the set of potential hash formats.

src/Libraries/Nop.Core/Domain/Security/EncryptionFormat.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nop.Core.Domain.Security {
+    public enum EncryptionFormat {
+        TripleDes = 0,
```
ibcooley: I'm not a maintainer here, but why add support for TripleDes?  Isn't it much less secure than AES?

PR #1671, opened 2016-08-14
Fix for bundling CSS under a virtual directrory (fixes relative urls i…

shoecake:
This is a fix for websites under a virtual directory. there is a bug in the Microsoft.AspNet.Web.Optimization project where the code does not take into account the application path when transforming relative links.

e.g. a css line of:
background: url('../img/rss-icon.png');

gets transformed to /img/rss-icon.png when it should get transformed to /virtualdir/img/rss-icon.png

It does not look like that code is being maintained as this fix has been available for some time. The original fix I took from https://github.com/benmccallum/AspNetBundling/blob/master/AspNetBundling/CssRewriteUrlTransformFixed.cs

Note, you may want to move the CssRewriteUrlTransformFixed.cs file and its namespace to somewhere more suitable.


RomanovM:
Thanks for your contribution. But after migration to the net core, this pull request is no longer relevant

Issue #1558, opened 2016-07-07
Admin area. Add warning notification


dchmir:
Already have support for error and success messages. It is necessary to add support for warnings (yellow)

also see contribution: https://github.com/nopSolutions/nopCommerce/pull/1684

PR #1684, opened 2016-08-16
Fix issue 1558 - Admin area. Add warning notification

MaxMommersteeg:
Implemented support for warning notifications. For both front- and back-end.

See issue:
https://github.com/nopSolutions/nopCommerce/issues/1558


mariannk:
Thank you for the contribution!

Issue #1717, opened 2016-09-01
Terms of services. Checkbox labels should be clickable (contribution)


Andrey-Pavlov:
For example, on checkout page click on label doesn't select a checkbox:
![label checkbox](https://cloud.githubusercontent.com/assets/7976740/18159737/c27949fe-7032-11e6-9b28-d01743c2d921.png)

See pull request - https://github.com/nopSolutions/nopCommerce/pull/1725

PR #1725, opened 2016-09-02
Fix: Terms of services. Checkbox labels should be clickable #1717

RoboBurned:
Replace span with label with "for" attribute.


AndreiMaz:
thanks a lot

Issue #1758, opened 2016-09-20
Disallow /admin from robots.txt


AndreiMaz:
And all other admin pages

http://www.nopcommerce.com/boards/t/44272/shouldnt-the-admin-directory-be-disallowed-from-the-robotstxt.aspx

PR #1761, opened 2016-09-22
#1758 Disallow /admin from robots.txt

Xzelsius:
Added disallow rules to robots.txt.

/admin
/admin/

First is to block crawlers from "foo.bar/admin"
Second is to block crawlers from "foo.bar/admin/{controller}"

It requires both, otherwise crawlers still try to open "foo.bar/admin"


AndreiMaz:
Thanks a lot! I've just done it (as one commit). I've also added it in a bit other order (we have all names sorted alphabetically)

PR #1762, opened 2016-09-22
Enable overriding settings file name in web.config. Only read Settings once.

kwestground:
Overriding settings filename is useful when using Azure App Sites continuous deployment (from git).
You can in that case have Settings_dev.txt for local development connection string and Settings.txt for production connection string by setting "NopCommerce.SettingsFile" in Azure App Settings.
Also contains fix for loading settings file only once (instead of ~15??).


RomanovM:

PR #1764, opened 2016-09-23
Set LastSuccessUtc in Task constructor

rossiter10:
Set LastSuccessUtc property to the database value in the constructor as well. Without it, if the task throws an exception, the LastSuccessUtc column in the database will be overwritten with NULL since the property is never initialized.


AndreiMaz:
Thanks a lot!

PR #1815, opened 2016-10-18
Prevent shopping carts unintentionally migrating between Customers

sgmelin9:
http://www.nopcommerce.com/boards/t/37151/wish-list-disapearing-from-customer-moving-to-another-customer.aspx#177254


RomanovM:

PR #1830, opened 2016-10-28
Remove localized topic's properties when topic has been removed.

ilich:
I found out that localized properties are not removed when I remove a topic. I fixed the problem in     commit 406cd13.


RomanovM:

PR #1844, opened 2016-11-04
Removed hard-coded nivo slider transitiion effect

MarkSizer:
Nivo Slider includes a settings class that allows you to specify the transition type. The view used to render the slider however had a transition hard coded for the third slide, which over-wrote the preferred value.


AndreiMaz:
Thanks! But it's by design to demonstrate different effects

PR #1854, opened 2016-11-09
Disallow users to impersonate as guests

mcliment:
If users impersonate as guests they enter in a loop that prevents them from logging in and finish the impersonation.


AndreiMaz:
thanks. created a work item - https://github.com/nopSolutions/nopCommerce/issues/1855

AndreiMaz:
@mcliment could you please provide a list of steps to reproduce the issue? It works fine. If you impersonate as guest, then you can always click "finish impersonation" button

mcliment:
@AndreiMaz Hi, you're absolutely right. The problem was that in the template the link was not showing up properly. Thanks for the tip and sorry for the inconvenience.

PR #1881, opened 2016-11-18
Show more detailed information of loaded assemblies

mcliment:
As explained here: http://www.nopcommerce.com/boards/t/45058/show-more-assembly-information-in-the-system-information-page.aspx#178867

* Show server variables and assemblies in a proper list (with bullets)
* Show server variable name in bold letters
* Hide server variables with raw HTTP header information that is repeated below
* Show assembly name in bold letters
* Show if assembly is in debug/release mode
* If in full trust, show assembly location (is commented out but CommonHelper has a method to get Trust)
* If in full trust, show assembly build date (we find this information useful when diagnosing problems)


AndreiMaz:
Thanks a lot!

mcliment:
You're welcome 👍

PR #1915, opened 2016-12-01
Initial changes + adding vendor name to category url

asultan80:


AndreiMaz:
What does it do?

AndreiMaz:
Thanks. but we do not plan these changes in nopCommerce

PR #1947, opened 2016-12-13
return empty content if there's noly one currency in database

JohnsonYuan:
I think you missed the 'return' keyword when there's only one currency.


JohnsonYuan:
Maybe on purpose do this?

Issue #1424, opened 2016-05-19
OrderTotal Rounding


AndreiMaz:
Would be nice to have the ability to configure OrderTotal rounding.
Eg. in Sweden decimals (ören/cents) are not used anymore, and when using discount codes with percentage, the order total get decimals.

 i want the order totals (only) to look like this:
Sub-total: 10,00
Tax         2,50
## Rounding:   0,50

Total:     13,00

See discussion at http://www.nopcommerce.com/boards/t/41958/suggestion-ordertotal-rounding.aspx (waiting for reply)

and https://en.wikipedia.org/wiki/Cash_rounding

See related work item - https://github.com/nopSolutions/nopCommerce/issues/677

Contribution: https://github.com/nopSolutions/nopCommerce/pull/1983 and https://github.com/nopSolutions/nopCommerce/pull/1992

mfalkao:
My I suggest a slightly different implementation.
Instead of using the suggested RoundingType it would be sufficient to state the precision as decimal (like 0.01, 0.05, 0.5) and a rounding type the states midpoint rounding (to even, AwayFromZero).

Rounding could then be implemented as:
v: the value to round
p: precision as decimal
m: midpoint rounding

q = Math.Truncate(v/p);
r = Math.Round(v/p - q, p);
result = (q + r) * p;

Furthermore I would suggest to have two rounding settings per currency. One for Price and one for Amount.

skoshelev:
Closed #1424

Issue #677, opened 2016-01-19
Rounding (away from zero). Configurable (solution attached)


mariannk:
Seems like currently the tax calculation of NC doesn't include rounding based on the sub-total. Here is an example: 

If the Sales Order Sub-total is: $100.05 
Tax percentage is: 10% 

The calculation result from current NC is: $10.00, while if we consider the the most frequently used rounding (rounding to 1 when great than 0.5), it should be $10.01 

Solution: http://www.nopcommerce.com/boards/t/35157/suggestion-rounding-of-tax.aspx

See related work item - https://github.com/nopSolutions/nopCommerce/issues/1424

Contribution: https://github.com/nopSolutions/nopCommerce/pull/1983

AndreiMaz:
done

PR #1983, opened 2016-12-19
Feature  rounding options configurable from admin. Related #677, #1424

BrandonSpor:
Because RoundingHelper.RoundPrice handles all rounding, I would like to suggest adding the ability to configure some basic settings for how the rounding is handled.

The first option (RoundingPrecision) allows users to pass in the number of decimal places they would like to round to. Default would be 2. Right now there is no validation if a user enters a negative number. I could add it if suggested.

The second option (UseRoundingPrecisionForPriceDisplay) is would mostly be used by b2b implimentations where a buyer might want to see the entire price. In this case the same number of digits used for rounding would also be shown for the price on a product. It is disabled by default to use the standard 2 places.

The third option (RoundAwayFromZero ) just allows the admin to use this as a rounding options. While there has not been a lot of demand for this, it has come up.

Lastly, I am passing the Catalog Settings into the helper method so that the helper method does not have to resolve the settings. In most cases the settings were already loaded in the class where the helper method is being used. 

By allowing these features out of the box, it would make it easy for plugin developers to handle these cases without modifying core code or over-riding many services, especially since all of this logic passes through the single line of code in the RoundingHelper.

### DIFF
--- a/src/Libraries/Nop.Core/Domain/Catalog/CatalogSettings.cs
+++ b/src/Libraries/Nop.Core/Domain/Catalog/CatalogSettings.cs
@@ -358,5 +358,20 @@ namespace Nop.Core.Domain.Catalog
         /// Gets or sets a value indicating whether need create dropdown list for export
         /// </summary>
         public bool ExportImportUseDropdownlistsForAssociatedEntities { get; set; }
+
+        /// <summary>
+        /// Gets or sets a the number of significant digits in the RoundingHelper
+        /// </summary>
+        public int RoundingPrecision { get; set; }
+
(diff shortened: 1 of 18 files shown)


AndreiMaz:
@BrandonSpor Thanks a lot! We'll definitely use it. In the meantime, could you also please answer my questions at http://www.nopcommerce.com/boards/t/41958/suggestion-ordertotal-rounding.aspx#176813

RomanovM:
Thanks for your contribution. We have already implemented it. See https://github.com/nopSolutions/nopCommerce/issues/1424

Issue #2040, opened 2017-01-18
Performance. LanguageService should uses the "static cache" instead of "per request cache"


MungoWang:
Due to the IWorkContext is a per request instance, so we will get a new instance when using the IWorkContext.WorkingLanguage which is a populated property and will invoke the method GetAllLanguages.

AndreiMaz:
It's by design. Entitie are not cached in EF

PR #2041, opened 2017-01-18
#2040-Change ICacheManager for LanguageService from "per request" to "static"

MungoWang:


AndreiMaz:
Thanks! But it's not by design. please see comment at https://github.com/nopSolutions/nopCommerce/issues/2040

EF entities doesn't support second-layer caching

Issue #2022, opened 2017-01-12
Admin section links to help (external site). Open in a new window


AndreiMaz:
It would be a nice enhancement to have these links open in a new tab or window

Source: http://www.nopcommerce.com/boards/t/45741/admin-section-links-to-help-and-public-store.aspx

PR #2043, opened 2017-01-19
#2022 Admin section links for help are opened in new tab.

manojkulkarni30:
#2022 A new field called OpenUrlInNewTab is added in SiteMapNode.cs file which allow to configure particular URL to open in new tab through sitemap.config file. @AndreiMaz Please review the PR.


AndreiMaz:
Great! Thanks a lot for this contribution!

PR #2084, opened 2017-02-02
Fix #2083 - Force https when requesting current page url in a secure connection

### LINKED ISSUE #2083: When using SSL behind a proxy (ie Cloudflare), admin locale is not en-US
In order the admin zone to work properly with numbers, it's important that the Culture is set to 'en-US'. This is done in `SetWorkingCulture` in `Global.asax.cs`, calling `CommonHelper.SetTelerikCulture` when the url matches with admin.
The problem is that it's comparing `WebHelper.GetThisPageUrl(false)` with `WebHelper.GetStoreLocation()'. The former method calls internally to 'WehHelper.GetThisPageUrl` and in this case, just returns `HttpContext.Request.Url`, which does not take in account any header that may indicate that the original request (and the effective store location) is https.
So, the urls don't match and the Culture is not set to `en-US` so if your store culture uses commas instead of dots to indicate decimal numbers, you're screwed.

mcliment:
This is how I fixed it in our production code.

### DIFF
--- a/src/Libraries/Nop.Core/WebHelper.cs
+++ b/src/Libraries/Nop.Core/WebHelper.cs
@@ -206,6 +206,12 @@ namespace Nop.Core
                 }
             }
             url = url.ToLowerInvariant();
+
+            if (useSsl && url.StartsWith("http://", StringComparison.Ordinal))
+            {
+                url = url.Replace("http://", "https://");
+            }
+
             return url;
         }
 


RomanovM:
Thanks for your contribution. We have implemented it in https://github.com/nopSolutions/nopCommerce/commit/cf7fbacb8b690d70eb7f191c197ddc04f21f75f3.

RomanovM:
Related task https://github.com/nopSolutions/nopCommerce/issues/2083

PR #2099, opened 2017-02-07
Fix #1790 Publish Inserted, Changed and Deleted Events for OrderItems

### LINKED ISSUE #1790: Publish Inserted, Changed and Deleted Events for OrderItems for Fulfillment
For fulfilling orders via third parties, it would be convenient if events regarding inserted, changed and deleted to OrderItems were published individually. Currently, the only way to do this properly is with custom checks.

I have never tried a pull request and I am a crappy coder, but I could try to tackle this.

Source: http://www.nopcommerce.com/boards/t/44525/fulfillment-feature-request.aspx

darocker22:
We also needed to write a plugin that fired when OrderItems were Added and or Updated. The Delete event was already in the code.

### DIFF
--- a/src/Libraries/Nop.Services/Orders/IOrderService.cs
+++ b/src/Libraries/Nop.Services/Orders/IOrderService.cs
@@ -116,6 +116,18 @@ namespace Nop.Services.Orders
         IList<OrderItem> GetDownloadableOrderItems(int customerId);
 
         /// <summary>
+        /// Inserts an order item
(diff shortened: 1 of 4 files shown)


RomanovM:
Related task #1790

RomanovM:
Thanks for your contribution. But this should be done for all navigation properties not just for order items. Still discuss in https://github.com/nopSolutions/nopCommerce/issues/1790

Issue #373, opened 2016-01-18
Description for checkout/product/specification attribute values (contribution)


mariannk:
Description for checkout/product/specification attribute values: 

https://nopcommerce.codeplex.com/SourceControl/network/forks/rudgr/attributevaluedescription 

and submitted pull request: 

https://nopcommerce.codeplex.com/SourceControl/network/forks/rudgr/attributevaluedescription/contribution/8508

Let's decide whether we should support description for attributes or values

rudgr:
This implementation allows store owners to add Description to checkout/product/specification attribute values.
For example this would allow you to show blue info marker icons that could show more info to customer about the attribute value option. (for example why customer should choose this option and pay more ;-)

[http://demo.nopcommerce.com/build-your-own-computer](http://demo.nopcommerce.com/build-your-own-computer)

rudgr:
hi @mariannk 

was wondering if this nice feature will be included in nop3.80? ;-)

thanks!

AndreiMaz:
![f020d45a-de7b-11e6-8c8b-ba525607db49](https://cloud.githubusercontent.com/assets/5019111/22120941/e793c164-de92-11e6-9429-67bd1d5b7bbe.png)

rudgr:
sorry @AndreiMaz my russian is not so good... what does this screenshot mean? tx

AndreiMaz:
@rudgr it's just an idea how it could be displayed (descriptions) - hints (popups)

rudgr:
great!:)

RomanovM:
After some consideration we've decided not to implement this functionality.

PR #2118, opened 2017-02-15
New feature for adding product attribute value description

Dochef:
This pull request is according feature request #373 

Initially the pull request was made on Codeplex and now we recreated it here for your convenience. This implementation allows store owners to add Description to checkout/product/specification attribute values.
It would be great if we see it in v3.90

Спасибо,
Д. Дочев


RomanovM:
Thanks for your contribution, but we are still thinking about how better to implement this feature. Let's discuss in the https://github.com/nopSolutions/nopCommerce/issues/373

PR #2119, opened 2017-02-15
Show weight/dimensions

seasong:
Show weight/dimensions


RomanovM:
Thanks for your contribution. Right now we don't plan to implement it. But we open new issue for it https://github.com/nopSolutions/nopCommerce/issues/2685

PR #2140, opened 2017-02-20
Product editor settings dialog checkbox check status problem 

JohnsonYuan:
1. 
In product edit page, \Presentation\Nop.Web\Administration\Views\Product\_ProductEditorSettingsModal.cshtml
I think the current logic is wrong:, and current logic is :

`
                        $('input[type=checkbox].select-all-fields').each(function (index, value) {
                            if ($('input[type=checkbox]:checked', $(this).closest('.panel')).length > 0) {
                                $(this).attr('checked', true);
                            }
                        });
`

for example, "Advanced product types " category has 4 options, if "Is gift card
" is checked, "Advanced product types " will be checked according current logic.

"Advanced product types " should be checked when all options checked.

Advanced product types 
- Is gift card
- Downloadable product
- Recurring product
- Is rental

2. 
IN CommonController.cs 

CurrencySelector method:
`if (model.AvailableCurrencies.Count == 1) 
          Content(""); 
return PartialView(model); `

LanguageSelector method: 
if (model.AvailableLanguages.Count == 1)
                Content("");
return PartialView(model);

I think we should return empty content, not PartialView when items count is 1.


RomanovM:
@JohnsonYuan Thanks for the contribution. But since the creation this pull request, many files have been modified or deleted. Could you please update it to the latest version?

Issue #2025, opened 2017-01-13
Shipping plugins should properly handle products marked as "free shipping"


AndreiMaz:
Intro: Shipping plugin developers should not manually handle such products. Let's just not pass them to plugins. We can filter them in the "CreateShippingOptionRequests" method of "ShippingService". Handle properly if all items are free. Or even of some are free (multiple requests) - ensure no empty requests are created. Otherwise, it can cause issues in plugins. Ensure that estimate shipping works fine when all items are free. On the other hand, it'll cause issues when we have, for example, a "free shipping" product but with not "free shipping" associated product. So maybe, it's better to leave for plugin developers to handle it

Let's do in our built-in plugins

PR #2163, opened 2017-02-25
ShippingFixedOrByWeight: do not consider FreeShipping items in weight…

mfalkao:
… calculation. Find FixedRate also in ByWeight calculation.
Shipping.FixedOrByWeight plugin can be enhanced.
1) When there is more than one item to ship, those that have "ship for free" are included in weight calculation. As they should ship for free, shipment fee shouldn't consider them.
2) A fixed rate shipping can also be set in ByWeight mode and GetFixedRate in FixedOrByWeightComputationMethod should consider this.


RomanovM:
Thanks for your contribution. We have already implemented it in https://github.com/nopSolutions/nopCommerce/commit/69c054a6d11a456386e53d63638c6fc9e786d67d

RomanovM:
Related task https://github.com/nopSolutions/nopCommerce/issues/2025

Issue #1964, opened 2016-12-14
BrowscapXmlHelper performance issue


AndreiMaz:
Intro and details: http://www.nopcommerce.com/boards/t/45428/reducing-the-memory-usage-on-startup.aspx

I decided to get to the bottom of this and downloaded the trial version of dotMemory and snapshot the memory usage during the peak time memory consumption (you need to be very quick as the GC may reduce the memory at any point).
It turned out the there are 500 MB ... yes 500 MB in Strings allocated in the memory heap.
When investigated who allocated so much stings in the memory it turned out that it is the constructor of the BrowscapXmlHelper class.
You can easily spot why this happens as there is a huge 40 MB file App_Data\browscap.xml that is being loaded in a String into the memory. Then there are several Replace methods that actually create new strings rather than modifying the existing one.

By simply commenting the the Initialize method invocation in the constructor of the BrowscapXmlHelper, then I couldn't make my nopCommerce to take more than 200 MB even after several restarts.

So the implementation of the BrowscapXmlHelper class needs to be corrected to ensure all nopCommerce users will run smoothly their nopCommerce websites on a shared hosting.

see http://stackoverflow.com/questions/6524528/string-replace-vs-stringbuilder-replace

skoshelev:
Closed #1964

PR #2181, opened 2017-03-01
Fix for BrowscapXmlHelper to use significant less resource per reques…

jessek1:
…t by using cache.

Change for BrowscapXmlHelper's instance to be managed by Autofac container.

### DIFF
--- a/src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
+++ b/src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -15,11 +14,12 @@ namespace Nop.Services.Helpers
     /// </summary>
     public class BrowscapXmlHelper
     {
+        private readonly List<string> _crawlerUserAgentsRegexp;
 
-        private ConcurrentDictionary<String, bool> _userAgentsCache;
-       
         public BrowscapXmlHelper(string userAgentStringsPath, string crawlerOnlyUserAgentStringsPath)
         {
+            _crawlerUserAgentsRegexp = new List<string>();
+
             Initialize(userAgentStringsPath, crawlerOnlyUserAgentStringsPath);
         }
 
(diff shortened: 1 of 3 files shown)


src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
```diff
@@ -14,12 +15,11 @@ namespace Nop.Services.Helpers
     /// </summary>
     public class BrowscapXmlHelper
     {
-        private readonly List<string> _crawlerUserAgentsRegexp;
+
+        private ConcurrentDictionary<String, bool> _userAgentsCache;
```
skoshelev: I don't think using ConcurrentDictionary is necessary for this task because this list a never changes after creating .

src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
```diff
@@ -101,7 +104,11 @@ private static string ToRegexp(string str)
         /// <returns>True if user agent is a crawler, otherwise - false</returns>
         public bool IsCrawler(string userAgent)
         {
-            return _crawlerUserAgentsRegexp.Any(p => Regex.IsMatch(userAgent, p));
+            // check against cache object
+            bool result;
+            if (_userAgentsCache.TryGetValue(userAgent, out result))
```
skoshelev: You have missed that browscap.xml file stored not the agents, but the mask for identifying them, so your implementation is not finalized and will not carry out the main task.

src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
jessek1: Agree, original impl was storing request user agents with a max

src/Libraries/Nop.Services/Helpers/BrowscapXmlParser.cs
jessek1: Will look into, originally had request ua's and their result stored, may need to do discovery on the performance of the permutations but likely still significantly faster than the regex.

skoshelev: It would be great if you come up with some decision on removing Regex.IsMatch challenges in the task where the user agents describes by templates.

jessek1: New logic updated in UserAgentHelper, uses a found cache with fallback to a cookie and then a final fallback to BrowscapXmlParser.

src/Libraries/Nop.Services/Helpers/UserAgentHelper.cs
```diff
@@ -49,7 +54,33 @@ public virtual bool IsSearchEngine()
...
+                var cookie = _httpContext.Request.Cookies[IS_SEARCH_ENGINE_REQUEST_COOKIE_KEY];
+                if (cookie != null)
+                    return cookie.Value == "1";
+
+                // check browscap
+                result = _browscapXmlHelper.IsCrawler(userAgent);
+
+                // if IS search engine, save to cache
+                if (result)
+                    _foundSearchEnginesCache.TryAdd(userAgent, true);
```
skoshelev: This approach problem is that  it becomes possible to organize a simple attack, the result of which will be the exhaustion of all available RAM and closing nopCommerce with the OutOfMemory exception

jessek1: Yes good point, there should be a ceiling and honestly a mechanism to maintain itself, i.e ttl or sort strategy.

Could use memory cache manage for inherent ttl expiration?

RomanovM:
Thanks for your contribution. We have implemented it. See https://github.com/nopSolutions/nopCommerce/issues/1964

PR #2213, opened 2017-03-10
Use <a> tags instead of <div> tags for the product thumbnails

poyker:


mariannk:
Hey Boyko! Could you please clarify: what are these changes for?

poyker:
Hi Marianna,

This was basically a "revert-request" for one of the previous versions of nopCommerce.
At that time we believed that such a change will require pre-styling of the thumbs elements by all theme vendors and requested you to revert it but you did not agree. So now when all theme vendors have already styled these new elements, this change does not make sense anymore (actually it will cause problems and will require again re-styling of the thumbnails). 

So you can simply ignore this pull request :)

AndreiMaz:
@poyker thanks

Issue #2200, opened 2017-03-07
Log should save URL "as is" (do not lowercase)


turbochris:
So I was working on a feed provider which send a hash key in SHA512. It was a get request.
example:
www.mystore.com/feed/feedprovider?hash=XXXX&salt=XXXX

So while debugging I used the log to test the url. Problem with this is that I discovered that "GetThisPageUrl" in webhelper returns url in lowercase. When dealing with authentication it's important to have casing. So I am asking that the "InsertLog" get's its own method to insert url "as-is" without any modifications at all. This makes debugging easier and what the log is for.

Let's add a new parameter to GetThisPageUrl() method indicating whether it should return lowcased URL or not (default value set to "true"). And we'll pass "false" only when saving lor records

also see realrted work item https://github.com/nopSolutions/nopCommerce/issues/2407

AndreiMaz:
I think we can add a new parameter to GetThisPageUrl() method indicating whether it should lowcase URL (default value set to "true")

turbochris:
Agree :)

rdakar:
I got this issue to do, but a problem has arisen.

The method GetThisPageUrl() has two overloads:
GetThisPageUrl(bool includeQueryString)
GetThisPageUrl(bool includeQueryString, bool useSsl)

If we add a new bool parameter with default value, those who use the second method will switch to using the first method without knowing.
In nopCommerce solution, the second is used only in NopHttpsRequirementAttribute, and it's easy to change. But, in thirdy part plugins, this can be a silent problem.

Perhaps it is better to create a new method in WebHelper.

What do you think?

turbochris:
That does pose a problem. I agree that you should make sure it's compatible.

I think you should make a new method without optionals:
GetThisPageUrl(bool includeQueryString, bool useSsl, bool ignoreCasing)

The existing methods:
GetThisPageUrl(bool includeQueryString)
GetThisPageUrl(bool includeQueryString, bool useSsl)

Should all call the GetThisPageUrl(bool includeQueryString, bool useSsl, bool ignoreCasing) method

I don't see any possibility of this conflicting...

rdakar:
I agree, in this case we have no conflict.
But, following this line, another possibility would be to create the default parameter only in the method: GetThisPageUrl (bool includeQueryString, bool useSsl).

We would have the following scenario:
GetThisPageUrl (bool includeQueryString)
[...]

PR #2380, opened 2017-06-02
Feature #2200 - Log should save URL "as is" (do not lowercase)

rdakar:
Created a new default parameter in the method IWebHelper.GetThisPageUrl to return the page URL "as-is", without lowercase.

We now have the following scenario in IWebHelper:
GetThisPageUrl (bool includeQueryString)
GetThisPageUrl (bool includeQueryString, bool useSsl, bool ignoreCase = true)

Adjust DefaultLogger.InsertLog to use new parameter, passing value "false":
_webHelper.GetThisPageUrl(true, _webHelper.IsCurrentConnectionSecured(), false)


RomanovM:
Thanks for your contribution. We have already implemented it. See https://github.com/nopSolutions/nopCommerce/issues/2200

PR #2451, opened 2017-07-21
Replaced Html.Hidden with tag helpers 

hathanhhung:
Replaced Html.Hidden with tag helpers


hathanhhung:
close this. I miss some mistake on git desktop

PR #2454, opened 2017-07-22
Avoid calculating tax rate when it is not necessary

jasonmcboyd:
Background. A client is using a tax calculation plugin that is a little resource intensive. The product catalog is configured not to include tax in the product price but the way 'GetProductPrice' is currently implemented it still calculates the tax rate for every product on the page and then discards the result. This is adding about ~2000 milliseconds to my client's product catalog page. This is a simple change that just avoids calculating the tax rate if it is not necessary.


RomanovM:

Issue #2478, opened 2017-08-04
"Ignore" property of filter is ignored


AndreiMaz:
CheckAccessClosedStoreAttribute, CheckAccessPublicStoreAttribute, etc

For example, CheckAccessPublicStoreAttribute. Disable navigation for guests and you'll be redirected to the login page in a loop. ASP.NET Core tries to use BOTH filters now for some reasons (added to the action method and controller). And controller one should NOT be executed

Also see https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters

PR #2482, opened 2017-08-08
"Ignore" property of filter is ignored #2478

vitalya-kravchuk:
Now Customer controller described with attributes:
`[CheckAccessClosedStore(true)]`
`[CheckAccessPublicStore(true)]`
So, all methods in this controller do not check access to store (by default), but those that should check access - with attributes:
`[CheckAccessClosedStore]`
`[CheckAccessPublicStore]`
Approach is not perfect, but in this way you can control access to the login and other pages.


AndreiMaz:
@vitalya-kravchuk thanks a lot anyway!

Issue #2448, opened 2017-07-20
Code polishing


RomanovM:
1. Duplication of logic in two or more places.
2. Undesired dependencies. For applications divided into layers such as persistence, business and user interface, the dependencies between the layers need to be managed. Finding undesired dependencies such as the use of persistence code directly within the user interface layer is a strong indicator of a design problem. Move references to Nop.Core
3. Overly complex logic, like a large switch statement or series of nested if-else statements. Sometimes such logic can be drastically simplified through the use of object-oriented design principles.
4. Too many classes within one package.
5. Overly long file, in excess of a couple thousand lines. Even in Java where each public class is forced to be in its own file, it is possible to end up with very long files. This may be because there is too much logic within the class, or it may be because of many nested classes or private classes in the same file. In the latter case, these can be moved to one or more separate files.
6. Overly long methods. Anything longer than about one screen warrants a closer look to see if it can be simplified, usually via an extract method refactoring.
7. Poorly named classes, methods, fields or local variables. One indicator I look for is a one-sentence comment describing a vaguely named variable or method. Usually the variable or method can be renamed to capture the essence of the comment and the comment can then be deleted.
8. Lack of a class comment. A concise summary of the purpose of the class is often quite useful. If it is difficult to come up with such a summary, it is an indicator that the class may be trying to do too much or does not have a clear focus.
etc..
9. Improve the code design: use the initialization of classes and collections everywhere, replace the “for” loops to LINQ expression (where applicable)
10. Check the code with static analyzers, such as StyleCop and PVS-Studio
-----------------------------------
11. Too many methods within a class, particularly public methods. This is often an indicator that the class is trying to do too much and should be split into multiple classes.
12. Refactoring of big methods (split to several ones). For example, the PlaceOrder method from OrderProcessingService uses #region 3 times. The best practice says you shouldn't use regions inside methods. Methods should be short and each method should do only one thing. So instead of separating the method into 3 regions, you should refactor into 3 or more separate methods. The PlaceOrder method is not the only place where region is used inside methods, that's why the title was more general.
[...]

PR #2489, opened 2017-08-12
Minor refactoring

f1ana:
Based upon findings from PVS-Studio
Focused on high severity issues such as potential null referencing and copy-paste errors


RomanovM:
Thanks for the contribution. But since the creation this pull request, many files have been modified or deleted. Anyway we have issue for it https://github.com/nopSolutions/nopCommerce/issues/2448

f1ana:
Thanks for the update.  One thing that I've observed with this repository lately is that issues and pull requests often do not get any real attention until months later (such as this one), where the merge is no longer possible due to conflicts in the 2+ months that this has been here.  As a contributor, this is discouraging and makes me less likely to submit any features or improvements in the future.  I'm not sure what the resolution would be for this going forward, but I do not feel that the current process is working from a community standpoint.

PR #2506, opened 2017-08-22
Minor fix

youngmtool:
Line 107, SystemName to Configure


AndreiMaz:
Thanks. But it's done this way for all other views. Actually it doesn't matter as we use a custom template for this column

PR #2507, opened 2017-08-22
Minorfix of argument null check in pluginmanager

thomasjoscht:
simple minor fix in PluginManager Initialize method.


AndreiMaz:
Thanks a lot!

PR #2508, opened 2017-08-22
Fix Chinese typo: 信心 -> 信息

kchanlee:
_信心 (Xin Xin)_ means _confidence_ while _information_ means _信息 (Xin Xi)_.


AndreiMaz:
Thanks a lot!

PR #2537, opened 2017-09-05
Added space to locale resource in plugin install.

Moosebiter:


AndreiMaz:
@Moosebiter thanks a lot! Please also update upgrade.sql.
BTW, there's one more space missing in "Analyticsaccount". It would be great if you can fix it as well in the same commit

Moosebiter:
Sure thing @AndreiMaz , I added the additional space and updated the SQL file. I also squashed the commits to make the history a bit cleaner.

AndreiMaz:
@Moosebiter Great! Thanks a lot!

PR #2548, opened 2017-09-07
Update ProductTemplate.Grouped.cshtml

tmoodley:
This javascript will iterate through all the variants and add a new div with the name of the variant and link to a shortcut to the variant section of the page.

The current implementation is not user friendly, this was added to increase the usability of the page.  
heavy-based-pilsner 7

heavy-based-pilsner 8


RomanovM:

Issue #2137, opened 2017-02-19
Admin area grids. Make rows clickable (remove "Edit" buttons)


AndreiMaz:
Source: http://www.nopcommerce.com/boards/t/46252/suggested-changes-in-the-admin-site.aspx

AndreiMaz:
After some discussion decided not to implement it out of the box. We leave it for customization

PR #2552, opened 2017-09-08
navigating to product page in (admin) products list.

Arashz22:
by this edit like other E Commences(open cart , Etc.), in the admin area when you see products list you can easily navigate to the product page by clicking on the name of each product.


RomanovM:

RomanovM:
Related task https://github.com/nopSolutions/nopCommerce/issues/2137

PR #2588, opened 2017-09-24
ModifyQueryString and RemoveQueryString methods were failing in unit tests

alessandromontividiu03:
I don't think RemoveQueryString and ModifyQueryString should change anchor to lowercase, but some tests were failing because they were expecting anchor in lowercase.


AndreiMaz:
@alessandromontividiu03 Thanks a lot! You're absolutely right. Could you please please delete (not just update) "Can_remove_queryString_should_ignore_input_parameter_case" as it's obsolete now

AndreiMaz:
Thanks a lot!

Issue #2592, opened 2017-09-28
Download entry is being created if ReturnRequests allowed but not file uploads (pull request)


Xzelsius:
Consider using OR instead of AND.

https://github.com/nopSolutions/nopCommerce/blob/b0312212e1d27ef42e96115a9ccd4a992478b413/src/Presentation/Nop.Web/Controllers/ReturnRequestController.cs#L182

_This bug only occurs if you 'fake' the upload request since orderSettings.ReturnRequestsAllowFiles = false prevents rendering the file uploader_

Xzelsius:
Additionally i didn't need to login to upload a file, maybe this should be fixed to

PR #2593, opened 2017-09-28
#2592 Prevent uploading files

Xzelsius:
if ReturnRequestsEnabled = true and ReturnRequestsAllowFiles = False

See #2592


AndreiMaz:
@Xzelsius Thanks a lot!

Issue #2601, opened 2017-10-01
Inactive Topics links on the sitemap page


MarinaAndreeva:
Links to Topics on the sitemap page (not xml) don't have **href** attribute. E.g. `<a>Contact Us</a>`

AndreiMaz:
@MarinaAndreeva Marina, it works fine. SeName is received in the view (cshtml)

PR #2602, opened 2017-10-01
Fix Topic's links on the sitemap page #2601

MarinaAndreeva:


AndreiMaz:
@MarinaAndreeva Thanks, Marina! Bt it works fine. SeName is received in the view (cshtml)

PR #2634, opened 2017-10-11
Change breadcrumb microdata to schema.org in the default template

MarinaAndreeva:
Add missing breadcrumb microdata to category template.
Change microdata vocabulary on the product page, because currently used
data-vocabulary.org was abandoned by Google. It is better to use
schema.org vocabulary.


skoshelev:
@MarinaAndreeva Thank you for your help

PR #2635, opened 2017-10-11
[FEATURE] Optional encryption of settings.json to prevent exposing password as cleartext.

mmccann-umph:
This feature addresses the narrow scope security case where misconfigured IIS allows for the downloading of settings.txt or settings.json.  

In the current codebase these files store database passwords and user accounts in as unencrypted clear text.  

This feature uses dotNet's ProtectedData API (scoped to the account the application is running under) to encrypt the values found in the dataSettings.json file so that in the event the file is exposed the sensitive information isn't available in clear text.


SamandarHijiakbarovich:
Can I join this project?

Issue #2644, opened 2017-10-14
Move jQuery library to the footer (public store)


AndreiMaz:
We already have other libraries and inline scripts moved to the footer. Should we do it for jQuery?

Just ensure that it'll work fine for \Views\Shared\ _CreateOrUpdateAddress.cshtml which has inline script without moving it to the footer. It should be ignored as it's used in AJAX requests on one-page checkout (e.g. billing and shipping address). This file doesn't have head and footer, hence we cannot move it. For example, states won't be loaded on the billing address step during checkout when you change a country

Ideas?

Xzelsius:
Maybe try to refactor inline scripts into dedicated script files?
I think of something like bootstrap does it.

Instead of inline script you could use data-attributes and initialize handlers in that dedicated JS file which will be loaded after jQuery.

Take this as a sample of what i mean

**_CreateOrUpdateAddress**
` <select asp-for="CountryId" asp-items="Model.AvailableCountries" 
                    data-trigger="country-select" 
                    data-url="@(Url.RouteUrl("GetStatesByCountryId"))" 
                    data-stateprovince="#@Html.IdFor(model => model.StateProvinceId)" 
                    data-loading="#states-loading-progress"></select>`

**public.countryselect.js**
`+function ($) {
    'use strict';

    if ('undefined' == typeof (jQuery)) {
        throw new Error('jQuery JS required');
    }

    if ($(document).has('[data-trigger="country-select"]')) {

        $('select[data-trigger="country-select"]').change(function() {
            var $this = $(this);
            var selectedItem = $this.val();
            var ddlStates = $($this.data('stateprovince'));
            var statesProgress = $($this.data('loading'));
            statesProgress.show();
            $.ajax({
                cache: false,
                type: 'GET',
                url: $this.data('url'),
                data: { 'countryId': selectedItem, 'addSelectStateItem': "true" },
                success: function(data) {
                    ddlStates.html('');
                    $.each(data,
                        function(id, option) {
                            ddlStates.append($('<option></option>').val(option.id).html(option.name));
                        });
                    statesProgress.hide();
                },
                error: function(xhr, ajaxOptions, thrownError) {
                    alert('Failed to retrieve states.');
                    statesProgress.hide();
                }
            });
        });
    }

}(jQuery);`

**_Root.Head**
`Html.AppendScriptParts(ResourceLocation.Foot, "~/js/public.countryselect.js");`

Xzelsius:
[...]

Issue #154, opened 2016-01-15
Add javascript events


mariannk:
We started working on Google Analytics Enhanced ecommerce plugin for nopCommerce. 
Unfortunately nopCommerce is lacking any javascript events for any user actions. For example there is no way to know when a product has been added to the cart. 
It would be nice if you raise javascript events like "productAddedToCart" etc. for the actions that happen via Ajax.
For example when you add a product via Ajax it would be nice to have something like this: 

$.event.trigger({ type: "nopAjaxCartProductAddedToCartEvent", productId: productId, quantity: addToCartProductQuantity }); 

This way we can subscribe to such events and submit the data to Google Anaytics for example. 
Unfortunately the Remove from Cart does not use Ajax, so there is no way this to be done for now. 

Source: http://www.nopcommerce.com/boards/t/38391/suggestions-for-the-upcoming-version-370.aspx?p=2#156114

also see http://www.nopcommerce.com/boards/t/43297/improvements-before-380-release.aspx#171728

also see http://www.nopcommerce.com/boards/t/39325/add-javascript-events-in-the-payment-flow.aspx

P.S. We already have such event defined in \Views\Product_ProductAttributes.cshtml

AndreiMaz:
After some consideration we've decided not to implement this functionality out of the box and leave it for customization

PR #2648, opened 2017-10-15
#2644 Moved JS for country select to separate script.

Xzelsius:
Refactored inline JS for country select into a separate script.
The script now is 'reusable', meaning you only have to place the proper data-attributes to the country select.

This pull-request only contains changes for front-end. So no changes in admin.
I also didn't touched all the default plugins, maybe there is refactoring required too.

### DIFF
--- a/src/Presentation/Nop.Web/Views/Customer/Info.cshtml
+++ b/src/Presentation/Nop.Web/Views/Customer/Info.cshtml
@@ -15,37 +15,6 @@
     @await Component.InvokeAsync("CustomerNavigation", new { selectedTabId = CustomerNavigationEnum.Info })
 }
 
-@if (Model.CountryEnabled && Model.StateProvinceEnabled)
-{
-    <script type="text/javascript" asp-location="Footer">
-        $(document).ready(function () {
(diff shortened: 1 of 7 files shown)


AndreiMaz:
@Xzelsius Thanks. Seems like it doesn't workin the following case:
1. Go to billing address on one-age checkout
2. Choose "New address"
3. Choose "USA" and American states will be loaded. But don't fill other required fields
4. Click "Continue"
5. You'll validation errors. But now when you change a country, new states won't be loaded

Xzelsius:
For me it does work or i don't get what you meant

case

Are you sure your browser cache doesn't produced this?
Since i changed some minor parts in OPC java script.

DmitriyKulagin:
Hi, @Xzelsius 

After some changes ([#154 Added some events](https://github.com/nopSolutions/nopCommerce/commit/72692094e6713e834f1ed5640efc0dec644abad8) and [Updating third-party libraries](https://github.com/nopSolutions/nopCommerce/commit/408b0b06a459bc094476445297123dc5b2e33028)) we checked your pull request and decided to add it, despite the fact that you closed the branch.

Thanks a lot!

PR #2735, opened 2017-11-23
Add identificators for external products load

ArturAlekseev:
We often need to load products automatically from external sources and update them (from product suppliers).
To identify that product is loaded from external supplier and update it we need two references for it.
One to identify supplier (I use ExternalType for it) and one to identify product in supplier price list (or on the web page). Also we need same identificators for manufacturer and for categories (to copy supplier manufacturers and categories list). Then they can be used to automatically copy products to shop published categories and etc.
I suggest we add such identificators to original database. Maybe with different names. But I think name should start from "External" for easy remembering.


ArturAlekseev:
Hm. I'v deleted this fork (thought this commit will also disappear) and made a new one from another fork here: https://github.com/nopSolutions/nopCommerce/pull/2736

Issue #2737, opened 2017-11-23
Categories load. Performance optimization


ArturAlekseev:
Ok. I'v started to update v4.0 for my needs and here is the first issue.
Will try to resolve it myself and publish commit but so far will write about it.

We use this function:
PrepareCategorySimpleModels();
That loads all categories and prepares their models recursively.
This function is used 2 times:
For top menu generation and for navigation.

The problem is seen when there are many categories. For example 20000. Then everything goes extremely slow.

First we do not need to load all categories for top menu because not all are there.
Second we do not need to load all categories for category navigation (only those that are higher than selected and one level lower).
Ofcourse if we load different categories sets then we should not cache results (now result of all is cached). But anyway - with each categories change it will be regenerated and it is not acceptable for large categories list.

And another one:
foreach (var category in categories)
            {
                var categoryModel = new CategorySimpleModel
                {
                    Id = category.Id,
                    Name = category.GetLocalized(x => x.Name),
                    SeName = category.GetSeName(),
                    IncludeInTopMenu = category.IncludeInTopMenu
                };

GetLocalized & GetSeName executes 2 queries to database for each category.
It is very slow.
We should load localized names and se names using IN statement for all categories before foreach word and then use loaded list inside foreach.

I will add CommonSetting called LargeDatabase and create different code for when it is set to true. So that we can use both approaches.

Xzelsius:
GetLocalized and GetSeName use a setting to decide whether to cache all entries at first usage or not. See LocalizationSettings. So this shouldn't be a problem.

But a setting for "LargeDatabase" would be bad. _(Since "large database" depends on the power of your SQL server)_

I think it would be better to just write a plugin for your large database where you override services/factories that struggle with your large database.

ArturAlekseev:
No matter which power of sql server you have - 1000 sql server requests will be nearly 1000 times slower than 1 request.
It is not possible to override PrepareCategorySimpleModels function work using plugin. It is not supported.

AndreiMaz:
@ArturAlekseev it's by design. we load all categories. but we do it only once (for all pages). It takes some time but only during the first site visit. So it's by design

ArturAlekseev:
I know that it is by design and not a bug. But this design is not usable for large categories count. Because first time categories are loaded is very long for large categories count. And for each time cache is cleared it will be regenerated. And these:
Name = category.GetLocalized(x => x.Name),
SeName = category.GetSeName(),
for 20000 categories is like forever.

[...]

PR #2741, opened 2017-11-26
Many categories part 1 - Nop.Web top menu and navigation speedup

ArturAlekseev:
Description:
1. We add setting for LargeDatabase.
2. We make different approach for top menu load and navigation load.
For top menu only included in top menu categories are loaded and cached separately.
For navigation load only categories that are needed for tree generation are loaded. And they are not cached (maybe we should also add child categories to selected one to also display them in the tree - for easier navigation).
During first execution of recursive proc we also load all slugs and all localized names for all categories that will be used later (using 2 sql requests).
In total this makes only few requests to database instead of some requests per loaded category.
Top menu is cached anyway so LargeDatabase setting is not used for it.
LargeDatabase setting is used for navigation part. If it is true then new approach is used. If it is false then  it loads all categories as before and cache them and cache used.

Part 2 description:
In Admin area by default all categories are loaded for views for user to be able to select category. It very long too load and prepare if there are a lot of categories and also even if they are cached, then their list will be sent to user with each request. Also mobile devices will just die because of such huge lists. So based on LargeDatabase switch I stop populating AvailableCategories variable and if it its count == 0 then in views we show only editor where user can insert category id. Its everywhere except product add/update where we need extra field to be able to add new categories. There comma separated ids list is used. Also part 2 contains fix for common statistics mentioned here: https://github.com/nopSolutions/nopCommerce/issues/2744
And also categories population is removed from search part because its used in products search in the public store and again its not possible for user to wait their population. Category search is hidden there because customers can't know category ids to search by id.

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/CategoryService.cs
+++ b/src/Libraries/Nop.Services/Catalog/CategoryService.cs
@@ -14,6 +14,10 @@ using Nop.Services.Customers;
 using Nop.Services.Events;
 using Nop.Services.Security;
 using Nop.Services.Stores;
+using Nop.Core.Infrastructure;
+using Nop.Services.Localization;
+using Nop.Core.Domain.Localization;
+using Nop.Core.Domain.Seo;
 
 namespace Nop.Services.Catalog
 {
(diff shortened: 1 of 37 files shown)


AndreiMaz:
related to https://github.com/nopSolutions/nopCommerce/issues/2737

ivanslater:
@ArturAlekseev do you know how many ms it makes nop faster? Can you show some performance comparison with original source code?

ArturAlekseev:
I have > 20000 categories in one shop. With original code shop is unusable. You need to wait > 10 min to load shop for the first time. With changed pages load is lower than 1 second.

src/Libraries/Nop.Services/Catalog/CategoryService.cs
```diff
@@ -176,6 +180,63 @@ public virtual void DeleteCategory(Category category)
             }
         }
         
+        public virtual List<Category> GetAllCategoriesInTopMenu()
+        {
+            List<Category> forallstores = _categoryRepository.Table.Where(x => x.IncludeInTopMenu == true && !x.Deleted && x.Published && !x.LimitedToStores).ToList();
+            List<Category> forthisstore = _dbContext.SqlQuery<Category>($"select * from Category where IncludeInTopMenu = 1 and Deleted = 0 and Published = 1 and LimitedToStores = 1 and Id in (select EntityId from StoreMapping where StoreId = {_storeContext.CurrentStore.Id} and EntityName = 'Category')").ToList();
```
Xzelsius: This makes nopCommerce harder to maintain

src/Libraries/Nop.Services/Catalog/CategoryService.cs
```diff
@@ -176,6 +180,63 @@ public virtual void DeleteCategory(Category category)
...
+            List<int> ids = new List<int>();
+            while (parent != 0)
+            {
+                parent = _categoryRepository.Table.Where(x => x.Id == parent).Select(x => x.ParentCategoryId).First();
+                ids.Add(parent);
+            }
+            foreach (var id in ids)
+            {
+                childrenallstores = _categoryRepository.Table.Where(x => x.ParentCategoryId == id && !x.Deleted && x.Published && !x.LimitedToStores).ToList();
+                childrenthisstore = _dbContext.SqlQuery<Category>($"select * from Category where ParentCategoryId = {id} and Deleted = 0 and Published = 1 and LimitedToStores = 1 and Id in (select EntityId from StoreMapping where StoreId = {_storeContext.CurrentStore.Id} and EntityName = 'Category')").ToList();
```
Xzelsius: This makes nopCommerce harder to maintain

src/Libraries/Nop.Services/Catalog/CategoryService.cs
```diff
@@ -176,6 +180,63 @@ public virtual void DeleteCategory(Category category)
...
+            var loadLocalizedValue = true;
+            var lService = EngineContext.Current.Resolve<ILanguageService>();
+            var totalPublishedLanguages = lService.GetAllLanguages().Count;
+            loadLocalizedValue = totalPublishedLanguages >= 2;
+            if (loadLocalizedValue)
+            {
+                var workContext = EngineContext.Current.Resolve<IWorkContext>();
+                var leService = EngineContext.Current.Resolve<ILocalizedEntityService>();
+                List<int> catids = categories.Select(x => x.Id).ToList();
+                var properties = _dbContext.Set<LocalizedProperty>().Where(x => x.LanguageId == workContext.WorkingLanguage.Id && x.LocaleKeyGroup == "Category" && x.LocaleKey == "Name" && catids.Contains(x.EntityId)).ToList();
```
Xzelsius: Why not using `IRepository<LocalizedProperty>`?

src/Libraries/Nop.Services/Catalog/CategoryService.cs
```diff
@@ -176,6 +180,63 @@ public virtual void DeleteCategory(Category category)
...
+                var urlrecords = _dbContext.Set<UrlRecord>().Where(x => x.EntityName == "Category" && catids.Contains(x.EntityId)).ToList();
```
Xzelsius: Why not using `IRepository<UrlRecord>`?

src/Presentation/Nop.Web/Areas/Admin/Components/CommonStatistics.cs
```diff
@@ -47,25 +56,38 @@ public IViewComponentResult Invoke()
...
+                    NumberOfLowStockProducts = _productService.GetLowStockProducts(0, 0, 1).TotalCount +
                                              _productService.GetLowStockProductCombinations(0, 0, 1).TotalCount
-            };
+                };
+            else
+                model = new CommonStatisticsModel
+                {
+                    NumberOfOrders = _dbContext.Set<Order>().Where(x => !x.Deleted).Count(),
+
+                    NumberOfCustomers = _dbContext.SqlQuery<int>("select count(*) from Customer where Deleted = 0 and Id in (select Customer_Id from Customer_CustomerRole_Mapping where CustomerRole_Id = 3)").First(),
```
Xzelsius: This makes nopCommerce harder to maintain

ArturAlekseev: Xzelsius, is you are against LargeDatabase setting - propose something better.
About categories drop down:
1) If you have 20000 categories it takes forever to generate categories list for dropdown (because list contains full path for each category)
2) This list is very large and unusable.
3) It is transferred to client with each request. If client is mobile device then its terrible not only for bandwidth but also for cpu/ui usage.

_dbContext - used to execure SQL WHERE IN query.

If it makes harder to maintain - we can move sql queries to some one file like it is done with cache.

I think the goal is to make shop versatile and not easiest to maintain.
Easiest to maintain is one product shop. But nopcommerce with a LOT of its features is not the one by default.

Xzelsius: If there are 20 hardcoded SQL queries spread over 10 services and all for example are accessing Product and you for any reason change something related to that entity you probably have to change 20 SQL queries too. And thus have to remember where those queries are.

Instead of _dbContext.Set<TEntity> you should use IRepository<TEntity>.Table...

Regarding the dropdown problem with 20000 categories, how about a ajax powered search dropdown?
1000 times better than to know category ids for 20000 categories or to do a lookup anytime i want to map a category to a product.

ArturAlekseev: It's just a fast solution. Need to update my plugins first and I'm not sure these changes will be approved into nopcommerce. I don't need ajax search mostly because nearly all products are loaded/updated automatically.

ArturAlekseev: Instead of _dbContext.Set<TEntity> you should use IRepository<TEntity>.Table

Btw are there any real reasons that we should use IRepository instead of IDbContext?

Xzelsius: > Btw are there any real reasons that we should use IRepository instead of IDbContext?

**Except that nopCommerce does it this way**, i think the only reason will be plugins. A plugin could replace the repository for lets say UrlRecord and provide a custom implementation for storing them, an thus IDbContext could lead to an error.

AndreiMaz:
made a little bit simpler way

PR #2760, opened 2017-12-02
Thumbnails table reference

ArturAlekseev:
This is an ultra fast way to get thumbnails urls.
We add table that contains pictures ids and their thumbnail file names.
This way we easily generate urls without accessing file system. We easily delete thumbnails without searching files using wildcards on file system (that is terribly slow when there are a lot of thumbnails and killing hdd).

Code to generate table:
```
SET ANSI_NULLS ON
GO

SET QUOTED_IDENTIFIER ON
GO

CREATE TABLE [dbo].[Thumb_Mapping](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[PictureId] [int] NOT NULL,
	[ThumbSize] [int] NOT NULL,
	[FileName] [varchar](max) NOT NULL,
 CONSTRAINT [PK_Thumb_Mapping] PRIMARY KEY CLUSTERED 
(
	[Id] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
GO
```

We also need to create non clustered non unique index for PictureId with included column ThumbSize.

### DIFF
--- a/src/Libraries/Nop.Core/Domain/Media/Thumbnail.cs
+++ b/src/Libraries/Nop.Core/Domain/Media/Thumbnail.cs
@@ -1,0 +1,9 @@
+﻿namespace Nop.Core.Domain.Media
+{
+    public partial class Thumbnail : BaseEntity
+    {
+        public int PictureId { get; set; }
+        public int ThumbSize { get; set; }
+        public string FileName { get; set; }
+    }
+}
--- a/src/Libraries/Nop.Data/Mapping/Media/ThumbnailMap.cs
+++ b/src/Libraries/Nop.Data/Mapping/Media/ThumbnailMap.cs
@@ -1,0 +1,16 @@
+﻿using Nop.Core.Domain.Media;
+
+namespace Nop.Data.Mapping.Media
+{
+    public partial class ThumbnailMap : NopEntityTypeConfiguration<Thumbnail>
+    {
+        public ThumbnailMap()
+        {
+            this.ToTable("Thumb_Mapping");
+            this.HasKey(p => p.Id);
+            this.Property(p => p.PictureId);
+            this.Property(p => p.ThumbSize);
+            this.Property(p => p.FileName);
+        }
+    }
+}
--- a/src/Libraries/Nop.Services/Media/AzurePictureService.cs
+++ b/src/Libraries/Nop.Services/Media/AzurePictureService.cs
@@ -76,7 +76,8 @@ namespace Nop.Services.Media
             MediaSettings mediaSettings,
             NopConfig config,
             IDataProvider dataProvider,
-            IHostingEnvironment hostingEnvironment)
+            IHostingEnvironment hostingEnvironment,
+            IRepository<Thumbnail> thumbRepository)
             : base(pictureRepository,
                 productPictureRepository,
                 settingService,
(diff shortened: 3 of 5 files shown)


skoshelev:
@ArturAlekseev Thank you for your work, but we don't plan to implement this changes in near future.

ArturAlekseev:
Interesting to know why. Not that I just want my solution to be merged.
This is a terrible bug in implementation where each time HDD is stressed for searching thumbnail files.

skoshelev:
@ArturAlekseev First of all, this is not a bug because the file system doesn't work much in this case, all operating systems have index features, which increase file operations performance. You suggest us to add a lot of queries to the database, we think this can lead to a significant site performance drop during the times when it is actively used.

ArturAlekseev:
Well it is wrong.
1. "file system doesn't work much in this case" - wrong because original code:
`var currentFiles = System.IO.Directory.GetFiles(thumbDirectoryPath, filter, SearchOption.AllDirectories);`
is killing HDD when there are a lot of pictures. Maybe 1000 is ok. But 200000 ...
It is done in function where delete picture thumbs occurs.
Now look.
Imagine we have added 10000 new products in database. And their pics are marked "IsNew". Now google bot or yandex box is starting to analize shop.
And we get 10000 requests to file system to search for files using GetFiles function.
With a lot of pictures in pics folder you get 10 seconds to generate page instead of 1 second. This is not thoughts. I'v experienced it. And it is not the result of slow server.
Ofcourse you can write external code that will generate thumbnails for all products during products add. But why generate all if you can just fix shops code...

2. Also you are checking HDD file existance for each thumbnail that should be shown during page request. 20 products on page and you get 20 checks per page. Now think that google or yandex bot is working on site. So many checks... And check for record in database is much faster than check for file existance in file system. Not to mention non ssd hdds workflow...

That is why i'v made this change to my shop. After that my shop is working extremely fast.

skoshelev:
@ArturAlekseev Ok. We'll check your code for database performance, also we'll check our core code too. Thank you for your help.

skoshelev:
@ArturAlekseev I admit that you are right that in a certain situation our approach can load a hard disk too much, however, your solution to this problem, given its low probability, seems to me superfluous. Yet, thank you for your wish to contribute to our product. We have created [a task](https://github.com/nopSolutions/nopCommerce/issues/2983) based on your proposal. A couple of points if you want to help us further: we don’t use the ExecuteSqlCommand function to execute simple queries, as this make us inflexible, we document all public methods, we use standard rules and camel notation for parameter naming.

ArturAlekseev:
Will it be added to future releases? Or I should write a plugin?

arzybek19:
> Will it be added to future releases? Or I should write a plugin?

Hi 
have you ever made plugin? If not can you give us some instructions how implement your code. Thanks in advance

ArturAlekseev:
I will make a plugin soon. It will be free.

Thou it has one problem so far I experience on site with many pictures updates.

I don’t know how to fix it. Some thumbnails do not generate on first request.

My code do not interfere with thumbnails generation so I don’t know why it happens.

PR #2761, opened 2017-12-02
Campaigns - make urls absolute

ArturAlekseev:
This issue existed in v3.8. Looks like it is still in v4.0.
When we send campaign emails with images their urls are relative (do not contain reference to domain). This way customer do not see images in emails.
These changes urls to full in send test email and in send mass email.


RomanovM:
Thanks for your contribution. 
It's not only for campaigns but for all sent emails. Moreover it is better to solve this problem for the TinyMCE editor, and not to use workarounds. Here is a new [work item](https://github.com/nopSolutions/nopCommerce/issues/2789) for it.

PR #2804, opened 2017-12-28
Product Attribute Combination Pictures: Each product (item for sale i…

mplummer:
…n the store e.g. Sunglasses) had its own picture, however, each product attribute combination (any possible combination of attributes that the product may have e.g. Frame & Lens Color –> Metal & Blue) would be forced to share the products default image. There are Attribute Images in NopCommerce, but they do not cater for the full permutation. For example, above you could see the different Sunglasses Frames, but each with only one lens color; or each lens color but with only one frame.

We have added a Picture per Product Attribute Combination on the Admin Side, and made the Store Front show this combination if selected (otherwise it defaults back to the previous logic), as well as in the Order List and Shopping Cart.

### DIFF
--- a/src/Libraries/Nop.Core/Domain/Catalog/ProductAttributeCombination.cs
+++ b/src/Libraries/Nop.Core/Domain/Catalog/ProductAttributeCombination.cs
@@ -6,6 +6,11 @@ namespace Nop.Core.Domain.Catalog
     public partial class ProductAttributeCombination : BaseEntity
     {
         /// <summary>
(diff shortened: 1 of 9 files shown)


RomanovM:
@mplummer Hi.
We have slightly modified your code and have implemented this feature. Now you can set images per attribute combination. Changes are [here](https://github.com/nopSolutions/nopCommerce/commit/71ef1e97b30b26c5005f13f7fae22fe880740c43).
Thanks for the suggestion and contribution.

PR #2851, opened 2018-01-26
update Info view - add localize for birth day

MJ-Vakili:
For languages other than English, the date of birth will be displayed in the local info customer form.


AndreiMaz:
@MJ-Vakili Thanks a lot!

PR #2854, opened 2018-01-29
Update SitemapGenerator.cs to use gregorian date in sitemaps

ni-fara:
Currently sitemap uses culture specific dates for generating modification dates of products, which leads to dates like "2561-01-29" for cultures which their default calendar is something other than gregorian calendar. This update fixes this issue in sitemaps' <lastmod> fields.


AndreiMaz:
@ni-fara Thanks a lot

PR #2936, opened 2018-03-28
Allowing unicode characters in picture file names

ni-fara:
Image file name is one of the factors of image SEO improvements. Currently picture files get the product name which is good for site admins who enter product names in english. But for languages which use unicode characters for product names, picture file names changes to "-". So what you got is some meaning less file names both to search engines and humans. This update corrects this functionality.


RomanovM:
We need to consider this issue in more detail. Here is a work item https://github.com/nopSolutions/nopCommerce/issues/2957. 
Thanks for suggestion

PR #2937, opened 2018-03-28
A little performance improvment

ni-fara:
The edited GetSeName method is called in other methods, some of them are during saving process of admin pages and some of them occur during user interactions like viewing tags. The ToCharArray method is called for every character of the name string, this update moves the ToCharArray method out of foreach loop.


RomanovM:
There is no need for such code changes. But anyway thanks for the suggestion.

PR #2947, opened 2018-04-06
Add paging css class names as fields

halityurttas:
Paging css classes occur big problem for integrate some html templates. But accomplish if we make changes css class names.


DmitriyKulagin:
@halityurttas Thanks a lot!

PR #2977, opened 2018-04-29
order for script tag helper

Peymanpn:
hi guys.
i have added Show Order to script tag helper so some one may need to load scripts in an order.

in some cases you may want to load some scripts before another and each tag may be requested from any parts of the page.
for example you may want to load jquery or it's plugins before another scripts.
also it works for both Script files and inline scripts.
i hope it will be usefull.
examples:
`
Html.AppendScriptParts(ResourceLocation.Footer, "~/js/public.common.js", order:"1");
`
or
`
<script src="~/js/admin.common.js" asp-location="Footer" asp-order="2"></script>
`


AndreiMaz:
@mariannk что скажешь?

mariannk:
@Peymanpn hi.
Thank you very much for your pull request! 
Unfortunately we decided not to implement this functionality out of the box.

PR #2980, opened 2018-05-02
Filter Shipping By Weight Records

wsypooh:
Filter Shipping By Weight records by store, warehouse, shipping method, country, state and zip. This will make it easier to manage more than 50 records.

Developer notes:
- Added collection in Configuration Model
- Added UI and script in ByWeightByTotal.cshtml
- Instead of calling GetAll, created Search method based on the filters selected.

I did some unit tests and it seems to be working.

filter shipping by weight records


skoshelev:
@wsypooh Thank you for your work, however, I have a couple of remarks: 
1. Function SearchShippingByWeightRecords does almost the same as FindRecord function does, you should do something, for example, extract the same functional into protected method, or remove your function and make changes in ours.
2. You should separate search panel from the list, like in Product/List view

wsypooh:
@skoshelev Thank you for your feedback. Will make the necessary changes and will let you know once updated.

wsypooh:
@skoshelev I committed the updates based on your suggestions.
1. Created FindRecord that returns IPagedList but uses the same functional code of the original method.
2. Separated search panel from the list.

Let me know if you find any issues. Thank you.

skoshelev:
@wsypooh Thank you for your help

wsypooh:
You are welcome! Thank you as well.

PR #3034, opened 2018-06-13
Cache Busting for JS and CSS files/bundles

bfranklin825:
Added version strings to .js and .css files so browsers will know when to pull updated ones. The version number (i.e. "?v=9383849332") is based off of the last write time for the .js or .css file. This will only change the version when the file's contents have been changed. 

Also the bundle name that is generated is based off of the last write time of each file included in the bundle, so the name for the bundled file will be different whenever any file in the bundle has changed.


src/Presentation/Nop.Web.Framework/UI/PageHeadBuilder.cs
```diff
@@ -626,13 +627,48 @@ public virtual string GenerateCssFiles(IUrlHelper urlHelper, ResourceLocation lo
...
+        /// keeps the browser from using an older cached version of the file when it should use a newer one
+        /// </summary>
+        /// <param name="url">The relative URL of the file to get the version for.</param>
+        /// <param name="includeVersionEquals">Should we include a version identifier?</param>
+        public virtual string GetVersion(string url, bool includeVersionEquals = true)
+        {
+            // Get the roots
+            var webRoot = _hostingEnvironment.WebRootPath + Path.DirectorySeparatorChar;
+            var contentRoot = _hostingEnvironment.ContentRootPath + Path.DirectorySeparatorChar;
+            var versionEquals = includeVersionEquals ? "?v=" : "";
```
DmitriyKulagin: Use string.Empty rather than "".

bfranklin825:
Not really sure there is a difference between "" and string.Empty, other
than the fact that one is longer to type.

As Microsoft explains here:
https://msdn.microsoft.com/en-us/library/system.string.empty.aspx

They state: The value of this field is the zero-length string, "".

So the value of string.Empty is "".

DmitriyKulagin:
@bfranklin825 Thanks a lot!

bfranklin825:
@DmitriyKulagin You're welcome!

Issue #3035, opened 2018-06-13
Missing package (RhinoMocks) in Nop.Tests


aruss:
nopCommerce version: 4.10

Steps to reproduce the problem: 
Clone the repository and try to build, Nop.Tests will fail because RhinoMocks package is not installed.

skoshelev:
RhinoMocks library is obsolete and doesn't support .Net Core, we decided to use Moq library instead RhinoMocks

PR #3036, opened 2018-06-13
Added missing RhinoMocks package

aruss:
Just added missing nuget package, otherwise solution does not compile after a fresh clone

issue: #3035


skoshelev:
RhinoMocks library is obsolete and doesn't support .Net Core, we decided to use Moq library instead RhinoMocks

PR #3057, opened 2018-06-26
Update EfRepository.cs

JadynWong:
only update()  no tracking entity

https://github.com/aspnet/EntityFrameworkCore/blob/e300fdfe4ac314cc9e65520a15b9b7609a6c6a21/src/EFCore/DbSet%60.cs#L202

### DIFF
--- a/src/Libraries/Nop.Data/EfRepository.cs
+++ b/src/Libraries/Nop.Data/EfRepository.cs
@@ -119,7 +119,10 @@ namespace Nop.Data
 
             try
             {
-                Entities.Update(entity);
+                if (!Entities.Local.Any(x => x.Id.Equals(entity.Id)))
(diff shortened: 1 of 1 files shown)


AndreiMaz:
@JadynWong could you please clarify what it actually does. Not sure that I understand this logic

JadynWong:
@AndreiMaz  As dbset<T>.update()'s describtion in the note
` All properties of the entity will be marked as modified.`
If you use dbset<T>.update() ,it will udpate all attribute .
Ef core defaults to updating properties of tracked entity modifications.
I think the method should use,when the entity no track

JadynWong:
sorry.I just saw `CONTRIBUTING.md`
Should I close this PR?.

Issue #3086, opened 2018-07-05
Replace js "Confirm" to Nop-action-confirmation window for Admin/Order/AddShipment


DmitriyKulagin:
The task is to rewrite the logic of displaying the additional message so that it is calculated on the model, not in the javascript. Thus, already at the model level, it will be decided which confirmation window to display on the page.
In the event model, we can not track the negative decision for confirmation, which is now generated in the "confirm ()".
It is necessary to use "asp-additional-confirm".

justindongqiang:
@DmitriyKulagin I will take a look at this issue and I will refer to your  today's commit.

justindongqiang:
@DmitriyKulagin I am pretty much done with the changes and just wanted to make sure that if we need to have confirmation window for both Save and SaveAndContinue events.

DmitriyKulagin:
@justindongqiang Yes it is necessary for both events.

justindongqiang:
@DmitriyKulagin I checked in additional changes to make sure confirmation window only opens when meets the condition of warehouse quantity. Please review and let me know if you have any questions.

1. As soon as you bind button id with nop-action-confirmation tag helper, the button click would open confirmation window always. This works fine for pretty much all the case, such as a Deletion button, but no in this case.

2. The confirmation window has to open when a condition met. Otherwise, it won't display confirm window. I have added workaround code to resolve this case. Basically I have looked what nop-action-confirmation did, and removed its auto generated code by tag helper on the fly, in focus event, it reset button status so it wont call confirmation window method.

3. I made a little more changes to make the code neat and dry, like jquery chaining and self 'this'.  

4. The script was included in the if statement "if (itemsFromMultipleWarehouses.Any())", I will need to remove it and make a little adjustment to sb sting to return a true if there is no items. The reason to remove line of 'if (itemsFromMultipleWarehouses.Any())' is because, in case there is no warehouse settings, and the entire script inside if statement won't happen, and the script I wrote to remove confirmation tag helper won't happen either,  the confirmation window should not appear, but we already bind the save button with nop-action-confirmation, this is causing confirmation window open where we donot want it open. I hope I made it a little clear.

Let me know if you have any questions.
 
Thanks,
Justin

justindongqiang:
Any updates? Andrei. Perhaps we could get this in and refactoring tag helper later.

DmitriyKulagin:
Closed #3127

PR #3127, opened 2018-07-10
#3086 Replace js "Confirm" to Nop-action-confirmation window for Admi…

justindongqiang:
@DmitriyKulagin  Please review and let me know your thought on "3" as follows. 

There are a couple of questions here:

1. Does this confirmation window look OK? It came from "Admin.Orders.Shipments.Products.Warehouse.QuantityNotEnough" and confirmation text has to come from localized resource as it is specifically for warehouse quantity computation. So I'd need to pass it into asp-additional-confirm.

2. Since we have both button trigger confirmation, I had to pass a button value to confirmation method in order to to differentiate between Save button and SaveAndContinue button, otherwise the confirmation windows would open twice.

3. The confirmation should only appear if meets warehouse quantity condition. Now that we have "<nop-action-confirmation asp-button-id="shipment-save">" in place and it causes the confirmation window to open all the time even if there is no warehouse selected. I thought a couple of ways doing this, like having a hidden button to act as a save button or adding an if condition to warp <nop-action-confirmation> it out. I would like to hear your suggestions so I can move toward the right direction.  

image

### DIFF
--- a/src/Presentation/Nop.Web/Areas/Admin/Views/Order/AddShipment.cshtml
+++ b/src/Presentation/Nop.Web/Areas/Admin/Views/Order/AddShipment.cshtml
@@ -31,6 +31,8 @@
                 <i class="fa fa-floppy-o"></i>
                 @T("Admin.Common.SaveContinue")
             </button>
+            <nop-action-confirmation asp-button-id="shipment-save" asp-action="AddShipment" asp-additional-confirm='@T("Admin.Orders.Shipments.Products.Warehouse.QuantityNotEnough")' />
+            <nop-action-confirmation asp-button-id="shipment-save-continue" asp-additional-confirm='@T("Admin.Orders.Shipments.Products.Warehouse.QuantityNotEnough")'/>
             @await Component.InvokeAsync("AdminWidget", new { widgetZone = AdminWidgetZones.OrderShipmentAddButtons })
         </div>
     </div>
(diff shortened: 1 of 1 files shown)


src/Presentation/Nop.Web/Areas/Admin/Views/Order/AddShipment.cshtml
```diff
@@ -62,15 +64,15 @@
...
                                 });
-                                $('#shipment-save-continue').click(function() {
-                                    return validateWarehouseAvailability();
+                                $("#shipment-save-continue-action-confirmation-submit-button").bind("click", function () {
+                                    return validateWarehouseAvailability("continue");
                                 });
                             });
 
-                            function validateWarehouseAvailability() {
+                            function validateWarehouseAvailability(saveOrContinue) {
```
DmitriyKulagin: "saveOrContinue" rename to  "action"

DmitriyKulagin:
@justindongqiang 
Hi.
The confirmation window looks correct.
The window should not always appear. Its behavior should not change in comparison with the previous version. A sentence with the addition of a condition I like more, because code in this case should get clearer.

src/Presentation/Nop.Web/Areas/Admin/Views/Order/AddShipment.cshtml
justindongqiang: ok will do

justindongqiang:
@DmitriyKulagin Wll hidden buttons wont work but I found a workaround that I am working on now. Will let you know when I finish testing and I want to make sure it works without any issues.

justindongqiang:
Any updates? Perhaps we could get this in first and refactoring confirmation tag helper lateral.

DmitriyKulagin:
@justindongqiang 
Justin, 

I analyzed the code, it looks good. But there remained a small inaccuracy that needs to be corrected. Formatting of the text in the message was lost. See how it is done elsewhere.

image

justindongqiang:
@DmitriyKulagin okay it is fixed. Thanks! Let me know if you any questions.

Issue #3092, opened 2018-07-07
Cache Busting for JS and CSS files


AndreiMaz:
We had this pull request https://github.com/nopSolutions/nopCommerce/pull/3034 

but it's better to re-implement it:
1. Using built-n feature https://github.com/aspnet/Mvc/blob/release/2.1/src/Microsoft.AspNetCore.Mvc.TagHelpers/ScriptTagHelper.cs#L40
2. Do not use System.IO.Path.
3. Make it ocnfigurable in admin area so a store owner can disable it (there could be permission issues on some hosting)

DmitriyKulagin:
After some discussion, it was decided not to implement it.

RomanovM:
Moved to #4791

PR #3129, opened 2018-07-10
#3092 Cache Busting for JS and CSS files (Enhancement/Feature)

bfranklin825:
Added 4 admin settings under SEO settings (that is where bundling and minification is so I figured that was a good place to put them). 

- EnableCacheBusting (for wwwroot files)
- EnableCacheBustingForContent (for content files- i.e. plugins)
- EnableCacheBustingForJavascript (for .js files)
- EnableCacheBustingForCss (for .css files)

I figured someone might want to flexibility as far as whether or not they want to do cache busting for .js or .css or content or wwwroot... so I added the ability to pick and choose. They are all false by default.

Now PageHeaderBuilder does not use System.IO to get the last write time of file, instead it uses a FileVersionProvider to get a version for the files, like the TagHelper uses, but it will change the provider based on where the file is located - wwwroot or content folders. If we just use the TagHelper it would only version the files located in the wwwroot folder because it uses the WebRootFileProvider by default. So I'm switching the provider dynamically so all files can be versioned.


RomanovM:
Thanks. Based on the proposed pull request (together with #5240), I created a separate [branch](https://github.com/nopSolutions/nopCommerce/tree/issue-3092-files-cache-busting), perhaps we'll reopen and continue working on the issue #3092.

Issue #3216, opened 2018-08-06
Linq Query for getting cart should be it's own method.


taylorglenn:
nopCommerce version: 4.2

Steps to reproduce the problem: 
The linq query in the ShoppingCartController class is copied/pasted 17 times.  It's cleaner to create a method for getting the cart using this query in the ShoppingCartService class.   I'd love to make this very easy improvemnet to the codebase.

taylorglenn:
Hi!  Do you mind if I ask why this was immediately placed on hold?  Just curious.

JohnsonJohnsonYuan:
@taylorglenn , why GetShoppingCart(Customer customer, ShoppingCartType shoppingCartType, int storeId, bool ordered = false) function add ordered parameter and order by id descending by default?

```c#
 if (ordered)
                shoppingCart.OrderByDescending(sci => sci.Id);
```

if I want to order by created time, but this parameter doesn't work.

taylorglenn:
You're right.  It doesn't make sense to have the method to get the cart also sort the cart.  I've removed that functionality.  If someone wants to sort their cart, they can do this and have much more freedom:
`var cart = ShoppingCartService.GetShoppingCart(customer, shoppingCartType, storeId).OrderByDescending(sci => sci.<whatever-property-you-want-to-sort-by>).ToList()`

PR #3217, opened 2018-08-06
Issue 3216 Linq Query for getting cart should be it's own method.

taylorglenn:
Created a new method for getting the cart in ShoppingCartService class. Replaced all instances of LINQ query used to get the cart in ShoppingCartController class with the new method.


RuS984:
Closed #3216

PR #3278, opened 2018-09-09
word corrections - portuguese (Brazil)

alessandromontividiu03:
changed some wrong words in portuguese (Brazil) to their correct words.


AndreiMaz:
@alessandromontividiu03 Thanks a lot!

PR #3355, opened 2018-10-05
issue 3332, added new property to export specification attribute for …

kem-storis:
Added specification attribute option value to ExportSpecificationAttribute, added reference to additional property in ExportManager to pull that value and include into the export spreadsheet data for a product


kem-storis:
@RomanovM: just wondering if this will be merged into the next release or what is holding up this pull request.  Thanks!

skoshelev:
Hi @kem-storis. Sorry, this is my fault, I'll try to check your code next week

kem-storis:
Thank you, I appreciate it.

-Ken

src/Libraries/Nop.Services/ExportImport/ExportManager.cs
```diff
@@ -545,6 +548,7 @@ private int ExportSpecificationAttributes(Product item, PropertyManager<ExportSp
                     ShowOnProductPage = psa.ShowOnProductPage,
                     DisplayOrder = psa.DisplayOrder,
                     SpecificationAttributeOptionId = psa.SpecificationAttributeOptionId,
+                    SpecificationAttributeOptionValue = allSpecificationAttributes.FirstOrDefault(sao => sao.SpecificationAttributeOptionId == psa.SpecificationAttributeOptionId).SpecificationAttributeOption.Name,
```
skoshelev: You may use psa.SpecificationAttributeOption?.Name insted your code

skoshelev:
Hi @kem-storis. 

Thank you for your help, but you did only one part of changes, without importing it's not have a sense. Also you may see how we're exporting/importing the productType field (https://github.com/nopSolutions/nopCommerce/blob/develop/src/Libraries/Nop.Services/ExportImport/ExportManager.cs#L1170) for SpecificationAttribute we need the same way.

kem-storis:
@skoshelev, we were only concerned with getting the attribute value in an export.  In the case of an import, if the id and value of the attribute do not match, would the value be used to locate the existing attribute or the id?  Which one would "win"?  Why are we being required to also add the import feature when we are only concerned with the export and receiving the value of an attribute which has already been associated to a product?

skoshelev:
Hi @kem-storis. As I said before, we changed the mechanism for importing related entities, and now we use either identifiers or names. Regarding the need to implement a complete set of functions on your part, you are not obliged to do this, but we cannot accept partial functionality of a PR. If you want to help us become better and add functionality interesting for you to the core, please implement full support for importing and exporting a new field.

kem-storis:
Thank you @skoshelev, I would much appreciate your further input regarding how to handle using two qualifications for matching an input value for product import. Specifically, in this case, I have been unable to move forward as I'm not sure which case would be the Nop standard, as importing both the specification attribute id and value creates the need for additional business rules:

If id and value do not match:
1. Would it be considered not a match?
2. Would Id win over value or value win over Id?
3. In the case of a value, as it is type string, would there be any additional restrictions/requirements for a match?

Thank you,
Ken

skoshelev:
Hi @kem-storis. Our approach to the import and export of related entities clearly indicates what we should be used to exporting: names or identifier, and we never use both methods for the same entity at the same time, and the import algorithm should handle both possible situations depending on the content of the data file

Issue #3353, opened 2018-10-04
"Notify for qty below" setting in product details doesn't work


ni-fara:
nopCommerce version: 4.1, 4.0

Problem: The "Notify for qty below" setting in product details doesn't work. It doesn't notify the store owner for stock quantities below the specified value. It only notifies the store owner when the stock quantity reaches number zero.

What I've done to produce the problem:
1- Create a product of type simple.
2-In "Inventory method" setting choose "Track inventory".
3- Set "Stock quantity" to number 6
4- Set "Notify for qty below" to number 4
5- Then I reduce the quantity by buying the product (each time only one item is bought)
6- No notifications are displayed in dashboard in "Common statistics" section when the product quantity reaches number 4 or 3.
7- Continue buying the product, and you can see no notifications are displayed until product quantity is equal to zero.

Edit: I found that "Notify for qty below" setting was completely ignored for filtering the related result set in source code and other values were used instead of it. So i made pull request #3356  to fix the issue.

ni-fara:
So "MinStockQuantity" propery is used for both "Low stock activity" and "Low stock report", but consider this situation that the store owner wants to see a product in low stock report  when the stock quantity reaches 5, and wants to disable the buy button when the stock quantity is zero. Currently it is impossible to do this. Two different needs are tight together.
 With changes of pull request #3356 both "low stock activity" and "Low stock report" work fine using the existing properties and codes for both simple products and combinations. Then we can add a new bool value to distinguish if the store owner want's to be notified with email too or not.

holydk:
Closed #3353

PR #3356, opened 2018-10-05
Fixing the bug: "Notify for qty below" setting in product details doesn't work

ni-fara:
The "Notify for qty below" setting in product details doesn't work. It doesn't notify the store owner for stock quantities below the specified value. This is because wrong values are used in queries instead of using "Notify for qty below" setting.
Queries get corrected to use the "Notify for qty below" setting (NotifyAdminForQuantityBelow) for filtering the result sets.
Please refer to issue #3353 for more information.


ni-fara:
Here [https://github.com/nopSolutions/nopCommerce/issues/3353#issuecomment-427925056](url) i suggested using this change and adding a bool value to deal with email notification.

Issue #3380, opened 2018-10-16
Product tags count ignores ACL


jamiepittman:
nopCommerce version: 4.00

Product Tags Count shows same count for all customer roles regardless of the number of products each respective customer role has access to

Steps to reproduce the problem: 
Create 2 products each with the same tag like _'Computer'_
Update one of the products to allow _'Administrator'_ as the only Customer Roles under the Access Control List
Navigate to the product page for the product you created that all users can access as both an Administrator and a guest.
  At the bottom of this page the Product Tags section is rendered with the product tags count next to it which will show **(2)** for all roles because it is calculated regardless of the Customer Role.
  **If you are viewing this as an Administrator the count is correct.** 
  **If you are viewing this as a guest then the count is incorrect and reflects the total including products your current customer role 'guest' cannot access.**

frederickrogan:
I am happy to pick this up. I have got a first attempt that works but want to refactor before submitting a pull request. 

It will require a new/updated stored procedure. Will I need to write something to handle migrations for this?

PR #3387, opened 2018-10-23
#3380 ACL consideration when viewing product counts by tag

frederickrogan:
Product pages have a series of links for viewing the catalogue with a tag filter. These links display the count for the number of items in the catalogue with that tag.

Currently a product might not be visible to the user (because of their role) but will still add towards this count for a tag.

This PR adds extends the counts calculation with an additional stored procedure that queries AclRecords (access control list) if enabled.


AndreiMaz:
Thanks a lot! We'll fix it soon with some additional checks

PR #3392, opened 2018-10-26
 Path Separator Fix for Non-Windows Systems

incetarik:
Path Separator Fix for Non-Windows Systems  …
Path separator was `\` for Unix like systems, changed to `/`.


RomanovM:
Thanks for your suggestion. 
We have more global changes to use the project in Unix like systems, including working with the file system. [See ](https://github.com/nopSolutions/nopCommerce/tree/issue-3095-Run-nopCommerce-on-Linux)if it meets your requirements?

Issue #2287, opened 2017-04-03
Async methods (performance)


DarthSonic:
Based on discussion http://www.nopcommerce.com/boards/t/41947/performance-testing-nopcommerce-370-on-azure-using-redis-blob-azure-sql.aspx here  are some suggestions for better performance in Azure enviroment (and even outside Azure)

also see  http://www.nopcommerce.com/boards/t/46712/nopcommerce-400-roadmap-and-estimated-release-date-lets-discuss.aspx?p=2#185433

We have to introduce more async methods (actually covering all request) - controllers, services, database calls, etc

lintho:
Good work. I hope the team will take this seriously!

chadwixk:
There's no way to Vote up in Git...but using a comment as such. Hope the Nop Team evaluates this seriously ...and soon!

almarax:
I'd be more specific in p.4 - do not query DB multiple times when you can easily use your own optimization improvement: 
QueryableExtensions.IncludeProperties<T>() 
to preload all related data in the first query, This will significantly lower used DTUs in Azure and make pages much more responsible, especially considering latency of every query on lower Azure plans.

martingust:
Great, I think this is a deal breaker for a lot of people.

pdunleavy:
Please can we add async calls to blob storage (AzurePictureService) this looks to be using a static reference which is torn down on every request, not sure of the internals of the storage api but getpicturethumb gets a lot of hammering and this could cause a scalability bottleneck if requests are being queued synchronously to access the container.

pdunleavy:
The category listing page uses a stored procedure to retrieve the paged list of products, I need more than QueryableExtensions.IncludeProperties() to include product picture records almarax. This would need refactoring to be multi result set...  http://stackoverflow.com/questions/10314032/entity-framework-5-stored-procedure-with-multiple-result-sets

pdunleavy:
Just want to say, we have now moved our site to a single Azure VM with the same specification. All pages now load 1 second faster than web app with sql azure. If this work item can be done we can consider moving back.

lintho:
Is there any updates on this? We host in azure, and we need better performance.

AndreiMaz:
Please sorry and ignore that it's in Russian (only for internal use)

Optimization. предлагаю обсудить, что мы можем сделать для версии 4.0
[...]

PR #3404, opened 2018-11-09
Async operations wrapped in a Sync way.

Leftyx:
I've changed the implementation for the sync operations calling async methods.

All the async method should always be awaited. It is not always the case in NopCommerce.

`async void` should always be [avoided](http://blog.stephencleary.com/2013/01/async-oop-2-constructors.html#what-not-to-do):

> At first glance, this seems like a reasonable approach: you get a regular constructor that kicks off an asynchronous operation; however, there are several drawbacks that are due to the use of async void.
> 
> The first problem is that when the constructor completes, the instance is still being asynchronously initialized, and there isn’t an obvious way to determine when the asynchronous initialization has completed.

Another [article](https://johnthiriet.com/removing-async-void/) worth reading on the subject.

And [this](https://cpratt.co/async-tips-tricks/) is a must read on different option when dealing with async operations in a sync methods.

I've included an helper `AsyncHelper `(with links to the source) which I've used in the past and it has worked pretty well. It is a bit expensive in terms of resource though.

It could be used like this:

`AsyncHelper.RunSync(() => this.RemoveByPatternAsync(pattern));`

There's never a real solutions to the [problem](https://stackoverflow.com/a/40344759/219406). The only option is to convert everything to Async. And I guess it make a lot of sense in ASP.NET CORE.

### DIFF
--- a/src/Libraries/Nop.Core/Caching/RedisCacheManager.cs
+++ b/src/Libraries/Nop.Core/Caching/RedisCacheManager.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Nop.Core.Configuration;
+using Nop.Core.Infrastructure.Threading;
 using StackExchange.Redis;
 
 namespace Nop.Core.Caching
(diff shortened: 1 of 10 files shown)


erikwuu:
+1 for converting everything to async

skoshelev:
Hi @Leftyx, thank you for your help, but we decided to leave all the issues with asynchronous until the implement of [this ticket](https://github.com/nopSolutions/nopCommerce/issues/2287), now it was decided to minimize asynchronous calls and change them to the synchronous

Issue #2743, opened 2017-11-26
"Form value count limit 1024 exceeded" errors


AndreiMaz:
I am new to nopCommerce. I downloaded and installed version 4.0 nopCommerce. While in configuration, under payment restrictions, I attempted to restrict all payment methods in all countries except US. When I clicked save, the system crashed and gave me the below error exception in 2017 Visual Studio Community.

The error occurred in the ApplicationBuilderExtensions.cs

Line: 81

The error message:  System.IO.InvalidDataException: 'Form value count limit 1024 exceeded.'

System.IO.InvalidDataException occurred
  HResult=0x80131501
  Message=Form value count limit 1024 exceeded.
  Source=Nop.Web.Framework
  StackTrace:
   at Nop.Web.Framework.Infrastructure.Extensions.ApplicationBuilderExtensions.<>c.<UseNopExceptionHandler>b__1_1(HttpContext context) in C:\Users\jimth_000\Source\repos\NopCommerce\Presentation\Nop.Web.Framework\Infrastructure\Extensions\ApplicationBuilderExtensions.cs:line 81
   at Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware.<Invoke>d__6.MoveNext()

I tried checking off one column at a time. The error occurs after 4 columns are saved. Even after restarting the application this occurs. I have also uninstalled and reinstalled the application several times and this error still occurs.

Can anyone please provide some guidance on why this happens and how to prevent this?

Source: https://www.nopcommerce.com/boards/t/49611/version-40-error-on-payment-restrictions-configuration.aspx

AndreiMaz:
See https://stackoverflow.com/questions/43305220/form-key-or-value-length-limit-2048-exceeded snd https://stackoverflow.com/questions/38357108/form-submit-resulting-in-invaliddataexception-form-value-count-limit-1024-exce

AndreiMaz:
tiket: Overriding Request Form max upload limit
commit: https://github.com/aspnet/Mvc/commit/23b7d8f62a2515e2dbea436ef118fb67e759e2d9#diff-366dd01d9e3d472a3fdd73c793f72db7

skoshelev:
Closed #2743

PR #3412, opened 2018-11-13
Form value count limit 2048 exceeded

DiogenesPolanco:
Filter to set size limits for request form data


skoshelev:
Hi @DiogenesPolanco, thank you for pull request. But In the module Microsoft.AspNetCore.Mvc already exist necessary attribute RequestFormLimits, and we used it to fixed this #2743 tiket, see this [commit](https://github.com/nopSolutions/nopCommerce/commit/01000e674f254ef8a336571354ec610d29adbe07). But we won't to do this for all method in all controlers, we think it not safe.

PR #3413, opened 2018-11-13
Add or update plugin locale resources with dictionary

DiogenesPolanco:
var Resources = new Dictionary<string, string>()
            {
                { "Plugins.Widgets.sample.Picture1", "Picture 1" },
                { "Plugins.Widgets.sample.name", "Test Name" }
            };
            _localizationService.AddOrUpdatePluginLocaleResources(Resources);


RomanovM:
We don't plan to implement it now. Anyway, thanks for your suggestion.

PR #3463, opened 2018-12-17
Small project file cleanup: move OutputPath to shared property group

Neme12:
This is just a small cleanup of the plugin project files. Since `OutputPath` is the same for debug & release, it can be moved up to the shared property group instead of having it duplicated. Visual Studio understand this just fine in the new project system.


skoshelev:
@Neme12 Thank you for your help.

PR #3484, opened 2019-01-09
Use latest .NET Core SDK version.

MaxMommersteeg:
Targeting `2.2` will select the latest version, which currently is `2.2.102`. Patch updates should never break the build so this should be fine.


skoshelev:
Hi @MaxMommersteeg. Thank you for suggesting. But we set the exact version of the framework to avoid compatibility problems during testing, this version corresponds to the one on which we compile the official release.

PR #3485, opened 2019-01-09
Remove Microsoft.NETCore.App reference from Nop.Core.

MaxMommersteeg:
As by Microsoft's recommendation we should not explicitly reference `Microsoft.NETCore.App` or `NETStandard.Library` when targeting .NET Core or .NET Standard.

----

> When targeting .NET Core or .NET Standard, never have an explicit reference to the Microsoft.NETCore.App or NetStandard.Library metapackages via a <PackageReference> item in your project file.

[MSDN documentation](https://docs.microsoft.com/nl-nl/dotnet/core/tools/csproj#recommendations)

### DIFF
--- a/src/Libraries/Nop.Core/Nop.Core.csproj
+++ b/src/Libraries/Nop.Core/Nop.Core.csproj
@@ -18,6 +18,7 @@
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="6.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection.Redis" Version="0.4.1" />
     <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.2.0" />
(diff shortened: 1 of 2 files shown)


MaxMommersteeg:
Was trying to add all individual packages earlier on but referencing `Microsoft.AspNetCore.All` makes this all a bit easier.

Leftyx:
**Deprecate Microsoft.AspNetCore.All** is [deprecated](https://github.com/aspnet/AspNetCore/issues/3418).
And this is another [link](Microsoft.AspNetCore.App) helpful explanation.

MaxMommersteeg:
Thanks, I did not realize that. Will see if I can correct the PR in a few hours if feasible.

RomanovM:
We are a little late in accepting this pr, there is already .NET 5. 
Anyway, many thanks for your contribution.

PR #3504, opened 2019-01-17
Fix feature 3386

MalcolmFord:
I added a simple check to see if the parent product id was equal to any of the new products that were being added. If the two matched, then I skipped over that product.


RomanovM:
Thanks. Resolved in https://github.com/nopSolutions/nopCommerce/commit/01f17b58a0b2021d9a4999b6c60d5a5c2d61c9d1

Issue #3095, opened 2018-07-07
Run nopCommerce on Linux


AndreiMaz:
Also see https://github.com/nopSolutions/nopCommerce/issues/415

VR-Architect:
Hi, I see Issue# 3144 was closed and referenced to this issue. Just wondering, is this issue implying that support for Linux is not currently available and on the roadmap? I need to make production decisions for the servers to support the business timelines today. If Linux is not currently supported, does anyone know when they expect to support it? Thanks to anyone who may know.

VR-Architect:
Received a [comment ](https://www.nopcommerce.com/boards/t/54054/nopcommerce-410-beta-released-please-share-your-impressions.aspx?p=6)stating Linux support is coming in v4.2

skoshelev:
Closed #3095

PR #3505, opened 2019-01-18
Update InstallationLocalizationService.cs

CV80:
If directory path contains '\',  Replace to '/'. 
Because on Linux, the path contains '\' will cause error, but on Windows it always well both '\' and '/'.


RuS984:
Hi. We have an other [ticket](https://github.com/nopSolutions/nopCommerce/issues/3095). 
We solved path generation problem. Please see this [commit](https://github.com/nopSolutions/nopCommerce/commit/d12e19883c959f913507c0ad95bc108dd10da663) 
Please see this [branch](https://github.com/nopSolutions/nopCommerce/commits/issue-3095-Run-nopCommerce-on-Linux), we fixed many problems with work on UNIX system.

CV80:
OK, nice work you done & much better.
I just concert on the Installation Error, but no more time to the FileProvider.MapPath() method reference by many places.

PR #3537, opened 2019-02-11
Fix for 3529 - Update PerRequestCacheManager.cs

toddca:
Here is a better fix for 3529 which uses the ReaderWriterLockSlim as recommended by @skoshelev


skoshelev:
@toddca Thank you for your help.

PR #3549, opened 2019-02-19
Demo at demo.nopcommerce is out of date

shanselman:
Demo at demo.nopcommerce is out of date, checking link to the newer https://www.nopcommerce.com/demo.aspx

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Browse our shopping cart demo store to get a feel for nopCommerce shopping cart.
 
 Front End | Admin area
 ----|------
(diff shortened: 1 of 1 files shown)


AndreiMaz:
Thanks a lot, Scott! But the link is valid.

Please let me clarify. We have two separate demo website. The first one for frontend (https://demo.nopcommerce.com/) and the second one for admin area (http://admin-demo.nopcommerce.com/)

But https://www.nopcommerce.com/demo.aspx is the informational page on our official site that contains links to both demo websites

shanselman:
Ah, gotcha. My bad. The Lumia pic through me since it's so old.

PR #3583, opened 2019-03-09
Removed CurrencyExchangeRateModel inheritance from BaseNopModel

charankasu:
Memorydump
This issue is happening due to a memory leak while binding the model. Please refer attached screenshot of the memory dump that shows that number of CurrencyExchangeRateModel objects in the memory while reproducing this issue. As per ASP.NET Core documentation model binding is taken care by the framework with an option of extending it with custom methods. I don't see any extensions for model binding in this case , so removed the inheritance of CurrencyExchangeRateModel. Tested the functionality (applied the exchange rate for a single currency as well as for all currencies (Apply All)) and it is working fine after this change


RomanovM:
Thanks for the suggestion. We have added a workaround for this issue. See details https://github.com/nopSolutions/nopCommerce/commit/86d17d5a47206a2921cbb5063c57c8f2a9097c39

Issue #3589, opened 2019-03-13
Stop products with custom AttributeXml from merging


justinjstark:
I am integrating nopCommerce with a 3rd-party product customizer via a plugin. The customizer returns a DesignId which I am adding to the attributeXml of the ShoppingCartItem, similar to gift cards. This works great but the problem I'm having is two products with different DesignIds are getting merged in the cart when I want them to remain as separate line items.

Option 1: I could create and register my own IProductAttributeParser and override AreProductAttributesEqual. But this method only considers product-mapping attributes, not custom attributes. This same method is also used when finding product attribute combinations, so any changes have the potential of breaking core product attribute functionality.

Option 2: I can do like what is being done for gift cards and add logic in ShoppingCartService.FindShoppingCartItemInTheCart. The issue here is I have to override this entire method and all of its logic the equality logic is contained within a loop over all shopping cart items. This loop has a lot of logic and I worry about keeping it up-to-date with nop upgrades.

My proposal:
I'm going to issue a pull request that extracts the inner part of the for loop into its own method. This way I can register my own IShoppingCartService that overrides only this submethod to check if the item being added is equivalent to a single shopping cart item. I can do my custom check and then defer to the base method.

Thoughts?

skoshelev:
Closed #3589

PR #3590, opened 2019-03-13
Extract shopping cart item equality in ShoppingCartService

justinjstark:
For issue #3589. Extract the single shopping cart item equality comparison so it can be easily overridden by plugins.


justinjstark:
I unintentionally changed the gift card comparison to not ignore case. This is fixed with the latest commit.

skoshelev:
Hi @justinjstark. Thank you for your help.

Issue #3604, opened 2019-03-18
Non-developer internal server error page not returning 500


samhippie:
nopCommerce version: 
4.1
Steps to reproduce the problem:
Add a code or data change that causes an internal server error (e.g. adding a divide by zero to a controller's action)
Run nop, making sure that the developer exception page is not enabled (see ApplicationBuilderExtensions's UseNopExceptionHandler)
Navigate to the page that causes that error
The response code will be something other than 500.

PR #3605, opened 2019-03-18
#3604 Fixed server error not returning 500 code

samhippie:
This fixes issue #3604 by using an action to set the the response code before returning the static page.


RomanovM:
Thanks for suggestion and contribution. 
We shouldn't use a CommonController for this, as it'll lead to an unwanted running of multiple action filters. Thus, we have created the new one for this purpose. See https://github.com/nopSolutions/nopCommerce/commit/80af103c2c8741fc8e79982a1acc36bda9ed601a

PR #3672, opened 2019-04-08
Modify DataSettingsManager class to check for enviroment connectionstring

rschreiber:
Hi Team,

Please consider this pull request which makes slight modifications to the _DataSettingsManager_ class in Nop.Core to check for an environment connection string called _DataConnectionString_ and make use of it instead of the connection string in _datasettings.json_

One of the benefits is to allow for a cleaner CI/CD pipeline in Azure, so that once the application is installed and functional, you can make use of the envirnment application settings for the connection string, which will allow for slot-swapping in Azure Web Apps to be used.

At present, slot-swapping can only be used if exactly the same database settings are in use, however, if you have a staging vs production slot, you can't use the existing Azure functionality.

Regrads,
Ryan


RomanovM:
Done in #4113. Thanks for this idea.

Issue #3633, opened 2019-03-26
Tier prices cannot be updated


AndreiMaz:
nopCommerce version: 4.20 BETA

Steps to reproduce the problem: 
1. Add a tier price to a product
2. Try to udpate it

You'll get the following exception:

SqlException: The UPDATE statement conflicted with the FOREIGN KEY constraint "FK_TierPrice_Product_ProductId". The conflict occurred in database "dev-nopCommerce", table "dbo.Product", column 'Id'.
The statement has been terminated.
System.Data.SqlClient.SqlConnection.OnError(SqlException exception, bool breakConnection, Action<Action> wrapCloseInAction)

DbUpdateException: An error occurred while updating the entries. See the inner exception for details.
Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)

Exception: Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> System.Data.SqlClient.SqlException: The UPDATE statement conflicted with the FOREIGN KEY constraint "FK_TierPrice_Product_ProductId". The conflict occurred in database "dev-nopCommerce", table "dbo.Product", column 'Id'.
The statement has been terminated.

dborovsky:
Hi @AndreiMaz  I`ll check this bug. Set me as Assignees.

dborovsky:
@AndreiMaz @RomanovM I tested. I see, we get this error when we select start date as current date. If we select future date. there is no error. 

I will fix this bug today. Select me as Assignee.

dborovsky:
[PR3679](https://github.com/nopSolutions/nopCommerce/pull/3679)

AndreiMaz:
@dborovsky thanks. it shouldn't be fixed this way because each tier price should have a product. It's about new changes in AutoMapper

dborovsky:
@AndreiMaz  hm... ok. i`ll check new changes in AutoMapper

dborovsky:
@AndreiMaz  I changed a bit. Check please. if you meant something another, tell me.

skoshelev:
Closed #3633

PR #3679, opened 2019-04-11
#3633 Tier prices cannot be updated

dborovsky:


AndreiMaz:
already fixed. thanks

Issue #75, opened 2016-01-14
Move away from KendoUI grid


mariannk:
Anyway it's not included in the latest versions of KendoUI (open-source edition). It becomes outdated

Main suggestion: https://datatables.net/ and https://almsaeedstudio.com/themes/AdminLTE/pages/tables/data.html

Also see https://material-ui.com/ and https://devexpress.github.io/devextreme-reactive/react/grid/

Suggestion 3 https://github.com/tonytomov/jqGrid and http://www.trirand.com/blog/

Suggestion 4:  http://ui-grid.info/

see http://www.nopcommerce.com/boards/t/29633/using-kendo-ui-legal.aspx and  http://www.nopcommerce.com/boards/t/29633/using-kendo-ui-legal.aspx?p=7#148339 

Once moved we can start using font-awesome icons to edit/delete buttons (maybe)

f1ana:
Performance on ui-grid appears to be poor with semi-large datasets.

keremdemirer:
[D3](https://github.com/d3/d3) is interesting. 
https://d3js.org/
http://bl.ocks.org/ahmohamed/4c4980f31abdef64e7e2

ronenelahav:
if help needed in  https://datatables.net/
I would love to help integrate it into nop.

tigran-avdalyan:
https://datatables.net/ and 
https://almsaeedstudio.com/themes/AdminLTE/pages/tables/data.html
+1

AndreiMaz:
once implemented see https://github.com/nopSolutions/nopCommerce/issues/2621

emumanu:
If we're moving away from Kendo UI I think [openui5](https://openui5.hana.ondemand.com), the opensource UI javascript library from SAP, offers great controls, and some very powerful ones like the Grid Table, Tree Table, planning calendar, multi input.... and also provides layout support, data binding, etc

Durkee:
I recommend to standardize on:
1. https://material-ui.com/
2. https://devexpress.github.io/devextreme-reactive/react/grid/

abilican:
[SlickGrid ](https://github.com/mleibman/SlickGrid) is awesome but not updated for a long time

nguyenhthai:
Will the new Datatable library support inline editing like before? I have checked out the library itself and seems like the feature is only available through paid version of the editor or build your own by using other libraries.

AndreiMaz:
@nguyenhthai that's right. we'll replace inline editing

[...]

PR #3699, opened 2019-04-17
Fix paging issues

spsinghats:
The take and skip post parameters were not being honored. And hence paging was not working.


RomanovM:
@spsinghats Please clarify what you mean? At now pagination works as expected.

exileDev:
See issue #75

PR #3756, opened 2019-05-01
Issue 2922 Incorrect routes matched for unrecognised single segment route

davidandrewpowell:


davidandrewpowell:
@RomanovM  I noticed that you had added the 'on hold / maybe wont' tag to this issue.

I was a little concerned that this PR may not get merged, possibly due to the large number of places that need to call the '_GenericUrl_' route so I just wanted to run this by you.

In a nutshell the '_GenericUrl_' route AND the '_Product_', '_Category_', '_Manufacturer_', '_Vendor_', '_NewsItem_', '_BlogPost_', '_Topic_' and '_ProductsByTag_' routes all have the same route format of "`{sename}`".  The only difference is that '_GenericUrl_' is a `GenericPathRoute` and the others are `LocalizedRoute`'s.

For urls such as [www.site.com/doesnotexist](), it tries to match the '_doesnotexist_' part against the `GenericUrl` route (looking at the slugs in the database).  When these don't match it then tries to match this part against the next route which is the '_Product_' route.  This matches the '`{sename}`' segment and thus calls the `ProductDetails` action.

The '_Product_', '_Category_', '_Manufacturer_', '_Vendor_', '_NewsItem_', '_BlogPost_', '_Topic_' and '_ProductsByTag_' routes are all duplicates of '_GenericUrl_' (by pattern) and are not required to match **incoming** urls or create **outgoing** urls as the '_GenericUrl_' does this fine.

I have painstakingly gone through the code to find all references to these old routes as much as possible.  As (I'm sure you are aware) there may be a very small chance that I might have missed one, so I would not suggest merging this PR into the 4.2 release at this stage.  But if possible, I would suggest merging it in to the 4.3 develop branch after 4.2 has been released.  Hopefully then if there are any missed route names, these will reveal themselves.  I am confident that I have found all that I can.

Whilst none of this is ideal, I do believe that the routing should be corrected around the url slug matching functionality.

I will leave this with your best judgement.

Thanks

Issue #3800, opened 2019-05-06
Simple refactoring of NivoSliderPlugin class


davidandrewpowell:
nopCommerce version:  4.20 beta

After fixing a separate issue in `NivoSliderPlugin`, I would suggest the following simple refactorings for this class:

i) Separate larger code sections into smaller methods
ii) Separate the large `InsertPicture` line into multiple lines for simplicity
iii) Remove redundant source code for banner picture 3
iv) Removal redundant comments for //pictures and //settings
v) Call `_webHelper.GetStoreLocation(false)` only once rather than once for each picture.
vi) Correct spelling of PLugin to Plugin in comments.

davidandrewpowell:
See pull request [3801](https://github.com/nopSolutions/nopCommerce/pull/3801).

davidandrewpowell:
Following feedback on this, it has been decided not to perform these refactorings at present in order to keep the plugin install methods consistent across all plugins.

PR #3801, opened 2019-05-06
#3800 Refactor NivoSliderPlugin Install() and Uninstall() methods.

davidandrewpowell:


RomanovM:
@davidandrewpowell  Thanks for the suggestion. But there is no need to separate installation methods, we do it same for all our plugins, and it's better to leave such code forconsistency.

davidandrewpowell:
@RomanovM No problem.  In this instance, I guess it was a bit of a random refactoring anyway whilst I was fixing another issue in this class.  Whilst I do think separating out the methods would be better in terms of readability, I do also understand that (for the tiny amount of value added in this case) it would be better to keep the methods consistent across plugins.  I will close this PR and the associated issue. 
 Thanks for the feedback.

Issue #3813, opened 2019-05-09
No globalization support in aspnetcore-runtime-alpine


wombitchgr:
nopCommerce version: 4.20 beta

Steps to reproduce the problem: 

1. Deploy to docker in 2.2-aspnetcore-runtime-alpine 
2. Install site
3. Check products on start page and notice the dollar signs on product prices are all invariant

aspnetcore-runtime-alpine does not come with support for globalization thus fallbacking to the invariant culture. This is done to reduce dependencies and the size of the image. It leads to InvariantCulture beeing the one and only CultureInfo and causes lots of weird behaviour.

![nop_invariant](https://user-images.githubusercontent.com/1703885/57451510-c68c7a00-7261-11e9-83ba-200084c04557.jpg)

skoshelev:
Closed #3813

PR #3814, opened 2019-05-09
Issue 3813 - Support globalization in alpine runtime

wombitchgr:
To enable globalization support we need to add the icu-libs apk package to the runtime environment in the docker container. We also need the DOTNET_SYSTEM_GLOBALIZATION_INVARIANT variable set to false (default true in alpine runtime) to support globalization.


wombitcher:
fixes #3813

skoshelev:
Hi @wombitchgr. Thank you for your help.

Issue #3819, opened 2019-05-10
Make it easy for store admin user to check Redis and Blob SystemInfo


rudgr:
nopCommerce version: 
4.20-beta

Make it easy for store admin user to check Redis and Blob SystemInfo

related issue: 
#3797

skoshelev:
Closed #3819

PR #3820, opened 2019-05-10
Features/rudgr/9 add blob and redis to system info admin

rudgr:
Make it easy for store admin user to check Redis and Blob SystemInfo #3819 

image


skoshelev:
Hi @rudgr. Thank you for your help. I did some changes you can see them in [this](https://github.com/nopSolutions/nopCommerce/commit/b7856f08289df3aee05dc0b421728416e58ea094) commit

rudgr:
nice tx! :)

Issue #3847, opened 2019-05-16
Store InstalledPlugins.json in SQL DB vs Redis


toddca:
nopCommerce version: 4.2

Steps to reproduce the problem: 
In 4.2 a feature was introduced to will help those of us that run NopCommerce in a web farm of 2 or more web servers - the feature allows for storing the contents of InstalledPlugins.json in Redis instead on disk, that way when a plugin is modified on one server the other servers will pick up that change whenever they are restarted. 

This is a great idea however I don't really care for the implementation - we should be storing in the DB vs Redis. I cannot think of any reason why we cannot use the SQL DB just as we do Redis. In addition this would help reduce some of the issues with #3809 where for some reason the decision was made to use a number of hard coded DBIDs without making them configurable. This would at least reduce one such Redis DB. 

We do not require the perf of Redis here for the caching, the primary reason for this feature is to keep the web servers in sync as to what is going on which the DB does all the time.

AndreiMaz:
thanks. but mostly duplicates https://github.com/nopSolutions/nopCommerce/issues/4113

toddca:
@AndreiMaz How does this duplicate an issue which was opened 5 months after I created this one?

PR #3851, opened 2019-05-17
Fix for #3847

toddca:
This is a fix for the issue we have in #3847 - basically there is no reason why we cannot store plugin installation data in the DB at all times, it never needs to be stored on disk nor in Redis.

Please note, you may want to rework the creation of the new SQL table into your build process for SQL but I just added it into this proof-of-concept here to show that it works.


src/Libraries/Nop.Services/Plugins/DbPluginsInfo.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Data;
+using System.Data.SqlClient;
+using Newtonsoft.Json;
+using Nop.Core.Data;
+using Nop.Core.Infrastructure;
+
+namespace Nop.Services.Plugins
+{
+    public class DbPluginsInfo : PluginsInfo
```
skoshelev: we do not use direct calls to the database, except for the service commands(backup, restore, reindex, etc.)

src/Libraries/Nop.Services/Plugins/DbPluginsInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        public override void Save()
+        {
+            if (DataSettingsManager.DatabaseIsInstalled)
+            {
+                EnsurePluginsTable();
+
+                var text = JsonConvert.SerializeObject(this, Formatting.Indented);
+
+                using (var cmd = GetDbCommand($@"   TRUNCATE TABLE [dbo].[Plugins];
```
skoshelev: we refused to use the DBO schema, allowing users to work with another scheme

src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationPartManagerExtensions.cs
```diff
@@ -354,28 +354,9 @@ private static bool IsPluginDirectory(string directoryName)
...
-                : new PluginsInfo(_fileProvider);
-
-            if (PluginsInfo.LoadPluginInfo() || useRedisToStorePluginsInfo || !config.RedisEnabled) 
-                return;
-
-            var redisPluginsInfo = new RedisPluginsInfo(_fileProvider, new RedisConnectionWrapper(config));
-
-            if (!redisPluginsInfo.LoadPluginInfo()) 
-                return;
+            PluginsInfo = new DbPluginsInfo(_fileProvider);
```
skoshelev: using a new file storage method must be configurable

skoshelev:
Hi @toddca Thank you for your help, but unfortunately we cannot accept your changes, as they contradict our approach. See my comments on the code

src/Libraries/Nop.Services/Plugins/DbPluginsInfo.cs
toddca: Being that at the time this call is made we do not have a DBContext how would you suggest this is done?

src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationPartManagerExtensions.cs
toddca: Why would you not to always store the plug-in info in the DB? It makes the most since to be stored in the DB and I cannot see any benefit of storing outside of the DB over storing in the DB.

src/Libraries/Nop.Services/Plugins/DbPluginsInfo.cs
skoshelev: In the discussion to the task, I tried to explain to you why we do not store information about the plugins in the database, if we had a ready working idea, we would solve not only this task but also a number of related one, such as updating plugins, control version and so on. But we stell have no such solution.

toddca: @skoshelev just so I understand - since your team does not want to write directly to the DB outside of EF the choice was made to not store plugins in the DB? 

Basically, as of 4.2, there is no method of keeping a NopCommerce web server farm in sync when it comes to installed plugins other than running a Redis server. We cannot use Redis caching as there are huge performance issues, so we must run an entire redis server just so we can keep installed plugins in sync -- and we must do it this way because we don't want to access the DB outside of EF. 

No offence, but I really don't understand this logic. Maybe I am one of the only ones running NopCommerce on more than 1 web server?

joekrauss: I run three web servers in my environment, which some might say is overkill. I do however see spikes of 400+ active customers on my site at times and I don't experience downtime for every Windows update as I'm load balanced.

As for Redis I'd love to have it operating 100% so I could maximize caching between the 3 servers and better yet I don't have to bring down 2 of 3 servers to install/uninstall plugins. I'm on @toddca side regarding this issue.

Do you know if there is any immediate plans to resolve the Redis issues? If not maybe it shouldn't be acknowledged that NopCommerce supports web farms until the issues can be resolved. I certainly fell for the release notes that it was supported and purchased servers and equipment believing NopCommerce was ready for web farms.

skoshelev: As I already wrote, problems with redise are solved by increasing the waiting time in the connection string, we also have plans to switch to an asynchronous call system, which should increase the performance of useing the redise db. As for your question regarding the use of direct queries, now we are not ready to implement such an approach, especially in light of our tasks to support other databases. We have a desire to fully transfer the information about plugins to the database, but not in version 4.20

toddca:
I am not convinced increasing the timeout and switching to asynchronous is going to improve performance. When there are an average of 100+ round trips to Redis per page request that type of load won’t scale and allow for a well performing site. Increasing the timeout of calls to redis does not make the page return faster. When using Redis caching you can’t and should not just throw everything in cache - for example, localized strings should be stored in memory and never in Redis. Store stuff that is expensive to fetch so all clients benefit, but make sure when accessing this content is not too chatty, otherwise keep it in server memory.

-Todd
________________________________

Issue #3856, opened 2019-05-17
Feature: Unpublished products should include a robots header tag so they will not be indexed by search engines


qamanuel:
nopCommerce version: 4.0

Steps to reproduce the problem: 
The problem this is fixing is that a product gets indexed by a search engine but after the product is no longer available (unpublished) it is still showing up in search. We still want customers who have purchased the item to still be able to visit the product detail page even after it is no longer available. 

`<meta name="robots" content="noindex">`

This would make for a good core enhancement using a feature flag such as "catalogSettings.NoindexUnpublishedProductPage". If this is set to True then include the above meta tag on any product detail page where the product is unpublished.

![Untitled](https://user-images.githubusercontent.com/47311805/57951787-c4a16700-78b0-11e9-99eb-6cd362c2e277.png)

DmitriyKulagin:
After some discussion, it was decided not to implement.

PR #3857, opened 2019-05-17
#3856 Feature: Unpublished products should include a robots header ta…

qamanuel:
from 
https://github.com/nopSolutions/nopCommerce/issues/3856


RomanovM:
Thanks for the suggestion, but we don't plan to implement it now.

PR #3976, opened 2019-06-26
Redirect if logged in

JeremyCrookshank:
Redirect the user to the homepage if they're logged in and they visit the login page accidentally.


RomanovM:
Thanks for the suggestion.
But an already logged in user may want to bind external authentication to his account, for this he needs to go to the login page again, see details [here](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web/Views/Customer/Info.cshtml#L401), so there is no need to redirect him to the homepage in this case.

PR #3983, opened 2019-07-08
Update public.accordion.js

Zambroff:
On the OnePageCheckout page the location of the page does not move with the changing of the accordion sections.

This is not that noticeable on a desktop display but could become quiet annoying on mobile device's as you often end up at the bottom of the page when clicking continue to next section.

Stores that may have additional forms towards the footer of the page may loose sales due to customers not scrolling up to see if they filling in the correct form to complete the checkout process.


DmitriyKulagin:
@Zambroff Hi,

Thank you very much for contributing to the development of the product!

PR #4010, opened 2019-08-02
Added product link in the list.

zandiarash:
Added product link in the product list for navigating to the actual page.


DmitriyKulagin:
@zandiarash Thanks. But after some consideration we've decided not to implement this functionality.

PR #4015, opened 2019-08-06
Check if headers are sent before setting cookie

rodolphito:
This allows using `_workContext.CurrentCustomer` from signal-r hub callbacks, which set their own headers. Without this check, an error is thrown that complains that headers have already been sent.


rodolphito:
Is there a way to tell the bot to run the tests again? The check only failed because a connection timed out.

PR #4016, opened 2019-08-07
Order Note Status Change Localization

parshan2:
OrderNotes localization for order status changes.


DmitriyKulagin:
@parshan2 Thanks!

PR #4021, opened 2019-08-09
Low stock quantity

ArturSprinter:
Added low stock quantity field in model, which now will allow users to configure when client will see 'low stock' availability status.

Previous 'in stock' and 'out of stock' nop statuses are nice, but most store engines provides also 'low stock' state to inform clients they should consider buying product now (because product may disappear in near future from store).


src/Libraries/Nop.Core/Domain/Catalog/ProductStockStatus.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
DmitriyKulagin: Remove unnecessary Usings

src/Libraries/Nop.Core/Domain/Catalog/ProductStockStatus.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Nop.Core.Domain.Catalog
+{
+    public enum ProductStockStatus
```
DmitriyKulagin: Add summary

DmitriyKulagin:
@ArturSprinter This is a good idea, but we have to take care of how best to implement it. And since the work here is not complete, I created a new issue (#5089) for this. You can continue your development with a new pull request. Thanks.

ArturSprinter:
Thanks.
I will try to continue work on this feature, if I got some spare time.

Issue #4037, opened 2019-08-26
.UseKestrel is limiting and not aligned with aspnetcore22+


awe-source:
nopCommerce version: 4.2

Steps to reproduce the problem: 
This (I believe) forces the application to use Kestrel (or at least Break ANCM startup in some scenarios, such as inprocess operation) 

            var host = WebHost.CreateDefaultBuilder(args)
                **.UseKestrel(options => options.AddServerHeader = false)**
                .UseStartup<Startup>()

According to the doco - the "Kestrel vs IIS" setup is "intelligently" already done in  .CreateDefaultBuilder() and the "addserverheader = false should be done in ".ConfigureKestrel()"

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/index?view=aspnetcore-2.2&tabs=windows#in-process-hosting-model
https://docs.microsoft.com/en-us/aspnet/core/migration/21-to-22?view=aspnetcore-2.2&tabs=visual-studio

The current setup appears to be against the 2.2 migration advice
and limits the application to ONLY outofprocess operation.

There may be reasons that you have done this, but if not then this appears to me/my reading to be unnecessarily restrictive (eg: prohibiting performance gains from using IIS directly on windows).

Additionally I note that there is a note on the web.config element to delay "AspNetCoreModeuleV2" until moving to 3.0, and I don't know if the decision to roll this back has anything to do with this, but its worth noting here, in case there is some insider knowledge that has an impact here.

DmitriyKulagin:
Implemented in #3653

Issue #3652, opened 2019-04-02
Upgrade to .NET Core 3.1


AndreiMaz:
https://dotnet.microsoft.com/download/dotnet-core/3.0 and https://dotnet.microsoft.com/download/dotnet-core/3.1

Also see https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio

Once finished do not forget to check https://github.com/nopSolutions/nopCommerce/issues/3531 (revert if required)

emiliano84:
I think aps.net core 3.0 will be still in preview for NopCommerce 4.20 release date

mrnams:
@emiliano84
Whats NopCommerce 4.20 release date?

emiliano84:
@mrnams should be this month
https://github.com/nopSolutions/nopCommerce/milestones

a-patel:
@AndreiMaz 
Please release Nop 4.20 in the .NET Core 3.0

emiliano84:
> @AndreiMaz
> Please release Nop 4.20 in the .NET Core 3.0

if they're going to release Nop 4.20 this months, they could use only .Net core 3.0 PREVIEW ... and I really doubt would be a good idea

mrnams:
@AndreiMaz @a-patel @emiliano84 @RomanovM 
I feel .NET Core 3.0 itself not stable.
Is there anything which is highly impacting?
My philosophy is ZERO-TOLERANCE  for any Regression.
Dont add anything that is OK,but never regress any.

AndreiMaz:
We're just waiting to .NET Core 3.0.0 RTM
Once it's done we'll upgrade

mrnams:
we are past due by 4days, any change in release date @emiliano84

emiliano84:
> we are past due by 4days, any change in release date @emiliano84

@mrnams I don't work for nopcommerce, I just refer to the milestone link :)

about .net core 3.0, ms just released preview4, but still in preview

mrnams:
@AndreiMaz we are past due by 4days, any change in release date.

bulutkartal:
with Server-side Blazor? :)

ChrisJollyAU:
I think I read somewhere .Net Core 3 is supposed to be RTM around July or a bit after. Any release after that will probably be moved up to 3.0

a-patel:
@AndreiMaz 
.NET Core 3.0
[...]

PR #4046, opened 2019-08-27
Moved to IIS inprocess default. Fixes. https://github.com/nopSolution…

jspuij:
Moved to IIS inprocess default. Fixes: https://github.com/nopSolutions/nopCommerce/issues/4037
Fixed version specific dependency on ASP.NET core. See https://github.com/dotnet/sdk/issues/2253

Cached response times halve with IIS in process hosting instead of kestrel passthrough.
I've also removed the version specific reference to Microsoft.AspnetCore.App to allow framework-dependent deployment instead of self-contained. This is a prerequisite for in process hosting.
To allow this and not get errors of different assembly versions (2.0 instead if 2.2 mostly) every project that references Nop.Core also needs a package reference to Microsoft.AspNetCore.App. This is per https://github.com/dotnet/sdk/issues/2253.


Durkee:
@jspuij What a great check in!

DmitriyKulagin:
@jspuij Thanks a lot for your contributions. We've already completed this task (see #3652) and it's available in the "develop" branch. Your pull request was also very helpful for us.

jspuij:
You're welcome!

PR #4074, opened 2019-09-06
Do not ignore email in GDPR when UsernamesEnabled

rodolphito:
When UsernamesEnabled, the registration form still asks for email. This means that the email is collected info regardless of UsernamesEnabled. The GDPR requires exporting *all* information related to the user, so this must also be exported regardless of the UsernamesEnabled setting.


DmitriyKulagin:
Hi @rodolphito,  thanks a lot!

PR #4094, opened 2019-09-24
Add IMapper for Dependency Injection

LucStr:
Currently the only possible way to access the mapper is through extension methods like .ToEntity or .ToModel and through a static property: AutoMapperConfiguration.Mapper . In my opinion this should also support DI. This PR will implement this.


skoshelev:
Hi @LucStr. Thank you very much for the suggestion, but I see no reason to pass the static field of the static class through DI

PR #4107, opened 2019-10-01
Created IDataTableModelFactory

brakim:
Moved creation of Admin Order List DataTablesModel to Factory.
This will make the creation available for override.

This should be done for all DataTablesModel creations.


DmitriyKulagin:
@brakim Thanks. But after some consideration we've decided not to implement this functionality.

PR #4119, opened 2019-10-15
Hotfix: DoesTypeImplementOpenGeneric

twix20:
The method DoesTypeImplementOpenGeneric now will check all implemented interfaces of given type, not only the first occurrence.


skoshelev:
Hi @twix20. Many thanks for your help

PR #4127, opened 2019-10-23
Fixes issue if customer was previously deleted and re-created.

ajwaka:
This issue sprouted up for our Administrative users.  We create logins for all of our employees so access isn't shared.  One employee left due to medical reasons and her account was "deleted".  2 years later she came back so her account had to be re-created.  In the Customers table - there's now 2 records for her by email.

On login - it fails because when looking up the user by Email, it will always return the first ever account that was created - which is now deleted in our Use-Case.  By changing the ordering it will return all matching results that are ACTIVE first - then take the FirstOrDefault.  Thus allowing her to login and utilize the site.

Worth Noting: also in CustomerService.cs n GetAllCustomers call - line # 167 is set to only ever return users that are active - thus no way of re-activating old/deleted users.

While the administrative user is the most likely use-case - I could see this as a use-case for clients that only sell to certain users EG: no-freely available login.  If they have a customer that "leaves" them - they may "delete" the user to deactivate the account.  And if the customer comes back.  I'm just spit-balling other use-cases.  I'm not sure how deleting a user with orders affects all the rest of the sales history.  Maybe this is only a valid use-case in the Admin user instance because there are no orders for the  user.

### DIFF
--- a/src/Libraries/Nop.Services/Customers/CustomerService.cs
+++ b/src/Libraries/Nop.Services/Customers/CustomerService.cs
@@ -415,7 +415,7 @@ namespace Nop.Services.Customers
                 return null;
 
             var query = from c in _customerRepository.Table
-                        orderby c.Id
+                        orderby c.Deleted == true
                         where c.Email == email
                         select c;
             var customer = query.FirstOrDefault();
@@ -433,7 +433,7 @@ namespace Nop.Services.Customers
                 return null;
 
             var query = from c in _customerRepository.Table
-                        orderby c.Id
+                        orderby c.Deleted == true
                         where c.SystemName == systemName
                         select c;
             var customer = query.FirstOrDefault();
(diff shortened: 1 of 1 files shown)


DmitriyKulagin:
Hi @ajwaka,
We can not confirm the proposed changes you. The logic of the system does not imply that remote users can log in again under their old accounts. However, we recommend that you do not delete users, since the entire purchase history is connected with them and many store owners need it because of the financial statements.

ajwaka:
You are correct - they can't login and that's the issue.  Because user's are logging in by an email address.  If the email address exists, then is "Deleted" then the user must be re-created.  Thus there are 2 accounts with the same email address.   

The system provides no way to access/re-activate a "deleted" account.  It also doesn't allow you to find any previously deleted account.  So you must create a new account with the same email address.  But this fails on every login due to your logic.

Please test out the use-case and reconsider.

PR #4151, opened 2019-11-09
Include WebP compression for uploaded images

aurony:
WebP has become a de-facto standard for maximum compression of images. Nowadays every major browser supports WebP formats. I believe that will bring big benefits for faster loading and thus SEO also.

I have used [imageprocessor.org](https://imageprocessor.org/imageprocessor/plugins/webp/) and it worked out-of-box for the image compression! I think there are also some other reliable open-source plugins for WebP in .net Framework. 

So, is not it a good time to include WebP compression for uploaded images?


RomanovM:
See https://github.com/nopSolutions/nopCommerce/issues/2073

Issue #3988, opened 2019-07-16
Add to cart issue in IE11. public.common.js ln: 142, htmlcode.append(content); 


anilgajeraxit:
nopCommerce version: 4.20

Steps to reproduce the problem: 
When you click on "ADD TO CART" button on IE11. It will not append notification message in notification tag.

error in public.common.js file. line number 142.

http://prntscr.com/ofmczq

yb11206:
REPLACING 
htmlcode.append(content);
with 
htmlcode.appendChild(content)

Should fix this issue on IE 11

DmitriyKulagin:
Not reproduced for EDGE

PR #4163, opened 2019-11-20
Bar Notification Fix for IE 11

VenkateshKonatham546:
I have fixed issue with bar notifications on IE, there are few methods which are not supported by IE so changed them to the supported ones


DmitriyKulagin:
Thanks a lot for your contributions. We've already completed this task (see #3988) and it's available in the "develop" branch. Your pull request was also very helpful for us.

Issue #4165, opened 2019-11-20
Allow use of AppAuthentication with Azure SQL


justinjstark:
nopCommerce version: 
Latest

Steps to reproduce the problem: 
To prevent storing passwords in the SQL connection string, we'd like to use AppAuthentication which provides integrated token-based security using Azure AD.
https://docs.microsoft.com/en-us/azure/app-service/app-service-web-tutorial-connect-msi

I can implement this by doing the following:
* Add Microsoft.Azure.Services.AppAuthentication NuGet package to Nop.Data
* Add a flag to dataSettings.json: UseAzureAppAuthentication
* Run the following code to the NopObjectContext constructor if the flag is set

```c#
var conn = (System.Data.SqlClient.SqlConnection)Database.GetDbConnection();
 conn.AccessToken = (new AzureServiceTokenProvider()).GetAccessTokenAsync("https://database.windows.net/").GetAwaiter().GetResult();
```

Let me know your thoughts. If this is acceptable I will issue a pull request.

AndreiMaz:
After some consideration we've decided not to implement this functionality out of the box.

PR #4178, opened 2019-11-27
Allow use of integrated security with Azure SQL

justinjstark:
For issue #4165. This change adds a flag to datasettings.json called 'UseAzureIntegratedSecurity'. When enabled, when constructing a NopObjectContext, an access token is obtained using AppAuthentication.AzureServiceTokenProvider and assigned to the SqlConnection. This allows for connecting to SQL using Azure Active Directory with integrated security. It works locally in Visual Studio (once configured) and on Azure using a managed identity.

More info:
https://docs.microsoft.com/en-us/azure/app-service/app-service-web-tutorial-connect-msi

### DIFF
--- a/src/Libraries/Nop.Core/Data/DataSettings.cs
+++ b/src/Libraries/Nop.Core/Data/DataSettings.cs
@@ -32,6 +32,11 @@ namespace Nop.Core.Data
         public string DataConnectionString { get; set; }
 
         /// <summary>
+        /// Whether to request an Azure token to use when authenticating with Azure SQL. Allows for connecting to Azure SQL using integrated security with a managed identity. Using integrated security, a password does not need to be embedded in the connection string. See https://docs.microsoft.com/en-us/azure/app-service/app-service-web-tutorial-connect-msi.
+        /// </summary>
+        public bool UseAzureIntegratedSecurity { get; set; }
+
+        /// <summary>
         /// Gets or sets a raw settings
         /// </summary>
         public IDictionary<string, string> RawDataSettings { get; }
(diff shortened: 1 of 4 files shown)


src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/DbContextOptionsBuilderExtensions.cs
```diff
@@ -25,10 +27,17 @@ public static void UseSqlServerWithLazyLoading(this DbContextOptionsBuilder opti
 
             var dbContextOptionsBuilder = optionsBuilder.UseLazyLoadingProxies();
 
+            var sqlConnection = new SqlConnection(dataSettings.DataConnectionString);
```
justinjstark: This is a bad idea, isn't it. I think this will use the same connection everywhere.

justinjstark:
I'm throwing some notes down. Maybe we should wait for this issue to be resolved so we can pass this configuration into UseSqlServer:
https://github.com/aspnet/EntityFrameworkCore/issues/13261

justinjstark:
My initial code was bad because it created a single SqlConnection that would get used over and over. I've fixed it. Instead of passing a SqlConnection into UseSqlServer, we obtain the token from within the NopObjectContext constructor as suggested in some of the threads in https://github.com/aspnet/EntityFrameworkCore/issues/13261.

The installer is not able to use this new feature yet. I can maybe work on that in the future but the installer could use some bigger improvements for better Azure SQL support.

nguyenhthai:
@justinjstark Hi. I am trying to use this implementation to get my instance of Nop working with the Azure managed SQL. However, when deploying to Azure, the instance cannot connect to the DB despite I have set up correctly the managed identity and Active Directory.

Here is my connection string like in the documentation:
```
"Server=tcp:[servername].database.windows.net,1433;Database=[dbname];",
```

The error that I got from the log is that **System.Data.SqlClient.SqlException (0x80131904): Login failed for user ''.** The user somehow is empty. I have checked the AD setup and I can retrieve the access token from it.

Do you have any suggestion for this?

skoshelev:
After some consideration we've decided not to implement this functionality out of the box

PR #4199, opened 2019-12-04
Add LoginUsernamesEnabled to allow having usernames but requiring login with email address

rodolphito:
This is for security reasons—usernames are public and as such it is easier to target a specific account. If login is done with email instead, then there is no such way to target a user—the email is private.

Let me know if anything else needs doing to get this merged.


rodolphito:
Why did the build check fail? It builds successfully on my machine, how can I figure out what went wrong?

RomanovM:
Thanks. We'll consider these changes.

RomanovM:
Decided not to implement it, thanks for the suggestion.

Issue #4166, opened 2019-11-20
Allow storing DataProtectionKeys in an Azure blob storage container


justinjstark:
nopCommerce version: 
latest

Steps to reproduce the problem: 
Currently, data protection keys can only be stored on disk in App_Data or using Redis. .NET Core provides Microsoft.AspNetCore.DataProtection.AzureStorage which stores the key in Azure.
https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/implementation/key-storage-providers?view=aspnetcore-3.0&tabs=visual-studio#azure-storage

It can connect using a connection string or AppAuthentication. We should provide support for both. The data protection key should live in a private container separate from the media container.

I can work on implementing this if given the OK.

skoshelev:
Hi @justinjstark. Thank you for your contribution.

Closed #4166.

PR #4219, opened 2019-12-10
Allow DataProtectionKeys to be stored in Azure Blob Storage

justinjstark:
Resolves issue #4166 

This has been tested and we are running it in production.

This PR adds 3 appsettings.json settings:

**UseAzureBlobStorageToStoreDataProtectionKeys**
true/false: whether to store data protection keys in an Azure blob storage container

**AzureBlobStorageContainerNameForDataProtectionKeys**
The container name to store the data protection keys. The connection string used is AzureBlobStorageConnectionString but the data protection keys need to live in a private container separate from the media container.

**AzureKeyVaultIdForDataProtectionKeys**
Optional key vault ID. If non-empty, the key stored in the blob container is encrypted using this Azure Key Vault.


skoshelev:
Hi @justinjstark. Thank you for your help

PR #4223, opened 2019-12-13
Refactoring: optmized sorted by filter without a second list

Leftyx:
Removed a second list to return a sorted list of entities.

### DIFF
--- a/src/Libraries/Nop.Core/Infrastructure/Extensions/IEnumerableExtensions.cs
+++ b/src/Libraries/Nop.Core/Infrastructure/Extensions/IEnumerableExtensions.cs
@@ -1,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Nop.Core.Infrastructure.Extensions
+{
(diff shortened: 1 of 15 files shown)


skoshelev:
Hi @Leftyx. Thank you for suggests but it will not be relevant after solving this problem: https://github.com/nopSolutions/nopCommerce/issues/4897

Leftyx:
> 
> 
> Hi @Leftyx. Thank you for suggests but it will not be relevant after solving this problem: #4897

@skoshelev I don't think so. Some of the changes have nothing to do with the actual IN queries. Did you even bother to have a look at the changes ?

skoshelev:
Hi @Leftyx. I looked at your main proposal, which is related to sorting without using an additional array, carefully, we plan to completely rewrite the functionality of methods that use the old approach and the edits you proposed will most likely be incompatible with it. As for the rest of the edits, I apologize, I really did not analyze them, since I decided that they were related to the main task. Please make them separate PRs. We really appreciate your help, but it's very difficult to analyze edits when several different problems are solved in one PR.

PR #4229, opened 2019-12-19
Modified labels to use Id and included Aria-labels to comply with ADA

not2envy:
Change 1: added changes for quantity on product page for aria lable to comply with ADA
added aria-label for product quantity on both classes: qty-dropdown and qty-input   (_AddToCart.cshtml).

Added Locals for AriaLabel for Product.Qty.AriaLabel    (defaultResources.nopres.xml). 
				<LocaleResource Name="Products.Qty.AriaLabel">
				    <Value>Enter a quantity</Value>
				 </LocaleResource>

Change 2: Add property for label using Id and that Id to price span  (_ProductPrice.cshtml)

Change 3: Added for on label using Id from span that also need to include a Id on Search page and the same was done on Register page
			Added Label with Id for gender within (Register.cshtml):
			Added label with Id for price range within (Search.cshtml):

### DIFF
--- a/src/Presentation/Nop.Web/App_Data/Localization/defaultResources.nopres.xml
+++ b/src/Presentation/Nop.Web/App_Data/Localization/defaultResources.nopres.xml
@@ -17256,6 +17256,9 @@
   <LocaleResource Name="Products.Qty">
     <Value>Qty</Value>
   </LocaleResource>
+  <LocaleResource Name="Products.Qty.AriaLabel">
+    <Value>Enter a quantity</Value>
+  </LocaleResource>
   <LocaleResource Name="Products.RecentlyViewedProducts">
     <Value>Recently viewed products</Value>
   </LocaleResource>
--- a/src/Presentation/Nop.Web/Views/Product/_ProductPrice.cshtml
+++ b/src/Presentation/Nop.Web/Views/Product/_ProductPrice.cshtml
@@ -7,7 +7,7 @@
     <div class="prices">
         @if (Model.CallForPrice)
         {
-        @*call for price*@
+            @*call for price*@
             <div class="product-price call-for-price">
                 <span>@T("Products.CallForPrice")</span>
             </div>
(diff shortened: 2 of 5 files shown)


src/Presentation/Nop.Web/App_Data/Localization/defaultResources.nopres.xml
```diff
@@ -17256,6 +17256,9 @@
   <LocaleResource Name="Products.Qty">
     <Value>Qty</Value>
   </LocaleResource>
+  <LocaleResource Name="Products.Qty.AriaLabel">
+    <Value>Enter a quantity</Value>
+  </LocaleResource>
```
DmitriyKulagin: The value of the new locale must also be added to the update script (at the end of the list). This is done in order to be able to apply your changes without reinstalling the store.
_/upgradescripts/4.20-4.30 (under development)/upgrade.sql_

src/Presentation/Nop.Web/Views/Product/_ProductPrice.cshtml
```diff
@@ -40,19 +40,16 @@
...
                     @*display "Price:" label if we have old price or discounted one*@
-                    <label>@T("Products.Price"):</label>
+                    <label for="price-value-@(Model.ProductId)">@T("Products.Price"):</label>
                 }
                 @*render price*@
-                <span @if (string.IsNullOrWhiteSpace(Model.PriceWithDiscount))
-                        {
-                            <text>class="price-value-@(Model.ProductId)" </text>
-                        }>
+                <span @if (string.IsNullOrWhiteSpace(Model.PriceWithDiscount)) { <text> id="price-value-@(Model.ProductId)" class="price-value-@(Model.ProductId)" </text>  }>
```
DmitriyKulagin: Remove the double space at the end of the line

DmitriyKulagin:
@not2envy Thanks for your contribution.

PR #4232, opened 2019-12-19
Avalara: Use product name instead of description

justinjstark:
Use the product name for Avalara item code, not the short description.

Using short description, it looks like this in Avalara (ugly):
image


pastime28:
I fully support this. We are truncating the Avalara logs every 2 days because the size of this field blows the db up when using description. In turn, this makes it difficult to troubleshoot any integration issue without the logs or going to a db backup.

PR #4244, opened 2020-01-02
feat (backend): add webapi plugin

jorgepsmatos:
Added the following plugin https://github.com/SevenSpikes/api-plugin-for-nopcommerce as one of the default ones.

Only supports version 4.20 of the nopCommerce solution. The development branch is already on 4.30 and that's why I am targeting the master branch.

I suggest that we delete all branches except master and revert it to the tag `release-4.2.0` and we start using **GITHUB** flow instead of Git flow.


jorgepsmatos:
Targeted the wrong repo. Can be deleted.

Issue #4261, opened 2020-01-17
Hide "registered in the store" label when we have only one store


AndreiMaz:
There's no need to display this label in most cases (when we have only one store). Let's hide it in this case

DmitriyKulagin:
Closed #4261

PR #4263, opened 2020-01-19
Hide 'Registered in the Store' label if there is only one store

alessandromontividiu03:
https://github.com/nopSolutions/nopCommerce/issues/4261


DmitriyKulagin:
@alessandromontividiu03 
Thanks

Issue #4295, opened 2020-02-17
Customer attributes allow html, url, sql, etc.


jk-itidea:
nopCommerce version: all

Steps to reproduce the problem: 
When registering (or later when editing on "My Account"), html, url, sql, etc. can be entered into the "First name", "Last name" (and possibly other) fields.

This can be dangerous for a number of reasons, but one already observed in the wild is SPAM bots hijacking registration message to send unwanted content, by using victim email address to create an account and entering SPAM content into "First name" and "Last name". This is a serious issue if shop allows to create accounts without confirmation (and possibly even with confirmation) depending on the email template used.

In older versions those fields are decorated with AllowHtml attribute, I'm not sure about versions from 4 up, but the problem is still present.

manswami23:
I'm interested in working on this

AndreiMaz:
It's by design. And there's no any issue because all entered information is encoded later

PR #4300, opened 2020-02-20
Customer attributes allow html, url, sql, etc. #4295

manswami23:
Customer attributes allow html, url, sql, etc. Issue #4295
added validation in customervalidator to ensure that inputs are safe.


RomanovM:
We don't plan to merge these changes, see [this comment](https://github.com/nopSolutions/nopCommerce/issues/4295#issuecomment-593300990) in the related issue.
Anyway thanks for your suggestion and contribution.

PR #4315, opened 2020-02-28
fixes in mysql/StoredProcedures.sql to make it work correctly on mysql 8.0

evilbloodydemon:
- fixed db objects casing
- added replication-related declarations to sprocs definitions


exileDev:
Hi @evilbloodydemon! Thank you for your help.

PR #4329, opened 2020-03-09
bug wen new intall

angel6708:
before installing, Logger and PluginService are both unavailable.
when intall CommandTimeout shout be set to a lang span of time.


angel6708:
before installing, Logger and PluginService are both unavailable.
when install ,CommandTimeout should be set to a long span of time.

PR #4351, opened 2020-03-21
Use NopCommonDefaults.DbBackupsPath for static file provider

rhodosaur:
Simple fix. Allows download in the administration if backup directory is changed.


DmitriyKulagin:
@rhodosaur Thanks a lot!

PR #4370, opened 2020-03-26
Fixed problem with import language file for second language

sidatacom:


sidatacom:
Why did you not accept the pull but fix it by yourself ?
Is it really an opensource project ?

PR #4374, opened 2020-03-27
Improved IIS setting to use InProcess mode

vitorelli:
InProcess mode avoid issues with request timeout when using with IIS.

https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/aspnet-core-module?view=aspnetcore-3.1#attributes-of-the-aspnetcore-element


DmitriyKulagin:
@vitorelli The hosting model is an optional attribute, and its default value is "inprocess". Therefore, it makes no sense to add it. Thank you for your help!

vitorelli:
@DmitriyKulagin thanks for reply, this behavior was changed in .net core 3.1.

PR #4375, opened 2020-03-27
Fixed discount issues

vitorelli:
Fixed issue cause by adding discount type filtered result to cache, the filter is been used in memory after execute the query in database


skoshelev:
Thank you @vitorelli. See full changes in the https://github.com/nopSolutions/nopCommerce/commit/acb43f2fb681bd10bc2198b9c69be4dc074551a1

PR #4411, opened 2020-03-31
Fix ANCM In-Process Handler Load Failure

joelmo:
HTTP Error 500.0 - ANCM In-Process Handler Load Failure. https://github.com/dotnet/aspnetcore/issues/6111.


joelmo:
Fixes #4412

DmitriyKulagin:
Hi @joelmo
In fact, there are no arguments to change the AspNetCoreModuleV2 module to an older version. Perhaps if we find some serious problems associated with this, then we will return to this topic. Thanks!

PR #4442, opened 2020-04-02
Add New method

Sandeep911:
New method to fetch list of all the attributes with specified keyGroup and keyName.
For example , fetching list of all the phone attributes for customer.


skoshelev:
Hi? @Sandeep911. Thank you very much for your suggestion, but, unfortunately, at the moment I do not see a real application of this method in our code

PR #4454, opened 2020-04-02
Revise exception handler to not rethrow exception

Shane32:
I'm on Nop 4.1 where stack traces are hidden.  This is the code I used to fix the issue.  It seems that the problem was already somewhat fixed by utilizing `ExceptionDispatchInfo.Throw` in a prior commit.  This method does not modify the call stack at all.

Since it would seem that you've already fixed the problem, there's no need to pull this PR unless you think this technique is better.  Thanks!!


skoshelev:
Hi, @Shane32. Thank you very much for your suggestion, but we will leave our implementation as it seems to us more preferable.

Issue #4423, opened 2020-03-31
Multiple warehouse quantity box should allow only number


rajupaladiya:
nopCommerce version: 4.30 Beta

Steps to reproduce the problem: 
> Go to product edit page
> Inventory Section > Tick Multiple warehouses	
This allow admin to set warehouse wise quantity.

The quantity textbox should allow only numeric values instead of characters

![19-multiple_warehouse](https://user-images.githubusercontent.com/13154058/78064733-cdf3cf00-73af-11ea-913d-6cfd752e56f6.png)

rajupaladiya:
@DmitriyKulagin - What about bug label?

PR #4472, opened 2020-04-04
Fix of #4423 - textbox accepts only numeric instead of characters

jeeckeman:


DmitriyKulagin:
#4423

DmitriyKulagin:
@jeeckeman Thanks a lot!

PR #4474, opened 2020-04-04
Update Startup.cs

cihanturgut:


DmitriyKulagin:
Hi, @cihanturgut  
We already use these methods in the project - [ServiceCollectionExtensions.cs](https://github.com/nopSolutions/nopCommerce/blob/57607ba242fd66cc92fa2b48d34e21567db408db/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ServiceCollectionExtensions.cs#L274)

Issue #4467, opened 2020-04-04
Grid header width issue on product edit page when change product type simple to group


rajupaladiya:
nopCommerce version: 4.30 Beta

Steps to reproduce the problem: 
- Go to any simple product edit page
- Change product type simple to Grouped

See now  header of associated product grid
![002_associated](https://user-images.githubusercontent.com/13154058/78423316-27475100-7683-11ea-8bc5-a579611df426.gif)

PR #4475, opened 2020-04-05
Fix of column header sizing

jeeckeman:
Added the columns.adjust() function call.


DmitriyKulagin:
Associated with #4467

AndreiMaz:
thanks. fixed - https://github.com/nopSolutions/nopCommerce/issues/4467

Issue #4477, opened 2020-04-05
Fix some Simplified Chinese localization texts.


byteflying:
nopCommerce version: 
4.30 (develop branch)

Steps to reproduce the problem: 
1.I need to fix some Simplified Chinese localization texts.
2.I will carefully compared each line to make sure there was no mistake.
3.Thx 4 ur reading.

JadynWong:
All installation translation `installation. *. xml` should be changed.

The historical version of nop only supports 4.2. Version 4.3 added MySQL support, but the name of the translation key was tightly modified.
https://github.com/nopSolutions/nopCommerce/commit/0968d6cac4351b4f9a6f41b7c108996f199456d5#diff-3ba7e2ba30aad08aa89ad7b2b27bae72

byteflying:
already done

AndreiMaz:
Merged 
@byteflying @JadynWong Thanks a lot!

PR #4478, opened 2020-04-05
#4477 Fix some Simplified Chinese texts.

byteflying:
1.I  fixed some Simplified Chinese localization texts.
2.I carefully compared each line to make sure there was no mistake.

### DIFF
--- a/src/Presentation/Nop.Web/App_Data/Localization/Installation/installation.zhs.xml
+++ b/src/Presentation/Nop.Web/App_Data/Localization/Installation/installation.zhs.xml
@@ -97,7 +97,7 @@
     <Value>[推荐]</Value>
   </LocaleResource>
   <LocaleResource Name="RestartInstallation">
-    <Value>重新安装</Value>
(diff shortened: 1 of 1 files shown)


src/Presentation/Nop.Web/App_Data/Localization/Installation/installation.zhs.xml
```diff
@@ -97,40 +97,40 @@
     <Value>[推荐]</Value>
   </LocaleResource>
   <LocaleResource Name="RestartInstallation">
-    <Value>重启安装</Value>
+    <Value>重新安装</Value>
```
JadynWong, line 100: I think the original translation is correct.It's really restart the application.

AndreiMaz:
@byteflying Could you please update the pull request according to the comment of @byteflying above?

byteflying:
> @byteflying Could you please update the pull request according to the comment of @byteflying above?

Of course,I will do it this night.

PR #4546, opened 2020-04-12
Fix wrong decimal precision under certain culture in PayPal standard, smart button plugin

Meowzz95:
The current implementation uses toString("F") to convert money to string and send it to PayPal API. However, toString("F") gives 3 decimal places under some(if not all) culture settings resulting in rejection from PayPal API(only 2 decimal places is accepted).

So I think it is a good idea to enforce this 2 d.p. requirement by formatting it explicitly.


Meowzz95:
This change was first PRed in the plugin repo, but I realized the plugin has been moved to this main repo, so I'm making the PR here.

Original PR: https://github.com/nopSolutions/PayPal-SmartButtons-plugin-for-nopcommerce/pull/7

AndreiMaz:
@Meowzz95 Thanks. But there's no need to edit the PayPal Standard plugin. It's correct. And I referenced it just as an example of how it should be done

The only plugin that should be updated is the PayPal Smart Payment Buttons

Meowzz95:
Oh now I understand, will push a update soon

Meowzz95:
@AndreiMaz Let me close this and reopen another PR, sorry for the trouble.

Issue #4597, opened 2020-04-19
Video list issue with RoxyFileman in rich editor


rajupaladiya:
nopCommerce version: 4.30 Beta

Steps to reproduce the problem: 
Go to Admin > Category Edit page > Full description Rich editor (Any rich editor)

Go to insert > Video then upload any video.
Here are two problems
1. If video is large (more than 20 or 25 MB) then it not uploading, status shows success but in file folder video not save
2. Small size uploaded and saved in upload folder successfully but it not showing in RoxyFileMane browse window

Here is video
https://www.loom.com/share/2bc0087350574b32afb2a8401dd2a181

AndreiMaz:
@rajupaladiya 

1. Maximum file size is the default limitation of asp.net core. You can edit this setting
2. It's an image editor, not video. So it's OK and is limited by the "type" parameter at https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web/Areas/Admin/Controllers/RoxyFilemanController.cs#L82. But we'll think about limiting a list of type that could be uploaded - https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web/wwwroot/lib/Roxy_Fileman/conf.json#L12

PRudykIteo:
@AndreiMaz I came across same issue few days ago.
I can understand that You may not want videos in f.e. product description.
However, IMHO, behavior of RF is confusing - video file uploads correctly and RF shows brief success message but file disappears in a blink of an eye from Roxy window.

I was able to fix this and add a video to product description (mainly by tweaking "type" You've mentioned). I can send You a PR if You want.

rajupaladiya:
@PRudykIteo - if you do that then it would be good. I don't think @AndreiMaz sshow complete video and closed issue without check. 

@AndreiMaz - I was telling about video and I uploaded video from video menu, not from pictures.

![9](https://user-images.githubusercontent.com/13154058/79840031-f8471400-83d2-11ea-899e-4710c2ebb340.png)
![10](https://user-images.githubusercontent.com/13154058/79840039-fb420480-83d2-11ea-9d8a-d08f20cc2133.png)

PRudykIteo:
@AndreiMaz do You want me to share a possible solution in pull request?

AndreiMaz:
@PRudykIteo Sure, you're more than welcome

holydk:
Closed #4597

PR #4610, opened 2020-04-24
Issue 4597 handling videos

PRudykIteo:
This is a possible solution for issue #4597 .
This change allows Roxy Fileman to display many types of media files.
Please take a look at this list of extensions - it was based on list of extension icons in Roxy js files and may be incomplete.

I've included video and audio files, because Roxy treats them all as Media type. You may want to remove audio extensions, because tinyMCE show these files in "Add Video" tab.


holydk:
@PRudykIteo Hi. Thanks a lot for your help!

Issue #4133, opened 2019-10-29
Not overridden checkboxes display wrong data when the plugin configuration is saved for any store.


holydk:
nopCommerce version: 4.30

Steps to reproduce the problem:
- Go to plugin configuration edit page (example Qualpay)
- Select editing for “All store”
- Check any checkbox which can be overridden for other store
- Select editing for “your store”
- Try to save configuration

Then, the post method in controller gets wrong data and “view” display wrong data after.
 
### it may be related to [this](https://stackoverflow.com/questions/2697299/asp-net-mvc-why-is-html-checkbox-generating-an-additional-hidden-input)

![image](https://user-images.githubusercontent.com/37818094/67776358-41945780-fa71-11e9-85e1-8c721d5e51f9.png)
![image](https://user-images.githubusercontent.com/37818094/67776419-5670eb00-fa71-11e9-98b5-24eedf8ae90b.png)

holydk:
Closed #4133

holydk:
Validations are missed with `RedirectToAction`

AndreiMaz:
New info - https://www.nopcommerce.com/en/boards/topic/81557/nopcommerce-430-rc-is-available-please-share-your-impressions/page/13#264717

holydk:
Closed #4133

micheleinhorn:
sorry still does not work even though i made the changes in the commit
An unhandled exception occurred while processing the request.
AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)

System.Threading.Tasks.Task<TResult>.GetResultCore(bool waitCompletionNotification) in Future.cs, line 491
NullReferenceException: Object reference not set to an instance of an object.

Nop.Web.Framework.TagHelpers.Admin.NopOverrideStoreCheckboxHelper.Process(TagHelperContext context, TagHelperOutput output) in NopOverrideStoreCheckboxHelper.cs, line 126

    Stack Query Cookies Headers Routing 

    AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
        System.Threading.Tasks.Task<TResult>.GetResultCore(bool waitCompletionNotification) in Future.cs
        Nop.Web.Framework.TagHelpers.Admin.NopPanelTagHelper.Process(TagHelperContext context, TagHelperOutput output) in NopPanelTagHelper.cs

                panelContainer.InnerHtml.AppendHtml(output.GetChildContentAsync().Result.GetContent());

[...]

PR #4812, opened 2020-06-01
#4133 Fix wrong values in disabled checkboxes for multistore settings

MarinaAndreeva:
The issue described in #4133 is still reproducible. This is a possible solution for this issue.

The state of the **disable** attribute should always be the same for both **checkbox input** and its hidden counterpart (which always has a **false** value).


holydk:
@MarinaAndreeva Thanks! I merged you changes across '4.30-bug-fixes' branch.

Issue #3885, opened 2019-05-24
Potential refactoring of ExportManager.GetPictures(Product) method


davidandrewpowell:
nopCommerce version: 4.20 beta

Please correct me if I am wrong but I believe that the following `ExportManager.GetPictures(Product)` method:

```
protected virtual string[] GetPictures(Product product)
{
    //pictures (up to 3 pictures)
    string picture1 = null;
    string picture2 = null;
    string picture3 = null;
    var pictures = _pictureService.GetPicturesByProductId(product.Id, 3);
    for (var i = 0; i < pictures.Count; i++)
    {
        var pictureLocalPath = _pictureService.GetThumbLocalPath(pictures[i]);
        switch (i)
        {
            case 0:
                picture1 = pictureLocalPath;
                break;
            case 1:
                picture2 = pictureLocalPath;
                break;
            case 2:
                picture3 = pictureLocalPath;
                break;
        }
    }

    return new[] { picture1, picture2, picture3 };
}

```
can be simplified as:

```
protected virtual string[] GetPictures(Product product)
{
    var pictures = _pictureService.GetPicturesByProductId(product.Id, 3);
    return pictures.Select(picture => _pictureService.GetThumbLocalPath(picture)).ToArray();
}
```

Going further, it seems this method is only called when creating a `PropertyByName` in the `ExportManager.ExportProductsToXlsx(IEnumerable<Product>)` method:

```
...
new PropertyByName<Product>("Picture1", p => GetPictures(p)[0]),
new PropertyByName<Product>("Picture2", p => GetPictures(p)[1]),
new PropertyByName<Product>("Picture3", p => GetPictures(p)[2])
...
```

This means that 3 pictures are returned from the method each time it is called when only one is needed.  Instead, we can just return a single picture to give:

```
protected virtual string GetPicture(Product product, int pictureIndex)
{
    var pictures = _pictureService.GetPicturesByProductId(product.Id, 3);
    return _pictureService.GetThumbLocalPath(pictures[pictureIndex]);
}

...
new PropertyByName<Product>("Picture1", p => GetPicture(p, 0),
new PropertyByName<Product>("Picture2", p => GetPicture(p, 1),
new PropertyByName<Product>("Picture3", p => GetPicture(p, 2)
...
```

Note the renaming of the method above from `GetPictures` to `GetPicture`.  Note also that we have reduced the number of times that we call `_pictureService.GetThumbLocalPath(Picture)`.

[...]

PR #4815, opened 2020-06-02
Issue #3885 Refactoring of ExportManager.GetPictures(Product) method

roman-gherman:


roman-gherman:
@RomanovM @skoshelev anything about this?

src/Libraries/Nop.Services/ExportImport/ExportManager.cs
```diff
@@ -344,38 +344,20 @@ protected virtual string GetProductTags(Product product)
...
-
-            return new[] { picture1, picture2, picture3 };
+            // we need only the picture at a specific index, no need to get more pictures than that
+            var recordsToReturn = pictureIndex + 1;
+            var pictures = _pictureService.GetPicturesByProductId(product.Id, recordsToReturn);
+            return pictures.Count > pictureIndex ? _pictureService.GetThumbLocalPath(pictures[pictureIndex]) : null;
         }
 
-        protected virtual bool IgnoreExportPoductProperty(Func<ProductEditorSettings, bool> func)
+        protected virtual bool IgnoreExportProductProperty(Func<ProductEditorSettings, bool> func)
```
skoshelev, line 360: Hi @roman-gherman. Thank you for the work done, but in the future, please do not mix two different refactorings, it is more difficult to analyze the code. In this case, a typo was not worth correcting.

roman-gherman: @skoshelev yes, makes sense

Issue #4782, opened 2020-05-22
Ensure no exception is thrown when we have duplicated locale names


AndreiMaz:
If you have a language resource with the same key (there is no constraint in the database to ensure unique keys) then nopCommerce 4.30 will throw an exception because of this code assuming the resource name is unique:

```
var lsNamesList = _lsrRepository.Table
                .Where(lsr => lsr.LanguageId == language.Id)
                .ToDictionary(lsr => lsr.ResourceName, lsr => lsr);

```
Source: https://www.nopcommerce.com/en/boards/topic/80047/nopcommerce-430-bug-fixes-and-improvements/page/6#264900

roman-gherman:
fixed in https://github.com/nopSolutions/nopCommerce/pull/4816

skoshelev:
Closed #4782

sorenaxla:
We had to revert this commit in our solutions as the orderBy part didn't complete. As an result the solution did not proceed from the method and the solution could not start up when run localhost. We have a single language (updated from the Danish 4.20 version) in the solution that we first encountered this issue.

roman-gherman:
@sorenaxla what do you mean with: "orderBy part didn't complete"? 
did it hang there? where there any errors?

sorenaxla:
Yes, it hangs. I have tried in two different 4.30 solutions with two different languages in each and the issue occurs only with the new commit:

                  .GroupBy(lsr => lsr.ResourceName, (key, groupedLsr) => 
                      groupedLsr.OrderBy(lsr => lsr.Id).Last())

Note, that both sites are using the NopThemes Prisma theme (latest 4.30 version) which adds some strings, which might/could be the issues... just a note.

I have attached the xml for the english language extracted from one of the solutions.
[language_pack (1).txt](https://github.com/nopSolutions/nopCommerce/files/4735815/language_pack.1.txt)

joekrauss:
We are currently testing 4.3 with Nop-Templates and I see a 2 to 4 minute delay starting the website after applying the code changes to correct the duplication exception error. 

I verified the NopCommerce settings to load locals, SEO, etc. are not checked during startup. Removing Nop-Templates plug-ins & themes allows the site to quickly startup within 30 seconds. I have my own custom plugins enabled, so either the change impacted performance on NopCommerce startup or something is wrong with Nop-Templates plugins / theme.

roman-gherman:
[...]

PR #4816, opened 2020-06-02
Issue 4782 Ensure no exception is thrown when we have duplicated loca…

roman-gherman:
Fix for https://github.com/nopSolutions/nopCommerce/issues/4782
And small performance improvement

### DIFF
--- a/src/Libraries/Nop.Services/Localization/LocalizationService.cs
+++ b/src/Libraries/Nop.Services/Localization/LocalizationService.cs
@@ -435,6 +435,8 @@ namespace Nop.Services.Localization
 
             var lsNamesList = _lsrRepository.Table
                 .Where(lsr => lsr.LanguageId == language.Id)
+                .GroupBy(lsr => lsr.ResourceName, (key, groupedLsr) => 
+                    groupedLsr.OrderBy(lsr => lsr.Id).Last())
                 .ToDictionary(lsr => lsr.ResourceName, lsr => lsr);
 
             var lrsToUpdateList = new List<LocaleStringResource>();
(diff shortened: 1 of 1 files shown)


src/Libraries/Nop.Services/Localization/LocalizationService.cs
```diff
@@ -454,16 +456,11 @@ public virtual void ImportResourcesFromXml(Language language, StreamReader xmlSt
...
-                        lrsToInsertList.Add(name, lsr);
+                    lrsToInsertList[name] = lsr;
                 }
             }
 
-            foreach (var lrsToUpdate in lrsToUpdateList) 
-                _lsrRepository.Update(lrsToUpdate);
-
+            _lsrRepository.Update(lrsToUpdateList);
             _lsrRepository.Insert(lrsToInsertList.Values);
```
skoshelev, line 464: Hi @roman-gherman. Thank you for your contribution, I looked at your solution, everything is fine, but lrsToInsertList can now become really a List and not a Dictionary

roman-gherman: Hi @skoshelev, 
This is an Add or Update: lrsToInsertList[name] = lsr; 
This was done to avoid doing lrsToInsertList.ContainsKey(name).
 Complexity for the dictionary is O(1), which makes it faster.

If I convert lrsToInsertList to a List, then I will have to do a check for:
```
var item = lrsToInsertList.FirstOrDefault(lsr => lsr.ResourceName == name)
if(item  != null)
item = new LocaleStringResource....;
else
lrsToInsertList.Add(...)

```
Which makes it more complex and slow

skoshelev: Yes, you may be right, I was mistaken with the opportunity to return to the List. Thanks again for your contribution.

Issue #4842, opened 2020-06-12
Copy Product. Unpublished categories are not copied


Ochapparel:
nopCommerce version: 
4.30

Steps to reproduce the problem: 
1) Set a category to NOT PUBLISHED
2) Copy a product that has that category assigned
3) The copied product does NOT get the category copied over
4) I tested this on demo site too

See PR  - https://github.com/nopSolutions/nopCommerce/pull/4855

**Non Published Category**
![image](https://user-images.githubusercontent.com/20989983/84511059-f93e4880-ac93-11ea-9868-02381ed2d334.png)

**Item to be copied WITH Non-Published Category**
![image](https://user-images.githubusercontent.com/20989983/84511005-e9266900-ac93-11ea-925e-b321595a3385.png)

**Copy Options**
![image](https://user-images.githubusercontent.com/20989983/84511009-eb88c300-ac93-11ea-8e6e-7de17fe33f2b.png)

**Copied item with NO CATEGORY copied over**
![image](https://user-images.githubusercontent.com/20989983/84511020-edeb1d00-ac93-11ea-8bbf-5e8934f98082.png)

PR #4855, opened 2020-06-16
#4842 Copy product is ignoring unpublished categoires

denisburlacu:


exileDev:
@denisburlacu Thanks a lot!

Issue #4850, opened 2020-06-16
Inheritance not able to be used in plugins given BuildTable implementation


matthewha:
nopCommerce version: 4.30

Steps to reproduce the problem: 
MigrationManager.cs BuildTable only takes the Type passed to it and does not incorporate any of the parent classes/attributes.
As a result, when using inheritance in a plugin, the BuildTable function does not include any of the columns that should be there due to inheritance from parent class/es. Unless there is a workaround that I'm not aware of, this means that each database table object will need to have lots of replicated code by not using a decent inheritance software architecture within a plugin.
What would be great is if the MigrationManager.cs incorporated a recursive search up the inheritance hierarchy until reaching BaseEntity, similar to that shown in https://docs.microsoft.com/en-us/dotnet/api/system.type.basetype?view=netcore-3.1

PR - https://github.com/nopSolutions/nopCommerce/pull/4857

denisburlacu:
#4857

matthewha:
Thanks Denis for doing this and nopcommerce team for including in next version

exileDev:
After some consideration, we've decided not to implement this functionality.

PR #4857, opened 2020-06-16
#4850 Added inheritance support mapping for base entity

denisburlacu:
Right now from type has DeclaredOnly binding gFlags which means we will get just declared properties form Current Instance. From some reason there are cases when we want to have extend base entity and based on this we will get all writable


exileDev:
We don't think this a good idea. It must be a controlled process

PR #4861, opened 2020-06-17
Fix #4828 Allow to access the password recovery when store closed

### LINKED ISSUE #4828: Allow to access the "password recovery" page when a store is closed
The "password recovery" page should be accessible when Store is Closed. Administrators are allowed to access pages when the store is closed but not if they have forgotten their password.

PR - https://github.com/nopSolutions/nopCommerce/pull/4861

denisburlacu:


DmitriyKulagin:
@denisburlacu Thanks a lot.

PR #4869, opened 2020-06-20
Fix currency exchange rate

cybermsv:
Снимок
1. Getting live rates by ECB provider produces an error when using the Russian interface (for example).
2. Bug №1 revealed a new bug in the Admin/CurrencyController:  uninitialized controls appeared.


DmitriyKulagin:
@cybermsv Thanks a lot.

Issue #4834, opened 2020-06-09
Problem with using different versions of libraries


skoshelev:
There's a problem when different versions of one library are used by other different plugins

![image (2)](https://user-images.githubusercontent.com/16875599/84145806-3a9be180-aa63-11ea-8bfe-26ac930f150b.png)

see details [here](https://www.nopcommerce.com/en/boards/topic/81849/43-couple-of-issues-when-building-source-code-on-azure-devops-build-pipeline). Also see https://github.com/nopSolutions/nopCommerce/issues/4877

Good suggestion is [here](https://www.nopcommerce.com/en/boards/topic/95753/bug-in-nop-460-beta-two-external-different-dll-files#301994) (let's display a warning to do not load one of the plugins with a different version)

RomanovM:
Look at https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions

skoshelev:
Also see [this comment](https://www.nopcommerce.com/en/boards/topic/95753/bug-in-nop-460-beta-two-external-different-dll-files#301994)

skoshelev:
see also [this one](https://learn.microsoft.com/en-us/dotnet/core/tutorials/creating-app-with-plugin-support)

skoshelev:
Closed #4834

PR #4876, opened 2020-06-23
Bugfix for plugin loading

sidatacom:
If a plugin have different nuget package references you got an error that the asselbly is not found.

The problem is explaint here:

https://devnet.kentico.com/articles/referencing-multiple-versions-of-the-same-assembly-in-a-single-application

A good plugin wrapper i found here:

https://natemcmaster.com/blog/2018/07/25/netcore-plugins/

I fixed the problems first but the acutall implementation is horrible.
So please refactor and implement also a isolation feature.
Actually all mvc controllers from all plugins will be loaded without control.


sidatacom:
Realy no response ?

skoshelev:
Hi @sidatacom. Sorry for so long waiting time. We are very grateful for your help. We still have an open issue https://github.com/nopSolutions/nopCommerce/issues/4834 in which we are looking for a solution to this problem. The bottom line is that we can’t accept your code yet, it boils down to the fact that with this approach, the plugin must come with all its dependencies and load all these dependencies into its domain, regardless of whether the same library is used somewhere else. Perhaps we do not quite understand how this is decided in your version. Suppose there are two plugins that use 10 of the same libraries as nopCommerce core, but each of the plugins uses a different version of one of them. After all, it is logical to keep only one modified version in the context of the plugin, since 9 others are already in memory. How do you propose to solve this issue with the reuse of resources

sidatacom:
That's wrong, we found a solution to load everything what we want as plugin. The solution is a hack because your solution of plugin loading is so bad.
We spend over 6 months in a real solution.
The problem of your plugin solution is that you use only the default assemblyloadcontext.
Also the loading and resolving of assemblies is very bad.
The biggest problem is that you scan all subfolders for dll's and you try to load everything.
We create a nop plugin to load nested solutions in the nop commerce solution so you can use nop as a SaaS shop for solutions. So we load over a /esi our pluginsolution completely isolated in the same host context and map it to this path. Wrote if you need help to implement I real good plugin loader solution.

skoshelev:
Hi @sidatacom. If you have a plugin system that is compatible with nopCommerce and doesn't have the drawbacks you mentioned, we would be very grateful if you could create a new PR and share it with us.

AHComp:
Of course this is the base logic of open source. I try to implement our logic from Nate in the current nop version.

PR #4887, opened 2020-06-24
Added fix for edit issue on a checkout attribute.

RalucaMch:
Added fix for edit issue on a checkout attribute when there's some other attribute with "checkboxes" type.


denisburlacu:
this can be closed because of #4896

PR #4892, opened 2020-06-25
Improved ImportResourcesFromXml, removed group by, added a safe dict

roman-gherman:


skoshelev:
Found a more simple way to improve performance, see https://github.com/nopSolutions/nopCommerce/commit/cefe82f74f860ae3145d415affd51b7e1475856a

Issue #4891, opened 2020-06-25
Add database type mapping for int64/long


AndreiMaz:
There is no type mapping in MigrationManager.cs for long type.

A line should be added to _typeMapping :
[typeof(long)] = c => c.AsInt64(),

Source: https://www.nopcommerce.com/en/boards/topic/82849/43-missing-database-type-mapping-for-int64long

See PR https://github.com/nopSolutions/nopCommerce/pull/4894

denisburlacu:
#4894

PR #4894, opened 2020-06-25
Add database type mapping for int64/long #4891

denisburlacu:
There is no type mapping in MigrationManager.cs for long type.

A line should be added to _typeMapping :
[typeof(long)] = c => c.AsInt64(),

Source: https://www.nopcommerce.com/en/boards/topic/82849/43-missing-database-type-mapping-for-int64long


exileDev:
Thanks!

PR #4895, opened 2020-06-25
Add support for minor versions of nop commerce

denisburlacu:


src/Libraries/Nop.Core/NopVersion.cs
```diff
@@ -9,5 +9,9 @@ public static class NopVersion
         /// Gets or sets the store version
         /// </summary>
         public static string CurrentVersion => "4.40";
+
+        public static string MinorVersion => "0";
+
+        public static string FullVersion => CurrentVersion + "." + MinorVersion;
```
skoshelev, line 15: Hi @denisburlacu Better to use string format: $"{CurrentVersion}.{MinorVersion}"

denisburlacu: ok will adjust, actually make sense to do for all fiedls here something like
public static string FullVersion {get} = $"{CurrentVersion}.{MinorVersion}", and based on this interpolation string will be executed only once

skoshelev: I agree, you may make changes and I'll merge your code.

denisburlacu: @skoshelev i pushed a new commit can you please have a look

Issue #4858, opened 2020-06-17
Cannot edit a checkout attribute when there's some other attribute with "checkboxes" type


AndreiMaz:
nopCommerce version: 4.30

Steps to reproduce the problem: 
1) Create new checkout attribute with 'checkboxes' control type and add some attribute values
2) Go to the sample 'Gift wrapping' checkout attribute
3) Click 'save' button

The following error is displayed - ![error](https://user-images.githubusercontent.com/5019111/84861907-cf619900-b07a-11ea-887f-c881f8d5967d.png)

See PR https://github.com/nopSolutions/nopCommerce/pull/4896

denisburlacu:
#4896

holydk:
Closed #4858

PR #4896, opened 2020-06-25
Edit Checkout Attribute issue from Admin

denisburlacu:
#4858
The issue is relatively simple, as was used `<nop-editor>` with a complex "asp-for" expression and in "NopEditorTagHelper.cs" we do
"var htmlOutput = _htmlHelper.Editor(For.Name, Template, new { htmlAttributes, postfix = Postfix });" so we leave to get type for input for Framework, but Framework is working with ModelExplorer, and with this complex expression from "asp-for" ModelExplorer can't get the right Type for the property, and will render always as "text".

as Is a problem just for Conditional Part, Checkbox will always have just 2 values, so any validation here dont make too much sense, so i think we are ok using `<input type=checkbox>`


holydk:
@denisburlacu  Thanks! I merged you changes via '4.30-bug-fixes' branch.

Issue #4193, opened 2019-12-02
Upgrade to AdminLTE 3


iyilm4z:
Based on bs4

[Here](https://github.com/ColorlibHQ/AdminLTE/releases) is the release and [here](https://adminlte.io/themes/dev/AdminLTE/index.html) is the demo.

Lymperis79:
I did some work for the refactoring..... if you interested let me know how to share it

DmitriyKulagin:
@iyilm4z You can share this via pull request.

iyilm4z:
@DmitriyKulagin this issue was just a notification, i have no pr, i wish i would have one:)

a-patel:
Just release new version [AdminLTE 3.0.2](https://github.com/ColorlibHQ/AdminLTE/releases)

iyilm4z:
The repo [admin-lte-dotnet](https://github.com/iyilm4z/admin-lte-dotnet) made by me shows how to implement Admin LTE 3 with  dotnet core.

DmitriyKulagin:
Closed #4193

Lymperis79:
Dear Dimitri,
I am glad even for this small help in such nice project!

Best Regards
Antonis

PR #4913, opened 2020-07-07
AdminLte3 integration in admin views

Lymperis79:
I integrated the AdminLte 3 - It might not be perfect but it is much better than the previous attempt that i made.


DmitriyKulagin:
Associated with #4193

DmitriyKulagin:
@Lymperis79 Thank you so much. Your input was a good starting point to understand how big changes would be required, but we couldn't leave this big task without doing it ourselves.

Issue #4363, opened 2020-03-25
Feature requests: Clear log with sliding windows 


Farshidgolkarihagh:
nopCommerce version: 4.2

Steps to reproduce the problem: currently there is a scheduled which truncates the log table on a set frequency. Will be very useful to modify this schedule/create new one which delete logs that are older than a set retention period. Number of days to keep logs will be defined via a configuration variable. 

Since I have done this change, I am happy to contribute.

AndreiMaz:
@Farshidgolkarihagh sure. you're more than welcome to contribute with pull request. thanks

PR #4915, opened 2020-07-07
#4363 - Clear logs with sliding window

Farshidgolkarihagh:
Clear log schedule with sliding window which is configured via a setting, Sorry for the delay in submitting the code.


manuel3108:
Why do you have put it on hold? I feel like this is really a missing feature in NOP

RomanovM:
See [comments](https://github.com/nopSolutions/nopCommerce/issues/4363#issuecomment-937549949), thanks.

PR #4957, opened 2020-07-31
Seo Optimization

mk148a:
I add up to 10 image and i am add meta description attribute for tags.


mk148a:
out of stock bug is fixed, and some microdata problems fixed

RomanovM:
Thanks. It'll be more acceptable if you split commits into separate pull requests each for its own issue.

Issue #4919, opened 2020-07-08
Display more user-friendly error when uploading non-picture file as a picture


AndreiMaz:
Steps to reproduce the problem: 
1. Upload any non-picture file (e.g. XML or .cs) as a picture. For example, on a product details page 
2. It gives a correct message that uploading failed. But also, 'Object reference not set to an instance of an object.' throwing error on picture controller side.

Let's handle the exception so it's not thrown. There should be just an error text without an unhandled exception 
![download](https://user-images.githubusercontent.com/5019111/86886520-c440dc80-c0ff-11ea-9869-c83c6e70e069.png)

Source: https://www.nopcommerce.com/en/boards/topic/84221/picture-controller-in-get-error-object-reference-not-set-to-an-instance-of-an-object

badijm:
Hi, I can grab this item if available.

badijm:
@AndreiMaz  Hi, I've send the PR. Lemme know if it works for ya.

AndreiMaz:
@badijm Please create the PR at https://github.com/nopSolutions/nopCommerce/pulls and share the link here

badijm:
> @badijm Please create the PR at https://github.com/nopSolutions/nopCommerce/pulls and share the link here

https://github.com/nopSolutions/nopCommerce/pull/4963

badijm:
@AndreiMaz  https://github.com/nopSolutions/nopCommerce/pull/4963

holydk:
Closed #4919

PR #4963, opened 2020-08-03
Display more user-friendly error when uploading non-picture file as a picture

badijm:
Display more user-friendly error when uploading non-picture file as a picture 

#4919 

Capture


holydk:
@badijm Thanks! I also added the some extensions and merged changes via '4.30-bug-fixes' branch.

Issue #4540, opened 2020-04-11
Hide the "register" link in the header when registration is disabled


Lymperis79:
nopCommerce version: 4.30

Steps to reproduce the problem: 
It is not a bug but a small suggestion. When the registration is disabled the registration link in the HeaderLinks page can be headen completely because there is no fuction of the link button!  
One way to do that is by creating a bool in the HeaderLinksModel and change the value by checking the _customerSettings.UserRegistrationType in the commonmodelfactory. From there the value will pass it to the HeaderLinks\default page file and hide the button link.

badijm:
Hi, I can pick this one too.

badijm:
@AndreiMaz I've sent the pull request. Can u take a look? Also, I have bunch others open PRs to @nopCommerce

holydk:
Closed #4540

PR #4978, opened 2020-08-12
Fix: Hide the "register" link in the header when registration is disabled #4540

badijm:
https://github.com/nopSolutions/nopCommerce/issues/4540

### DIFF
--- a/src/Presentation/Nop.Web/Factories/CommonModelFactory.cs
+++ b/src/Presentation/Nop.Web/Factories/CommonModelFactory.cs
@@ -356,6 +356,7 @@ namespace Nop.Web.Factories
 
             var model = new HeaderLinksModel
             {
+                IsRegistrationDisabled = _customerSettings.UserRegistrationType == UserRegistrationType.Disabled,
                 IsAuthenticated = _customerService.IsRegistered(customer),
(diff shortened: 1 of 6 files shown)


Lymperis79:
In case the registration is disabled in the login.cshtml left block there is a text that registration is not allowed! I think that this is not required and it should be only the login part.

badijm:
@Lymperis79 Sorry, I didn't quite get your point. Can you provide code example ?

Lymperis79:
in this path \src\Presentation\Nop.Web\Views\Customer\Login.cshtml the lines 26 till 33 are giving the message that the registration is disabled. If there is no way to register it is unnecessary information :D

badijm:
@Lymperis79 Got it. Thanks for that.

Lymperis79:
I found one more part that has to be hidden when registration is disabled!
in this path \src\Presentation\Nop.Web\Views\Customer\Login.cshtml lines 41 till 51! they show a text and a button for registration! not necessary!

badijm:
@Lymperis79 Alright. Pushed new changes.

badijm:
Hi, @Lymperis79 any updates on it?

holydk:
Closed #4978. Changes [here](https://github.com/nopSolutions/nopCommerce/commit/08574683f0795fdfe247cf3c55dd07183d738d20).

Issue #4966, opened 2020-08-04
EntityRepository.Update(entities) updating records one by one


KostaVlev:
Is there any particular reason  [EntityRepository.Update(entities)](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Libraries/Nop.Data/EntityRepository.cs#L105) is performing update in foreach loop?

```
public virtual void Update(IEnumerable<TEntity> entities)
{
     if (entities == null)
     throw new ArgumentNullException(nameof(entities));

      foreach (var entity in entities)
      {
            Update(entity);
      }
}
```

Seems like this can be done in bulk update as noted [here](https://github.com/linq2db/linq2db/issues/806)

> If you need to insert or update records from list using primary key fields, it could be achived with following merge command:

``` 
db.GetTable<T>()
    .Merge()
    .Using(yourList) // here you pass your list
    .OnTargetKey() // make decision on primary key
    .UpdateWhenMatched() // update all fields from mapping
    //.InsertWhenNotMatched() // insert all fields from mapping
    .Merge();
```

KostaVlev:
Adding `UpdateEntities<TEntity>(IEnumerable<TEntity> entities)` to the `BaseDataProvider.cs`

```
/// <summary>
/// Updates records in table, using values from entity parameter. 
/// Record to update identified by match on primary key value from obj value.
/// </summary>
 /// <param name="entities">Entities with data to update</param>
/// <typeparam name="TEntity">Entity type</typeparam>
public void UpdateEntities<TEntity>(IEnumerable<TEntity> entities) where TEntity : BaseEntity
{
        using var dataContext = CreateDataConnection();
        dataContext.Merge(entities);
}
```

and then updating the `EntityRepository.Update(IEnumerable<TEntity> entities)` with

```
public virtual void Update(IEnumerable<TEntity> entities)
{
       if (entities == null)
           throw new ArgumentNullException(nameof(entities));

       using (var transaction = new TransactionScope())
       {
            _dataProviderExtender.UpdateEntities(entities);
            transaction.Complete();
       }
}
```

Seems to be working. I have tetsed this with Sql.

@skoshelev if this looks good I can create PR?

skoshelev:
Hi @KostaVlev. I didn't test this suggests but it looks well, if you create PR it will be great

skoshelev:
[...]

PR #4979, opened 2020-08-12
EntityRepository.Update(entities) updating records one by one #4966

KostaVlev:
Refactoring `EntityRepository.Update(entities)` to perform update in batch.

#4966


exileDev:
We can't use it because MySQL & PostgreSQL don't [support](https://linq2db.github.io/articles/sql/merge/Merge-API.html#supported-databases) MERGE statement.

PR #4983, opened 2020-08-13
Fix orderby issue with linq2db 3.0.1

Shane32:
This patch fixes the following error when nopcommerce is used with linq2db 3.0.1:
```
A column has been specified more than once in the order by list. Columns in the order by list must be unique.
```
I searched Nop.Services for uses of `OrderBy`, but there may be similar issues elsewhere.

There should be no ill effects for prior versions of linq2db


exileDev:
Hi @Shane32! We already have the [branch](https://github.com/nopSolutions/nopCommerce/tree/Upgrade-to-LINQtoDB-v3) with fixing this issue. And we merge it with the develop branch soon.

Issue #5018, opened 2020-09-08
Azure blob storage NuGet package deprecated


wforney:
nopCommerce version: 4.4/source code from develop

Steps to reproduce the problem: The NuGet package for Azure blob storage is deprecated. There is a new one. I have created an update for this and adding the root files to the solution under a Solution Items folder. I will create a PR for it in a moment.

skoshelev:
Fixed by https://github.com/nopSolutions/nopCommerce/commit/4c562b00d56a32767ffdf6028761d9998a9da731 commit

Closed #5018

PR #5019, opened 2020-09-08
5018 azure storage nuget deprecated

wforney:
See issue #5018 ... The Azure storage NuGet package in develop branch is deprecated. This replaces it with the newer v12 one. Also this adds a solution items folder with the top level files in it to the solution so you can easily find them in the solution explorer.


tombeuckelaere:
Hi,

Because i'm having an issue with the azure blob storage functionality (https://github.com/nopSolutions/nopCommerce/issues/5050), I was referred to this pull request to adapt the changes.
I've downloaded the current version of develop and applied your pull request on this.

There are 2 issues with your pull request which prevent working functionality:
- Nopconfig no longer exists and should be replace with appsettings i gues
- When the previous functionality is fixed, meaning the settings can be fetched in the azurepictureservice, the adding of images doesn't work. It throws an exception because the _blobcontainerclient is null.

Can you take a look at this since this is a blocking issuse?

Thanks

2020-10-01-Nopcommerce-1

tombeuckelaere:
After further investigation it seems 2 fields should be made static since they are only initialized once. The check on _isInitialized prevents new initialization of those fields when an instance is needed.

**Before:**
```
private BlobContainerClient _blobContainerClient;
private BlobServiceClient _blobServiceClient;
```

**After**
```
private static BlobContainerClient _blobContainerClient;
private static BlobServiceClient _blobServiceClient;
```

RomanovM:
Thanks for your contribution, we've used your suggestion in [this commit](https://github.com/nopSolutions/nopCommerce/commit/4c562b00d56a32767ffdf6028761d9998a9da731) when updating to .NET 5.

Issue #5023, opened 2020-09-09
Accessibility - No Skip Link/Bypass Block


dylanhthomas:
nopCommerce version: 4.3

## Steps to reproduce the problem:

1. Navigate to a product page of demo site (as an example): https://demo.nopcommerce.com/build-your-own-computer
2. Hit 'tab' (and count the number of tabs) until you get to the first link of the page's unique content (in this case the image gallery)

## Result

It takes 45 tab clicks to get to the content of the page because the global navigation has to be traversed first. This happens on every page.

## Expected Results

User is able to skip repeated content if desired by having a "skip link" (also know as a "bypass block") available as the first item in the tab order.
User should be able to:
- tab once
- follow skip link (enter)
- tab once
- have item selected

## Fix

Implement a skip link. This has 3 components:

1. - the link itself
2. - the target anchor it links to
3. - the styles to only show the skip link when focused and position it appropriately

### 1 - Link
in _src/Presentation/Nop.Web/Views/Shared/_Header.cshtml_

`<div class="header">
    @await Component.InvokeAsync("Widget", new { widgetZone = PublicWidgetZones.HeaderBefore })
`
becomes

`<div class="header">
    <a class="skip" href="#main">Skip Navigation</a>
    @await Component.InvokeAsync("Widget", new { widgetZone = PublicWidgetZones.HeaderBefore })
`

### 2 - Anchor
in: _src/Presentation/Nop.Web/Views/Shared/_Root.cshtml_ 

`<div class="master-wrapper-page">`

changes to: 

`<div class="master-wrapper-page" id="main" role="main">`

(the id enables the link. the role applies the correct aria landmark to the main section of the page)

### 3 - Styles

Following is added to:
_src/Presentation/Nop.Web/Themes/DefaultClean/Content/css/styles.css
src/Presentation/Nop.Web/Themes/DefaultClean/Content/css/styles-rtl.css_

```
a.skip {
    position: absolute;
    top: -1000px;
    left: -1000px;
    height: 1px;
    width: 1px;
    text-align: left;
    overflow: hidden;
}
a.skip:active, 
a.skip:focus, 
a.skip:hover {
    position: initial;
    top: 0;
    left:0;
    display: block;
    height: auto;
    margin: 3px 0;
    overflow: visible; 
    padding: 2px;
    color: #444444;
    background: #ffffff;
    text-decoration: underline;
    text-align: center;
    width: 100%;
    font-size: 16px;
}
```

[...]

PR #5034, opened 2020-09-15
Add Skip link/Bypass block

dylanhthomas:
This resolves the missing skip link issue. If you navigate to any page, the link will be invisible. If you hit tab one time, you will see the below link. If you tab again, it will disappear. If you click or hit enter on the link, the focus will jump to the next focusable link in the main section.

skiplink

### DIFF
--- a/src/Presentation/Nop.Web/Themes/DefaultClean/Content/css/styles.css
+++ b/src/Presentation/Nop.Web/Themes/DefaultClean/Content/css/styles.css
@@ -80,6 +80,33 @@ a {
 a img {
 	opacity: 0.99; /*firefox scale bug fix*/
 }
+a.skip {
+    position: absolute;
+    top: -1000px;
+    left: -1000px;
+    height: 1px;
+    width: 1px;
+    text-align: left;
+    overflow: hidden;
+}
+a.skip:active, 
+a.skip:focus, 
+a.skip:hover {
+    position: initial;
+    top: 0;
+    left:0;
+    display: block;
+    height: auto;
+    margin: 3px 0;
+    overflow: visible; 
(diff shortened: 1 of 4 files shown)


dylanhthomas:
should close #5023

mariannk:
Hi @dylanhthomas,

I'm testing your solution and when I click "Enter" on the added link it doesn't jump to the main content block. Am I missing anything?

dylanhthomas:
You need to hit 'tab' again to select the next link. the link moves the focus to the div (which isn't focusable)
This is the expected behavior. 

There is a bit of an issue testing on the homepage, since the nivo slider has it's own focus styles and does not inherit from the main stylesheet (I will be fixing this in the future).

I've attached a video demoing on the homepage and an internal page.

Let me know if I can help more or if this does the trick.

mariannk:
@Add Skip link/Bypass block #5034 

> You need to hit 'tab' again to select the next link. the link moves the focus to the div (which isn't focusable)
> This is the expected behavior.
> 
> There is a bit of an issue testing on the homepage, since the nivo slider has it's own focus styles and does not inherit from the main stylesheet (I will be fixing this in the future).
> 
> I've attached a video demoing on the homepage and an internal page.
> 
> Let me know if I can help more or if this does the trick.
> 
> https://youtu.be/6UAPDHVLrxU

I recorded a video from my PC to show you that for some reasons it doesn't work for me https://youtu.be/bdyX0X-quQo

RomanovM:
Thanks. We've used your suggestion in [this commit](https://github.com/nopSolutions/nopCommerce/commit/55d2d9deb786c155f76bec59d29b4cf9521c0ae4).

PR #5045, opened 2020-09-23
Update CheckoutModelFactory.cs

nullbg:
In commit: 2a884e843426e949425713d4715c9446d1c588b6 on Aug 13, 2019, check for null country is removed, and that causing the problem when shop not using counters.
image


nullbg:
Bad one.

Issue #2090, opened 2017-02-05
Edit order details. Automatically update products fields (price, qty, total) when editing 


AndreiMaz:
Continuation of task (https://github.com/nopSolutions/nopCommerce/issues/8). Right now we update order totals. Product quantity will automatically calculate the total of the product itself. When I manually add the correct price for the product totals then it also has the correct order totals.

Make it configurable (as it could depends on some complex price rules). And I think it's better to implement it with javascript

Source: http://www.nopcommerce.com/boards/t/46077/order-price-doesnt-update-when-manually-add-a-product-to-order-from-admin.aspx#182880

jack0fshad0ws:
Hi, I think for consistency it'd be most useful to update product total when qty changed when "Auto update order totals" is on.
Thanks

cldbrr:
Hi, Is there update on this case?

PR #5095, opened 2020-10-15
Fix issue in updating order items quantity and price

ahmedfe:


RomanovM:
Thanks, but there are many more factors to consider when updating than just the quantity and unit price. The [related issue](https://github.com/nopSolutions/nopCommerce/issues/2090) is still open and we are thinking about how best to implement it.

ahmedfe:
Hi @RomanovM  ,
this update fixe a bug of not considering the quantity while update the total , the fix tested and work correctly .

PR #5104, opened 2020-10-19
Update defaultResources.nopres.xml

margelef:
fix typo in NoSelectedShippingOption.  User selects Ship To address, not Ship From.


holydk:
@margelef Hi, I also added the locale into migration. Thanks.

Issue #5096, opened 2020-10-16
Make a [store] as a soft deleted entity 


RomanovM:
nopCommerce version: develop

Store must implement the `ISoftDeletedEntity` and just be marked as Deleted when deleted to avoid losing of the related mapping data.

See also: https://github.com/nopSolutions/nopCommerce/issues/5035

badijm:
Hi, I can work on that, looks somple enough.

exileDev:
Closed #5096

PR #5110, opened 2020-10-23
Fix of 5096: Make Store as a soft deleted entity

badijm:
Fix of #5096 

* Inherit from **ISoftDeletedEntity** & add new column on UI, view model & resource file.


exileDev:
Hi, @badijm! But what about migrations? This issue isn't trivial. The first usage of the Store entity occurs before any migration has run. So we can't run a nopCommerce upgrade under the migration process now.

badijm:
I see your didn't expect it. I'm not sure currently how to handle this. Convert this to draft, will investigate when take time.

Issue #5102, opened 2020-10-19
Drop stored procedures


RomanovM:
nopCommerce version: develop

We could remove our custom stored procedures. This will allow not to depend on the specific implementation and SQL specifics for different db providers, we can keep the logic of the method in the code and use linq.

exileDev:
Closed #5102

Issue #5097, opened 2020-10-18
Don't use linq2db outside of Nop.Data


RomanovM:
nopCommerce version: develop

Remove direct access to LinqToDB namespace in all projects except Nop.Data, we should encapsulate all appropriate methods and properties in Nop.Data and use them in other projects. 
At the moment such usings remain in [PictureService](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Libraries/Nop.Services/Media/PictureService.cs#L6) and [BaseNopValidator](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web.Framework/Validators/BaseNopValidator.cs#L5).

exileDev:
Closed #5097

PR #5119, opened 2020-10-27
INopDataProvider ExecuteStoredProcedure custom timeouts

dfar-io:
In previous versions of NopCommerce, `IDBContext` allowed for setting custom timeouts for calls using ExecuteStoredProcedure(). This was removed in `INopDataProvider`, but I've had to custom add it into the codebase to allow for running a few client processes for some custom plugins.

I've set up this PR to be as minimally invasive as possible, to allow for previous usage, but allowing for setting a timeout on a particular stored procedure call that supersedes the `NopConfig` value.

### DIFF
--- a/src/Libraries/Nop.Data/BaseDataProvider.cs
+++ b/src/Libraries/Nop.Data/BaseDataProvider.cs
@@ -89,24 +89,25 @@ namespace Nop.Data
         /// <summary>
         /// Creates the database connection
         /// </summary>
-        protected virtual DataConnection CreateDataConnection()
+        protected virtual DataConnection CreateDataConnection(int timeout = 0)
         {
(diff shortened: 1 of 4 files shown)


exileDev:
We can't accept your suggestion now. Since we aren't sure what a way kind to choose to continue Stored procedures support in general. At this moment, we have plans to remove our SPs out of the box: #5102, #5097

dfar-io:
Any chance you're planning to keep the ExexuteNonQuery() (and related) methods on INopDataProvider? I could add the same timeout functionality to that method instead.

exileDev:
I'm not sure what we'll keep the destination place of these methods, but we are thinking about how to implement it yet.

dfar-io:
Sounds good, thank you for the consideration - will close for now and revisit in 4.40

Issue #4669, opened 2020-05-03
Same products adding into order instead of increase qty at order edit page


sangeetshah:
nopCommerce version: 4.3rc

Steps to reproduce the problem: 

1) Go to order edit page
2) Add same product into products grid 

![screenshot-nopcommerce430rc nopadvance com-2020 05 03-22_18_29](https://user-images.githubusercontent.com/10364140/80920261-4de4ce80-8d8c-11ea-9919-11e3263cf813.png)

Also a little bit related work item - https://github.com/nopSolutions/nopCommerce/issues/1920 (could be implemented together, with one workflow)

RomanovM:
It's by design. Perhaps we should reconsider this behavior.

AndreiMaz:
After some consideration we've decided not to implement this functionality out of the box.

PR #5163, opened 2020-11-16
Adding duplicate items to order should increment quantity

noncenz:
This should resolve issue #4669. 

These changes tested OK for me with `ordersettings.autoupdateordertotalsoneditingorder` both set and unset.

I followed the pattern used to prevent duplicate items in the shopping cart, but I'm happy to get feedback if there is a better approach.


RomanovM:
Decided not to implement this functionality, see [comments](https://github.com/nopSolutions/nopCommerce/issues/4669#issuecomment-967747752).

Issue #4268, opened 2020-01-22
Upgrade to .NET 5


RomanovM:
Let's do once released. But don't start the upgrade until it's officially released and all third-party libraries support .NET 5

dizx:
Good idea

Daniel-Guenter:
This would be really useful for us.  We have existing libraries that we need to link for a plugin that are not compatible with .net Core.  I am hoping that this would allow us to reference them, or at least reduce the number of projects that we have to change in order to reference them.

MafdyAzer:
@AndreiMaz  👍 
There is no longer a need for ".NET Standard" after" .Net 5" release, It'll be removed, only "net5.0"

![image](https://user-images.githubusercontent.com/57217759/83885557-cdc4d680-a746-11ea-8cb6-47865d0ddd9e.png)

https://devblogs.microsoft.com/dotnet/announcing-net-5-preview-4-and-our-journey-to-one-net/

AndreiMaz:
@MafdyAzer You're absolutely right. Now there's no need to consider ".NET Standard"

ryanbarlow1:
.NET 5 is currently scheduled to release in November 2020. Will version 4.40 move to .NET 5 or stay on .NET Core 3.1 (LTS)?

ghd258:
.NET 5 is release

ryanbarlow1:
I should clarify. Back when nopCommerce 4.2 was released, .NET Core 2.2 reached End of Support before nopCommerce 4.3 was released (.NET Core 2.2 lost support December 2019 and nopCommerce 4.3 wasn't released until May 2020). Those five months people were using nopCommerce 4.2 in Production with the .NET Core version not receiving security updates. To remediate this, should nopCommerce only upgrade to LTS versions of .NET, which have a much longer support lifespan? If not, how can this be prevented from happening in the future?

ghd258:
Upgrade to. Net 5 is very simple

DmitriyKulagin:
Closed #4268

RobinSCU:
@DmitriyKulagin did you experience any Performance improvments when you updated to .net core 5.0?

DmitriyKulagin:
@RobinSCU Yes, we have seen some performance improvements over the previous version. But since there were several more significant changes that affect performance, it is difficult to say how much the increase was in terms of migration to .NET 5.

SpeakOutLoud:
[...]

PR #5193, opened 2020-11-28
#4268 upgraded to net5.0, upgraded libs and azure client

AdaskoTheBeAsT:
#4268 upgraded to net5.0, upgraded libs and azure client


AdaskoTheBeAsT:
hi - unfortunately after upgrade it requires additional work - it seems that even if connection is not know SlugRouteTransformer is instantiated  and it causes NRE due to depndencies up to DataProviderManager.DataProvider

DmitriyKulagin:
@AdaskoTheBeAsT We are currently working on this task. We want to implement it ourselves in order to fully control the migration process and guarantee stable work. But thanks anyway for your help.

PR #5221, opened 2020-12-13
Convert boolean value to more user-friendly 'Yes' / 'No' string values

badijm:
* Convert boolean value to more user-friendly 'Yes' / 'No' string values

Note: I've moved convertion to extension to keep razor view clean.


RomanovM:
@badijm Actually there is no need to add a separate extension method for this. We already have an [example](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web/Areas/Admin/Views/CustomerRole/_CreateOrUpdate.cshtml#L144) of such a convertion. I [did the same](https://github.com/nopSolutions/nopCommerce/commit/784b6fc07a906bd695d739a815a744e3e94ddb97) for this case. 
Anyway, thanks for your contribution.

badijm:
@RomanovM thanks for clarifying it.

PR #5243, opened 2020-12-23
added throwing of exception when pageSize <= 0

s3500659:
added the throwing of exception in the case where pageSize is set to <= 0


RomanovM:
Thanks for you contribution, but as suggested in the issue description, we decided [to force use 1](https://github.com/nopSolutions/nopCommerce/commit/fd0cb0a7bbffaf223b0067656ee925a6e6405451) as the minimum page size.

PR #5302, opened 2021-01-25
add admin -> sidebar -> navigation > active item highlight

bo-bac:
just UX sugger 

image


AndreiMaz:
Thanks. But we prefer our design

Issue #3914, opened 2019-05-31
Different language for Admin area


RomanovM:
I was hoping that the new version 4.2 will have option to select different language for admin area. The reason for it is simple: not always is the translation correct (mainly in admin area) and most help documents are in English. If you do not want to offer English to your users but you as Admin want to use English to see correct terms you have no choice but enable English for public. I think that the language dropdowns in public store and in administration should be independent and the language grid should have another check box "Use in Admin".

Source: https://www.nopcommerce.com/boards/t/66117/suggestion-different-language-for-admin-area.aspx

mafdy:
One more thing, this mix between the selected language of the public store and the admin area cause Thousands of warnings on System/Log (182571 records of logs about Resource string warnings on my store logs right now 😆).
To be more clear, this happens when selected language for public store applied to admin area without being translated. So, I think we need also to separate them in the settings which are cached in the browser.

bo-bac:
Hello gentlemen,

if Dmitriy is not started the feature and you don't mind I would try to do this one.
I will consult with Dima if there are a questions here and ask him to review my PR. May be we will manage to add this to 4.4 
What do you think.

DmitriyKulagin:
Hello @bo-bac, 
Of course. Feel free to share your solution and we'll look at adding this feature in 4.40

bo-bac:
so proposed solution little bit differ from suggested but it almost nothing was changed)

implemented idea is just allow customer to select languages independently.
public (store) site has language selector and it works like it was implemented
admin area has its own selector and now it use
  - public value (if there is no its value selected)
  - its own if it is

what do you think

AndreiMaz:
After some consideration we've decided not to implement this functionality out of the box

P.S. Just import the English language pack with admin locales only to any other existing language

RomanovM:
See also https://www.nopcommerce.com/boards/topic/92829/multi-culture-admin-number-format-issue-when-switching-language

bellamatte:
Since 4.40 and KendoUI localization, not only the language (resource strings) changes, but also the culture and so the number format.
Switching language in the public store, might lead to unexpected and unwanted values edits in admin if the page was open before the language switch. In fact, the decimal separator could be different and misinterpreted.
Also, this might be not noticed, leading to confusion.

I hope this will be reopened.

PR #5335, opened 2021-02-11
add ability to select Admin area language independently of the public (store) language

bo-bac:
@DmitriyKulagin please review

implemented idea just allows customer to select languages independently.
public (store) site has language selector and it works like it was implemented 
admin area has its own selector and now it use
   - public value (if there is no its value selected)
   - its own if it is
 
#3914


RomanovM:
We still decided not to implement this feature. 
Anyway, thanks for your suggestion, it will be useful for customization purposes.

Issue #5089, opened 2020-10-09
Low stock quantity


DmitriyKulagin:
Add low stock quantity field in model, which now will allow users to configure when client will see 'low stock' availability status.

Previous 'in stock' and 'out of stock' nop statuses are nice, but most store engines provides also 'low stock' state to inform clients they should consider buying product now (because product may disappear in near future from store).

PR #5364, opened 2021-02-22
Feature low stock quantity

ArturSprinter:
I finally got some time to finish long time opened issue:
https://github.com/nopSolutions/nopCommerce/issues/5089

I hope the change will be included in nopCommerce 4.40 😃


ArturSprinter:
@RomanovM I wonder label you applied to this pull request is bit inconsistent with the one we can find on related issue https://github.com/nopSolutions/nopCommerce/issues/5089 (assigned by @DmitriyKulagin and @AndreiMaz)

To make everything clean you may link mentioned issue with this pull request 🙂

DmitriyKulagin:
@ArturSprinter We are now ready to include this task in the 4.50 roadmap. If you update your code with the latest changes in the working branch, we will do a full code review. For now, please clear all conflicts so we can continue.

DmitriyKulagin:
@ArturSprinter We did a preliminary code review, and here's what we would like to see in the solution:
1. In the pull request, 11513 lines were added and 11321 deleted, resulting in about 200 lines of changes, let's make it so that they are the only ones. Now modules such as InstallationService.cs, ProductModelFactory.cs have been completely reformatted. This makes it impossible to track changes for this task. And please merge all final changes into one commit before submitting for review.
2. On the product editing interface, the "Low stock activity" field must be moved up and made dependent on the "Display availability" setting
3. All new locales must be registered in Nop.Web.Framework.Migrations.UpgradeTo450.LocalizationMigration, and all new settings in Nop.Web.Framework.Migrations.UpgradeTo450.SettingMigration, otherwise users who will upgrade from earlier versions will not receive these changes.
4. With the "Display stock quantity" option disabled and the stockQuantity <product.LowStockQuantity condition, the "Low stock" message is displayed. But if you enable the "Display stock quantity" setting, then under the same conditions we will see "X in stock" in red. Instead, it will be much more informative to tell the user so - "X in stock (Low stock)". Because the indication by color alone is not enough, customers will not know about these colors, how many there may be, and what they mean.

DmitriyKulagin:
We decided to implement this on our own, since there was no feedback from you for a long time.

PR #5476, opened 2021-02-28
Configuration runner code optimization

Olek-HZQ:
For the first installation, configure all possible database types, and use the currently used database type for the installed database.


exileDev:
Thank you for the suggestion. Solved in #7608.

Issue #5066, opened 2020-10-02
Replace EPPlus


DmitriyKulagin:
[EPPlus](https://github.com/JanKallman/EPPlus) will from version 5 switch license from LGPL to Polyform Noncommercial 1.0.0 license. We need to replace it as its further update is not possible due to its commercial license.

Let's think about what alternatives we can use:
- [NPOI](https://github.com/dotnetcore/NPOI)
- [ExcelDataReader](https://github.com/ExcelDataReader/ExcelDataReader)
- [Open XML SDK](https://github.com/officedev/open-xml-sdk)

What else?

JadynWong:
ClosedXml[ClosedXML](https://github.com/ClosedXML/ClosedXML)

It is very easy for operating excel.
ClosedXml is base on OpenXml.
The original OpenXml is very powerful, but not easy to operate.

DmitriyKulagin:
Closed with commit https://github.com/nopSolutions/nopCommerce/commit/fbc65ba7df47e99ef58f01bf2a8519924297683c

PR #5501, opened 2021-03-05
Replace EPPlus with ClosedXML

JadynWong:
Reslove #5066


DmitriyKulagin:
@JadynWong Thank you very much, this is a good job!

Issue #5525, opened 2021-03-13
[Code Analysis] V3093 The '&' operator evaluates both operands. Perhaps a short-circuit '&&' operator should be used instead.


badijm:
Hi, this is the minor code issue revealed by static code analysis: 

`V3093 The '&' operator evaluates both operands. Perhaps a short-circuit '&&' operator should be used instead. FilePermissionHelper.cs 32`

**Update**: Created PR to fix it.

bigboybamo:
Is this issue still open because it looks like it has been corrected in FilePermissionHelper.cs

badijm:
Currently in `develop` still using '&' operator, not '&&' [CheckUserFilePermissions method]

PR #5526, opened 2021-03-13
Fix [Code Analysis - PVS Studio] V3093: Short-circuit '&&' operator should be used instead.

badijm:
#5525


seraishi:
Thanks for your decision.

Issue #5284, opened 2021-01-19
Admin Tables Footer not rendering on Child Table


purplepiranha:
nopCommerce version: 4.30

Steps to reproduce the problem: 
Add 'FooterCallback' to a ChildTable within the admin section

The footer is never rendered as the child table does not have a 'tfoot' section. As stated in Table.cshtml you need to add the footer before the table is created as DataTables doesn't provide a method for creating a footer.

Workaround:
In Table.cshtml after
```var modelName = ReplaceName(curModel.Name);
        if (curModel.ChildTable != null)
        {
```
insert
```// Fix for Footer on ChildTable
            var footerHtml = "";
            if (curModel.ChildTable.FooterColumns > 0)
            {
                @for (int i = 0; i < curModel.ChildTable.FooterColumns; i++)
                {
                    footerHtml = string.Concat(footerHtml, "<td></td>");
                }

                footerHtml = string.Concat("<tfoot><tr>", footerHtml, "</tr></tfoot>");
            }
```
and then change
`return childTable = '<table id="child' + d.@(curModel.PrimaryKeyColumn) + '" class="table table-bordered table-hover dataTable" width="100%" style="padding-left:2%;"></table>';`
to
`return childTable = '<table id="child' + d.@(curModel.PrimaryKeyColumn) + '" class="table table-bordered table-hover dataTable" width="100%" style="padding-left:2%;">@(Html.Raw(footerHtml))</table>';`

purplepiranha:
There also seem to be other issues with child tables. For example the Inline Edit doesn't seem to display the input box when the EditType is EditType.String. Also the parent table refreshes when a delete is performed, rather than just the child table.

DmitriyKulagin:
@purplepiranha We have made some changes to improve this functionality, you can find them in the develop branch. Please, check this commit -https://github.com/nopSolutions/nopCommerce/commit/4de48951e224c1c203568d5f77dca86c016d2707

purplepiranha:
Thanks @DmitriyKulagin. Unfortunately it doesn't appear to address those specific issues.

DmitriyKulagin:
@purplepiranha Then describe the problems in more detail so that we can investigate them. Alternatively, you can submit a pull request if you already have a solution.

purplepiranha:
@DmitriyKulagin -  the fix for the footer (as detailed above) is currently in an Azure repo, but when I get a chance I'll add it to a fork here and submit a pull request.

[...]

PR #5533, opened 2021-03-16
#5284 Admin Tables Footer not rendering on Child Table

purplepiranha:
This fixes the main issue described in #5284 Admin Tables Footer not rendering on Child Table.

It does not however address the other issue relating to inline editing/deleting, which appears to be more complicated to rectify.


purplepiranha:
Closed - Pushed some random commits by accident

Issue #5540, opened 2021-03-16
Azure Blob Storage error


untiedshoes:
nopCommerce version: 4.40

Steps to reproduce the problem: Enable Azure Blob storage configuration, test the site and various pages. Turn it off, view various pages, then enable it again.

The specified blob already exists.
RequestId:9c77bc34-301e-000b-6b74-1a8260000000
Time:2021-03-16T14:56:35.8995772Z
Status: 409 (The specified blob already exists.)
ErrorCode: BlobAlreadyExists

Headers:
Server: Windows-Azure-Blob/1.0,Microsoft-HTTPAPI/2.0
x-ms-request-id: 9c77bc34-301e-000b-6b74-1a8260000000
x-ms-client-request-id: b9186d65-41cc-4f2f-8512-f10a773942fe
x-ms-version: 2020-04-08
x-ms-error-code: BlobAlreadyExists
Date: Tue, 16 Mar 2021 14:56:35 GMT
Content-Length: 220
Content-Type: application/xml

at Azure.Storage.Blobs.BlobRestClient.BlockBlob.UploadAsync_CreateResponse(ClientDiagnostics clientDiagnostics, Response response)
   at Azure.Storage.Blobs.BlobRestClient.BlockBlob.UploadAsync(ClientDiagnostics clientDiagnostics, HttpPipeline pipeline, Uri resourceUri, Stream body, Int64 contentLength, String version, Nullable`1 timeout, Byte[] transactionalContentHash, String blobContentType, String blobContentEncoding, String blobContentLanguage, Byte[] blobContentHash, String blobCacheControl, IDictionary`2 metadata, String leaseId, String blobContentDisposition, String encryptionKey, String encryptionKeySha256, Nullable`1 encryptionAlgorithm, String encryptionScope, Nullable`1 tier, Nullable`1 ifModifiedSince, Nullable`1 ifUnmodifiedSince, Nullable`1 ifMatch, Nullable`1 ifNoneMatch, String ifTags, String requestId, String blobTagsString, Boolean async, String operationName, CancellationToken cancellationToken)
   at Azure.Storage.Blobs.Specialized.BlockBlobClient.UploadInternal(Stream content, BlobHttpHeaders blobHttpHeaders, IDictionary`2 metadata, IDictionary`2 tags, BlobRequestConditions conditions, Nullable`1 accessTier, IProgress`1 progressHandler, String operationName, Boolean async, CancellationToken cancellationToken)
[...]

PR #5541, opened 2021-03-16
Allow a user to allow Azure Blob Storage to overwrite, plus fix for #5540

untiedshoes:
Add a new media setting to allow a user to allow Azure Blob Storage to overwrite, plus fix for #5540


DmitriyKulagin:
@untiedshoes Thanks for the suggested solution, however we decided not to use the additional setting in _mediaSettings as overwriting is an expected action. See details here - https://github.com/nopSolutions/nopCommerce/commit/614450a9608bfd7402149245e63dfbaed9fcada0

PR #5559, opened 2021-03-21
Feature/#5555 disable add to basket button

gulaj:


RomanovM:
Thanks for the suggestion. Decided not to implement this functionality.

Issue #5584, opened 2021-04-02
Nop did not pass originating scheme (HTTP/HTTPS) to middlewares from proxy. 


nextfool:
nopCommerce version: all

If we deploy Nop behind a proxy, the HTTPS schema is not passed to those middlewares who may need the info (like the authentication middleware), this will cause the facebook auth plugin to send only HTTP callback url to facebook api.

To fix this, we need to add XForwardedProto header to request pipline.

https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0

```
        public void ConfigureRequestPipeline(IApplicationBuilder application)
        {
            application.UseForwardedHeaders(new ForwardedHeadersOptions
            {
                ForwardedHeaders = ForwardedHeaders.XForwardedProto
            });
            ServiceProvider = application.ApplicationServices;

            //find startup configurations provided by other assemblies
            var typeFinder = Resolve<ITypeFinder>();
            var startupConfigurations = typeFinder.FindClassesOfType<INopStartup>();

            //create and sort instances of startup configurations
            var instances = startupConfigurations
                .Select(startup => (INopStartup)Activator.CreateInstance(startup))
                .OrderBy(startup => startup.Order);

            //configure request pipeline
            foreach (var instance in instances)
                instance.Configure(application);
        }
```

RomanovM:
This should work properly when no proxy is used.
See also discussion here https://www.nopcommerce.com/boards/topic/82570/facebook-login-not-working-in-43

exileDev:
Closed #5584

PR #5585, opened 2021-04-02
forward original scheme (HTTPS)

nextfool:
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0


nextfool:
#5584

exileDev:
Thanks for your suggestion, but we did it a little differently. You can learn the solution in the target issue.

Issue #5608, opened 2021-04-12
Crimea is Ukraine


bo-bac:
nopCommerce version: 4.4, current

Steps to reproduce the problem: 
1 open App_Data/Localization/states.txt
2 as well in DB

Actual:
RU,Крым,82,True,1  
UA,Донецька область,Дон. обл.,False,5
UA,Луганская область,Луганська область,False,12

Expected:
~RU,Крым,82,True,1~  
UA,Автономна Республіка Крим,АР Крим,True,1
UA,Донецька область,Дон. обл.,True,5
UA,Луганская область,Луганська область,True,12

Lsmith998:
@AndreiMaz How can you discuss this!!! 
Crimea is belong to Ukraine without any discussion.

noploser:
It seems that nopcommerce thinking Crimea is Russia since 2019.

badijm:
Crimea is Ukraine.

stopnopcommerce:
We should stop using this software and continue to develop with it!!! 
Such behavior and companies should be condemned by the community.
**Crimea is Russia this is not by mistake!**
Another evidence we can find here: [https://www.nopcommerce.com/en/all-states-provinces-package](https://www.nopcommerce.com/en/all-states-provinces-package)

AndreiMaz:
We're out of politics. So let's make it easy. Now anyone can include it to any country he/she wants - https://github.com/nopSolutions/nopCommerce/commit/7918322d1914e01680078b1eaba61b46ba409b55

stopnopcommerce:
@AndreiMaz You are very clever, Crimea was removed from Russia, but now where should you put it?

bo-bac:
what about
UA,Донецька область,Дон. обл.,**True**,5
UA,Луганская область,Луганська область,**True**,12

skoshelev:
Hi @bo-bac. Of course, I have corrected this information, thanks for your help.

PR #5609, opened 2021-04-12
Crimea is Ukraine

bo-bac:
fix for #5608

it is about
App_Data/Localization/states.txt
but there is the same thing in DB please fix

IMHO we should patch 4.4
@AndreiMaz  what do you think


AndreiMaz:
please see and commit and comment at https://github.com/nopSolutions/nopCommerce/issues/5608#issuecomment-826330001

Issue #5615, opened 2021-04-15
js minification not working in version 4.40.3


vubiz-am:
nopCommerce version: 4.40.3

Steps to reproduce the problem: 
1. In admin section enable javascript bundle and minification
2. Try open checkout page

Error: Javascript not loading

![image](https://user-images.githubusercontent.com/54327114/114916004-4dedab00-9df2-11eb-869b-8c97f249e9ef.png)

AndreiMaz:
@vubiz-am It works fine out of the box, we cannot reproduce it.

Furthermore, it checkout URL looks weird (non standard). Are you using a third-party or custom checkout plugin?

vubiz-am:
I am using default PayPal plugin, checkout looks strange because JavaScript didn’t load to this page. Maybe this relates to sevenspike theme. This custom here. Issue only on 4.40. 4.30 has no problems with compressing JS into one bundle.

AndreiMaz:
@vubiz-am 
1. Which exactly default PayPal plugin? Standard or Smart Buttons?
2. One page checkout or multi-step?

vubiz-am:
Its PayPal standard plugin with one page checkout. Maybe sevenspike plugin some how affecting NOP JS bundle functionality. I'll try enable log file for checking any additional errors.

AndreiMaz:
It works fine out of the box

vubiz-am:
Yes, can be result of using sevenspike plugin. I directly tried to open this minified js file 
https://mydomain.com/home/bundles/-eerxu6pghxocnp8rmfuhzkgprhazj-d0rnagj9qdge.min.js
Network status 404, but page looks blank. Usually I can see default 404 IIS page, not in this case. I think JS minification fail on checkout page. Reason unknown. I will check log ...

vubiz-am:
No errors in log file, only log for request:
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/2 GET https://mydomain.com/home/bundles/-eerxu6pghxocnp8rmfuhzkgprhazj-d0rnagj9qdge.min.js 
File itself blank, 404 error.

Would be great if NOP can provide more info here in case of any errors. I can't go to plugin provider and ask to fix it without any related prove.

vubiz-am:
Its interesting file exist: https://mydomain.com/home/bundles/-eerxu6pghxocnp8rmfuhzkgprhazj-d0rnagj9qdge.js
[...]

PR #5620, opened 2021-04-16
Quick fix for #5615

vubiz-am:
Quick fix for https://github.com/nopSolutions/nopCommerce/issues/5615
Can be useful, if anyone else have similar problem.


AndreiMaz:
It's not somehow related to the described issue (.min prefix). Furthermore, it works fine out of the box

vubiz-am:
Yes it works out of the box. Problem that in most cases its not enough and we need to install plugins with lots of issues. Many issues related to quality of plugin API provided from NOP.

PR #5621, opened 2021-04-17
DBG test for Madeline

mcarlsonlib:
1. Swap price and item desc on home page
2. Minor changes to Nav bar
3. Automatic Approve text


AndreiMaz:
Not sure what this commits means

Issue #5645, opened 2021-04-25
Customer cached by CustomerGuid is not removed from cache when it should be


jackalvrus:
nopCommerce version: 4.40.3

Steps to reproduce the problem: 

When a customer is modified, the customer entity instance that is cached using CustomerGuid is not removed unless the SystemName property is set. This is not correct. It should always be removed.

skoshelev:
Closed #5645

PR #5646, opened 2021-04-25
Issue 5645 customer cache by guid

jackalvrus:
Fix for issue #5645


skoshelev:
Hi @jackalvrus. We fixed the inaccuracy you found with this [commit](https://github.com/nopSolutions/nopCommerce/commit/a35f5f2b570f0ee206941c3c353c4a0a19ddb311). We did not invert the conditional operator, since, in our opinion, such a record loss in readability.

jackalvrus:
@skoshelev Thank you for fixing that.

I'd just like to clarify that the reason I changed the conditional was because I believe that it made the code more readable. I believe that the way it was originally written is the reason that the bug was introduced. The way I wrote it makes it clear that the condition should only apply to specific operations, and it makes the function less sensitive to the order of operations.

PR #5648, opened 2021-04-26
#5647 Old price does not show on product listing page if product has …

### LINKED ISSUE #5647: Old price does not show on product listing page if product has tier prices
nopCommerce version: 
4.3

Create a product with tier prices Ex: 1 $10, 2 $8, etc
Add a discount for that product
Striketrough old price shows on product page but not on listing pages

mateusbellozupko:
To fix #5647 issue where products with tier prices don't have the old price on listing pages.


DmitriyKulagin:
@mateusbellozupko Thanks a lot!

PR #5665, opened 2021-04-30
Fix for adding footer to Child DataTable

purplepiranha:
Recreated pull request as branch wasn't clean before requesting.

This fixes the main issue described in #5284 Admin Tables Footer not rendering on Child Table.

It does not however address the other issue relating to inline editing/deleting, which appears to be more complicated to rectify.


DmitriyKulagin:
@purplepiranha Thanks!

Issue #5664, opened 2021-04-30
NopCommerce v4.40.x not setting correct mime types for picture uploads


purplepiranha:
nopCommerce version: 4.40.0 - 4.40.3

Steps to reproduce the problem: 
When using Azure to store pictures - any uploaded pictures (or generated thumbs) are stored with the content type of application/octet-stream (which is the default).
They should with the correct content type. e.g. image/png.
In version 4.3 this was working.

Having investigated further it seems that the mime type is always passed to the SaveThumbAsync function as string.Empty which means that the default content type is set.

I have corrected and tested locally. A PR will follow.

DmitriyKulagin:
Closed #5664

PR #5666, opened 2021-04-30
Fix MIME types on picture uploads

purplepiranha:
#5664 NopCommerce v4.40.x not setting correct mime types for picture uploads


DmitriyKulagin:
@purplepiranha Thank you for your contribution.

PR #5690, opened 2021-05-17
Demo plugin

danggadang:
Create demo plugin of creating table in database.


AndreiMaz:
Thanks. But we don't plan to add it out of the box.

PR #5723, opened 2021-06-01
add XSS protection config

### LINKED ISSUE #5675: html.Encode in titles and other texts
nopCommerce version: all

1 Install site use any Cyrillic language
2 Set page title for Home page or any
3 Go to that page and View page source
4 There will be something like this
`<title>WD1. SUBARUMIR - &#x437;&#x430;&#x43F;&#x447;&#x430;&#x441;&#x442;&#x438; &#x432; &#x43D;&#x430;&#x43B;&#x438;&#x447;&#x438;&#x438;</title>`

Expected
It seems we need not any additional encoring.

The same in several places over the code as well.

bo-bac:
here is two points to consider
**1**
it is possible to configure Html.Encode so it will allow not only BasicLatin chars to be unescaped
will fix  #5675 if add to appsettings.json 
```
"XssProtectionConfig": {
    "TextEncoderAllowedRanges": [
      "UnicodeRanges.All"
    ]
  }
```
discussed here https://github.com/aspnet/HttpAbstractions/issues/315 
and the same fix here https://github.com/OrchardCMS/OrchardCore/issues/4041 

**2** 
it seems we should use HTML Sanitation for rich HTML Input in such places like this
`@Html.Raw(Model.FullDescription)`
cuz today it possible to save something like <script>alert('qqq')</script> and it will run
this one for example https://github.com/mganss/HtmlSanitizer

### DIFF
--- a/src/Libraries/Nop.Core/Configuration/XssProtectionConfig.cs
+++ b/src/Libraries/Nop.Core/Configuration/XssProtectionConfig.cs
@@ -1,0 +1,38 @@
+﻿using System.Linq;
+using System.Text.Encodings.Web;
+using System.Text.Unicode;
+
+namespace Nop.Core.Configuration
+{
+    /// <summary>
+    /// Represents XSS protection parameters
+    /// </summary>
+    public partial class XssProtectionConfig : IConfig
+    {
+        /// <summary>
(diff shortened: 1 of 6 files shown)


exileDev:
Thanks, but we're closing it now. See linked issue.

Issue #5711, opened 2021-05-25
GetProductAttributeCombinationBySku method shouldn't return combination for deleted product


holydk:
nopCommerce version: 4.40+

Steps to reproduce the problem: 
1. I have two ProductAttributeCombinations with the same SKU. First combination for a deleted product
2. Try to get combination by SKU, you will get combination for deleted product.

We must exclude getting the ProductAttributeCombinations for deleted products. Just like we do it in the _GetProductBySkuAsync_.
https://github.com/nopSolutions/nopCommerce/blob/d2859f8ee907514ba04109b2ce5f1c32d92c6dc2/src/Libraries/Nop.Services/Catalog/ProductService.cs#L1201-L1216

fizmhd:
ProductAttributeCombination does not have any propery Deleted

holydk:
@fizmhd We can check to product was deleted via **inner join**

fizmhd:
Ok, now the description makes it more clear
Thanks for the reply.

patel-hitesh:
`using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;
using Nop.Core;
using Nop.Core.Caching;
using Nop.Core.Domain.Catalog;
using Nop.Data;

namespace Nop.Services.Catalog
{
    /// <summary>
    /// Product attribute service
    /// </summary>
    public partial class ProductAttributeService : IProductAttributeService
    {
        #region Fields

        private readonly IRepository<Product> _productRepository;
        private readonly IRepository<PredefinedProductAttributeValue> _predefinedProductAttributeValueRepository;
        private readonly IRepository<ProductAttribute> _productAttributeRepository;
        private readonly IRepository<ProductAttributeCombination> _productAttributeCombinationRepository;
        private readonly IRepository<ProductAttributeMapping> _productAttributeMappingRepository;
        private readonly IRepository<ProductAttributeValue> _productAttributeValueRepository;
        private readonly IStaticCacheManager _staticCacheManager;

        #endregion

        #region Ctor

        public ProductAttributeService(IRepository<Product> productRepository,
            IRepository<PredefinedProductAttributeValue> predefinedProductAttributeValueRepository,
            IRepository<ProductAttribute> productAttributeRepository,
            IRepository<ProductAttributeCombination> productAttributeCombinationRepository,
            IRepository<ProductAttributeMapping> productAttributeMappingRepository,
[...]

PR #5729, opened 2021-06-06
GetProductAttributeCombinationBySku method shouldn't return combinati…

dsalodki:
…on for deleted product

#5711


holydk (commented):
Hello. In #5711, we want to avoid getting combinations with the same SKU for deleted products. In your solution, it makes no sense to check if the product has been deleted in the _GetAllProductAttributeCombinationsAsync_ method.

PR #5739, opened 2021-06-17
The item "None" added to "Admin/Category" dropdown list.

Navidliverpool:
Item "None" added to "Admin/Category" dropdown list. It has no functionality.


Navidliverpool:
Please forgive me if i make mistake while working with Github.
I don't have much experince on it.

PR #5765, opened 2021-07-15
Increased code coverage to 100 percent for AffiliateService

noleksiv:


src/Libraries/Nop.Services/Affiliates/AffiliateService.cs
```diff
@@ -74,7 +74,7 @@ public virtual async Task<Affiliate> GetAffiliateByFriendlyUrlNameAsync(string f
 
             var query = from a in _affiliateRepository.Table
                         orderby a.Id
-                        where a.FriendlyUrlName == friendlyUrlName
+                        where a.FriendlyUrlName.ToLower() == friendlyUrlName.ToLower()
```
skoshelev, line 77: you should not mix changes in the source code and adding unit tests, this confuses and makes it impossible to determine where and why the source code was changed. Please add another pull request with these changes if you are sure that they need and describe why you think so

src/Tests/Nop.Tests/Nop.Services.Tests/Affiliates/AffiliateServiceTests.cs
```diff
@@ -100,21 +115,48 @@ public async Task CanGetAllAffiliates()
...
+        {
+            var newlyAffiliate = new Affiliate() { AddressId = 1 };
+
+            await _affiliateService.InsertAffiliateAsync(newlyAffiliate);
+
+            newlyAffiliate.AdminComment = _activeAffiliate1.AdminComment;
+            newlyAffiliate.FriendlyUrlName = _activeAffiliate1.FriendlyUrlName;
+
+            await _affiliateService.UpdateAffiliateAsync(newlyAffiliate);
+
```
skoshelev, line 139: This place should contain the code for loading an object from the database, otherwise the check will give nothing, since you assigned new data a couple of lines above

skoshelev:
Hi @s1enny. Thank you for your contribution, please make changes and we'll apply your code.

RomanovM:
Close due to inactivity. 
We'll increase code coverage ourselves.

PR #5775, opened 2021-07-23
Fix Select Shipping Options Params Ajax Call

felipesja:
Remove "$" symbol in query string url to pass "name" to controller "public virtual async Task<IActionResult> SelectShippingOption([FromQuery] string name, [FromQuery] EstimateShippingModel model, IFormCollection form)"

With $, name is getting null in controller parameter and return "No select shipping options" in ajax call.

As a result, the orderTotal is not updated with the selected and applied shipping value.


RomanovM:
Thanks for your help, was fixed here https://github.com/nopSolutions/nopCommerce/commit/842d92bc2f6e888e586c235d3e8c91451f353366.

PR #5777, opened 2021-07-27
Code Formatted & Cleaned, Unused using statements optimized 

nirzaf:


RomanovM:
Thanks for your work, but so many changes are hard to review. 
It's better if we do such cleanup and refactoring ourselves.

PR #5754, opened 2021-06-30
Fixed problem with SkiaSharp in the Docker container 

ricardopartner:

Issue #5563, opened 2021-03-23
SkiaSharp - Images don't resize when using the supplied Dockerfile


purplepiranha:
nopCommerce version: 4.40.1

Steps to reproduce the problem: 
Create a Docker image using the Dockerfile supplied with NopCommerce 4.40 / 4.40.1
An Alpine runtime image is produced.
Images do not resize when uploaded as SkiaSharp requires library ld-linux-x86-64.so.2 which is not present.

See [https://www.nopcommerce.com/en/boards/topic/90532/image-resize-not-working-on-alpine-since-upgrade-to-nopcommerce-44#283404](https://www.nopcommerce.com/en/boards/topic/90532/image-resize-not-working-on-alpine-since-upgrade-to-nopcommerce-44#283404) for solution/workaround.

There may be a better way to do this - I'm surprised that there is nothing in the SkiaSharp documentation that I can find as the latest preview release states that is supports Alpine x64 (I tried the preview before using the workaround and had the same problem).

skoshelev:
Closed #5563

killerwife:
Using commit cabffa399e getting same error as in https://github.com/nopSolutions/nopCommerce/issues/5639

Running a clean configuration with just a theme installed.

https://gist.github.com/killerwife/e20b6285be242704eba26e455e690589 - error from console shown here.

Commit https://github.com/nopSolutions/nopCommerce/commit/12dd4825dfb56bbc64e192cd35d25e3205427646 is in the build (doublechecked to make sure)

Is it possible it is specifically only because of the theme? Is there a nuget package the theme might be missing?

purplepiranha:
I can't see any reason why this doesn't work and having merged the latest release into the solution that I'm currently working on don't experience this problem.

Try looking at the Nop.Services project file and verifying that the following are both present and have the correct version:

```
<PackageReference Include="SkiaSharp" Version="2.80.2" />
<PackageReference Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.80.2" />
```

Also check to ensure that the following are correct on your DockerFile in case an old version has been left during a merge:

```
# create the build instance 
FROM mcr.microsoft.com/dotnet/sdk:5.0 AS build
```

```
# create the runtime instance 
FROM mcr.microsoft.com/dotnet/aspnet:5.0-alpine AS runtime 
```

Note - this will only work if you are using Alpine as your runtime. If you are using a different Linux OS then you may need to install extra packages while creating the image instead.

[...]

PR #5788, opened 2021-08-09
Install additional required packages in the Dockerfile.

damir-bajramovic:
When running the version `4.40.4` Docker container there are issues with `libSkiaSharp` library missing required dependencies. The issues are:
- Images not scaling properly (they are full-sized) in the shopping cart page/checkout
- Website crashing when trying to edit the product

Adding these packages solves the issues. 

We wanted to fix the issue for other users as well since it might be a problem for them too. We saw issues like [this](https://github.com/nopSolutions/nopCommerce/issues/5563). [This](https://github.com/nopSolutions/nopCommerce/pull/5754) also didn't resolve the issue in our case. 

Keep in mind that we are currently using version `4.40.4` with this fix implemented. Given that this is a development branch, and you are continuously developing new features and fixing bugs, this might not be an ideal solution for these problems in the future and the given state of the `develop` branch. We are just trying to help.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,6 +64,8 @@ ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false
 # installs required packages
 RUN apk add libgdiplus --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted
 RUN apk add libc-dev --no-cache
+RUN apk add freetype-dev --no-cache 
+RUN apk add fontconfig-dev --no-cache 
 
 # copy entrypoint script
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @damir-bajramovic. Thank you for your help. As of today, the problem with libSkiaSharp has been resolved under ticket #5563. If you create an image yourself, check that Linux compatible endings are installed in the entrypoint.sh file (\n instead of \r\n). Also, you may use the official docker image from [DockerHub](https://hub.docker.com/r/nopcommerceteam/nopcommerce/tags?page=1&ordering=last_updated)

Issue #5121, opened 2020-10-27
Sendinblue plugin. Use a nuget package


RomanovM:
nopCommerce version: develop

Let's use nuget package instead of dll in the project.

DmitriyKulagin:
Closed #5121

Issue #4912, opened 2020-07-07
Rename "SendinBlue" to "Sendinblue"


AndreiMaz:
That's the correct title.

Please note that Git ignores when you just rename files this way (replace "B" with "b"). So there should be an intermediate commit with some temporary file name

DmitriyKulagin:
See #5121

PR #5873, opened 2021-09-29
Update SendinBlue casing in NopCommerce.sln

manuel3108:
See #4912, #5121

I was just running a build on Bitbucket Pipelines and noticed this small fix. On Bibucket Pipelines the restore failed because the project could not be found


manuel3108:
Nevermind, seems to be the other way around. Don't know what I messed up during the bitbucket setup

PR #5882, opened 2021-10-05
Fix issues with TimeZone

hummer007:
Fix issues with TimeZone when run in docker container


skoshelev:
Hi @hummer007. Could you please describe in more details which problem you talking about?

hummer007:
When I go to Common --> Customer Setting , dropdown "Default store time zone" is empty if run the application in Docker

PR #5958, opened 2021-11-18
Use InvariantCulture to parse double

### LINKED ISSUE #5957: Test failure on decimal comma culture
`CanExportCustomersToXlsx` and `CanExportOrdersToXlsx` fail with exception when locale culture is set to use a decimal comma instead of point. 

```
  System.ArgumentException : Not a legal OleAut date.
  Stack Trace:
     at System.DateTime.DoubleDateToTicks(Double value)
   at System.DateTime.FromOADate(Double d)
   at Nop.Tests.Nop.Services.Tests.ExportImport.ExportManagerTests.PropertiesShouldEqual[T,Tp](T actual, PropertyManager`1 manager, IDictionary`2 replacePairs, String[] filter) in src\Tests\Nop.Tests\Nop.Services.Tests\ExportImport\ExportManagerTests.cs:line 133
```

jhgbrt:
Export tests failed when locale culture set to use decimal comma instead
of decimal point. The ToString() method used InvariantCulture, but then
double.Parse was called with no specific culture.

fixes #5957


DmitriyKulagin:
@jhgbrt Thank you, we did not notice this annoying error on our test bench.

PR #5984, opened 2021-12-04
Added a PermissionRecord for the ability to publish products

malaksedarous:
Added "PublishProducts" PermissionRecord to enable/disable the ability for the role to be able to publish products on creation and editing.


RomanovM:
Decided not to implement this feature, we'll leave it for customization. Thanks.

Issue #5593, opened 2021-04-07
Clear Cache With Distributed Cache


streamzyuk:
nopCommerce version: 4.40.2

Steps to reproduce the problem:
Run Multiple Instance.
Login to Admin Area.
Select Clear Cache.
Check Redis Keys.
Only the key that are set by the instance are cleared.

Distributed cache manager holds a list of keys which are set by the instance running. It uses this list to clear the cache keys from redis. But when reading it does not check the keys  list. So understand that this list is used only for clearing the keys.

It may add the keys to the list when TryGetItemAsync returns true but this will not solve the problem since still only a portion of the keys will be stored.

Is there any purpose for it.

I expect clear cache to clear all keys in redis server or cluster.

There is no meaning of clearing the keys set by instance. Since it may not even set anything if the instance is created at later stage. Other instances may already created the cache keys.

streamzyuk:
This looks like causing other problems.
Changing settings from the admin area. Since the cache keys are not on the instance, settingsService.ClearCache does not clear anything at all.

streamzyuk:
public async Task ClearAsync()
        {
            //we can't use _perRequestCache.Clear(),
            //because HttpContext stores some server data that we should not delete
            var appSettings = Singleton<AppSettings>.Instance;
            var distributedCacheConfig = appSettings.DistributedCacheConfig;
            foreach (var redisKey in _keys)
                _perRequestCache.Remove(redisKey);

            /*
            using var _ = await _locker.LockAsync();

            foreach (var key in _keys) 
                await _distributedCache.RemoveAsync(key);
            */
            try
            {
                var connectionMultiplexer = ConnectionMultiplexer.Connect(distributedCacheConfig.ConnectionString + ",allowAdmin=true");
                var endpoints = connectionMultiplexer.GetEndPoints(false);
                foreach (var endpoint in endpoints)
                {
                    var server = connectionMultiplexer.GetServer(endpoint);

                    if (!server.IsSlave)
                    {
                        var keys = server.Keys(0,"Nop.*");
                        foreach(var key in keys)
                        {
                            await _distributedCache.RemoveAsync(key);
                        }
                       
                       // await server.FlushAllDatabasesAsync();
                    }
                }
                await connectionMultiplexer.CloseAsync(true);
            }
            catch (Exception e) {
                Console.WriteLine(e);
            }
            _keys.Clear();
        }
[...]

PR #6015, opened 2021-12-09
Fix distributed cache key clearing(temporary solution)

JadynWong:
#5593
We can store the cache key in a dedicated cache set, so that multiple instances can read the same cache key set.
This is a temporary method, but it will be more available in a distributed environment than the current method.

Note that querying the cache key set will be very expensive when the number of cache items is very large. Also since there is no distributed locking, some cache keys may be lost when instantaneously storing a large number of caches in different instances.

### DIFF
--- a/src/Libraries/Nop.Core/Caching/DistributedCacheManager.cs
+++ b/src/Libraries/Nop.Core/Caching/DistributedCacheManager.cs
@@ -22,8 +22,8 @@ namespace Nop.Core.Caching
 
         private readonly IDistributedCache _distributedCache;
         private readonly PerRequestCache _perRequestCache;
-        private static readonly List<string> _keys;
         private static readonly AsyncLock _locker;
+        private const string CACHE_KEYS_KEY = "Nop.CacheKeys";
 
         #endregion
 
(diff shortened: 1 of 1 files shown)


Koushiq:
@JadynWong what do you mean by dedicated cache set ?

JadynWong:
> @JadynWong what do you mean by dedicated cache set ?

It may be a description error.

Now all the cache keys in the cache are stored in the static variable `keys`. To solve the current multi-instance problem, we can store `keys` as a cache item. So that each instance can read the same keys.

```c#
  var (isSet, keys) = TryGetItem<HashSet<string>>(new CacheKey(CACHE_KEYS_KEY));
  if (!isSet)
  {
      keys = new HashSet<string>();
  }
  keys.Add(key.Key);
  _distributedCache.SetString(CACHE_KEYS_KEY, JsonConvert.SerializeObject(keys));
```

JadynWong:
Sorry for the very late reply.
I just tested it and could not reproduce your problem. But don't rule out the issue.

> Note that querying the cache keys item will be very expensive when the number of the system cache items is very large. Also since there is no distributed locking, some cache keys may be lost when instantaneously storing a large number of caches in different instances.

I will close this PR, and let's wait for a more suitable solution. 🤔

PR #6017, opened 2021-12-09
Fix problems with googles latest microdata

manuel3108:
We started getting some problems in Google Search Console, as Google apparently changed the way their Microdata should look like. This PR adapts `_Microdata.cshtml` to adapt to the latest changes from Google

See https://developers.google.com/search/docs/advanced/structured-data/product?hl=de#microdata for more details


DmitriyKulagin:
@manuel3108 Thanks a lot!

PR #5972, opened 2021-11-25
Issue 5962: Refactor BBCodeHelper, NopHtmlHelper, PagerModel

Zalhera:
Standardized naming of fields and properties. Removed extra spaces, indents, etc.

PR #6019, opened 2021-12-09
reformat, rename fields, remove partial from single part classes

Aayenah:
This is my very first contribution to open source, so I apologize if I did something wrong. Please let me know.

Changes:
1. Renamed private fields with a _ prefix.
2. Ran auto formatter which I believe respects .editorconfig
3. Removed redundant "partial" from single-part classes/records. If this is an problem, happy to revert.
4. I wanted to rename the properties too but some of them have many usages outside of the 3 files mentioned in the issue. Happy to go ahead with the rename if you like.
5. Removed unused imports.
6. Shortened some if statements in some getters.


RomanovM:
Thanks, we merged #5972.

Issue #6033, opened 2021-12-14
Reset password does not work when an email contains a +


justinjstark:
nopCommerce version: 
Latest

Steps to reproduce the problem: 
- Create an account with an email like "test+test@gmail.com"
- Send the password reset link
- The URL in the email will contain "email=test+test@gmail.com" which model binds in MVC to email="test test@gmail.com".
- Clicking the reset link sends you to the homepage and does not allow you to reset your password

The URL parameters should be escaped as "email=test%2Btest@gmail.com". In a URL, the + acts as a space.

It appears this is caused in MessageTokenProvider.cs.
![image](https://user-images.githubusercontent.com/214307/146073863-d955273f-67b0-4019-b362-0557a4609845.png)

`urlHelper.RouteUrl(routeName, routeValues))` results in a properly escaped URL. But when you pass this into the PathString constructor, it escapes everything. You then UrlDecode it later resulting in an unecoded + in the email value. Microsoft's documentation specifically says not to use PathString with an encoded URL.

I don't have time to issue a pull request now but we fixed it by doing this:
![image](https://user-images.githubusercontent.com/214307/146074406-95ef1780-b03d-48df-9644-91511893969e.png)

RomanovM:
See also #3946

DmitriyKulagin:
@justinjstark Since version 4.30, [we pass the user's Guid](https://github.com/nopSolutions/nopCommerce/blob/d9c79c6b61ad9c5cdfc544f2b8bec16c3ceb13f3/src/Libraries/Nop.Services/Messages/MessageTokenProvider.cs#L1218-L1218) to the RouteUrlAsync method. How do you pass the Email parameter to the method?

justinjstark:
@DmitriyKulagin The user GUID is just a RouteValue? It gets appended and escaped correctly by UrlHelper.RouteUrl just like all other RouteValues. What am I missing?

DmitriyKulagin:
Closed #6033

PR #6036, opened 2021-12-15
Issues 3946 & 6033 - Fix URL encoding in RouteUrlAsync

### LINKED ISSUE #3946: Registration activation with "+" in the email addres will not be successfull
nopCommerce version: 4.20

If the activation link contains the email in an unescaped format like this (user+name@gmail.com) then the plus sign in the URL is represents a space, so the activation will not be successful for the email "user name@gmail.com".

Source: Premium Support #14907

justinjstark:
Fixes #3946 and #6033.

I don't know why this method is doing what it is doing nor what impact this change will have on the consuming services (ex: do any services expect the badly encoded URL that was previously provided). Feedback is appreciated. This fixes the email reset / activation issue but what other test cases need run?

### DIFF
--- a/src/Libraries/Nop.Services/Messages/MessageTokenProvider.cs
+++ b/src/Libraries/Nop.Services/Messages/MessageTokenProvider.cs
@@ -902,16 +902,12 @@ namespace Nop.Services.Messages
             if (string.IsNullOrEmpty(store.Url))
                 throw new Exception("URL cannot be null");
 
-            //generate a URL with an absolute path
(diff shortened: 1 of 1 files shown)


justinjstark:
Also we have customers experiencing this issue so it'd be good to get it fixed sooner than later. I'm a little surprised this issue has existed since June 2019 considering most modern email clients now support "+" aliasing and email addresses can contain other characters that were not being escaped correctly.

PR #6055, opened 2021-12-27
Fix store mapping authorization check on multi-store setups

srobati:
In a multi-store/single-store setup, we must rely on `storeId` param not `_storeContext.GetCurrentStoreAsync()`.


skoshelev:
Hi, @srobati. Thank you for your help

PR #6082, opened 2022-01-15
Resolves #6081 - Refactor to use ViewComponent's type instead of string when invoking it

### LINKED ISSUE #6081: Code Refactoring: Use ViewComponent's type instead of name for strong typing
nopCommerce version: 4.5-rc

Instead of
```
[ViewComponent(Name = "CheckMoneyOrder")]
public class CheckMoneyOrderViewComponent : NopViewComponent
{
}

public class CheckMoneyOrderPaymentProcessor : BasePlugin, IPaymentMethod
{
    public string GetPublicViewComponentName()
    {
        return "CheckMoneyOrder";
    }
}
```

Do this instead
```
public class CheckMoneyOrderViewComponent : NopViewComponent
{
}

public class CheckMoneyOrderPaymentProcessor : BasePlugin, IPaymentMethod
{
    public Type GetPublicViewComponent()
    {
        return typeof(CheckMoneyOrderViewComponent);
    }
}
```

This extends to the entire codebase utilizing ViewComponents including Payment plugins, Shipping plugins, External Login plugins, etc. By using a stronger typing approach, the usages of ViewComponents can tracked across the solution.

According to https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-6.0, 
image

DamienLaw:


DmitriyKulagin:
Hi, @DamienLaw 
Thanks for your input, even though we decided to implement this task ourselves, your code was a good starting point for this. Thanks!

Issue #6092, opened 2022-01-23
Add an additional property to `AdditionalTokensAddedEvent` to be used to filter by TokenGroups


DamienLaw:
nopCommerce version:  4.50-rc

In order to make the Tokens allowed in Message Templates more extensible, we could add a new property `TokenGroups` to [`AdditionalTokensAddedEvent`](https://github.com/nopSolutions/nopCommerce/blob/12dd4825dfb56bbc64e192cd35d25e3205427646/src/Libraries/Nop.Core/Domain/Messages/AdditionalTokensAddedEvent.cs)

```
public class AdditionalTokensAddedEvent
{
    public IList<string> AdditionalTokens { get; }

    /// <summary>
    /// Token groups which can be used to filter the AdditionalTokens
    /// </summary>
    public IEnumerable<string> TokenGroups { get; set; } //<-- Add this line
}
```

In the `GetListOfAllowedTokensAsync()` method of [`MessageTokenProvider`](https://github.com/nopSolutions/nopCommerce/blob/cf1cdbe9ccc7a5dfa44f3617e6ed92583e353f84/src/Libraries/Nop.Services/Messages/MessageTokenProvider.cs#L1507), we could supply the `TokenGroups` to `AdditionalTokensAddedEvent` right before publishing it.

```
public virtual async Task<IEnumerable<string>> GetListOfAllowedTokensAsync(IEnumerable<string> tokenGroups = null)
{
    var additionalTokens = new AdditionalTokensAddedEvent
    {
        TokenGroups = tokenGroups // <-- Add this line
    };
    await _eventPublisher.PublishAsync(additionalTokens);

    var allowedTokens = AllowedTokens.Where(x => tokenGroups == null || tokenGroups.Contains(x.Key))
        .SelectMany(x => x.Value).ToList();

    allowedTokens.AddRange(additionalTokens.AdditionalTokens);

    return allowedTokens.Distinct();
}
```

Now in the our Consumer class, we can use the `TokenGroups` to make a decision of whether to add a certain Tokens.
```
public class AdditionalTokensAddedConsumer : IConsumer<AdditionalTokensAddedEvent>
{
    public Task HandleEventAsync(AdditionalTokensAddedEvent eventMessage)
    {
        if (eventMessage.TokenGroups == null || eventMessage.TokenGroups.Contains(TokenGroupNames.OrderTokens))
        {
[...]

PR #6093, opened 2022-01-23
#6092 Allow AdditionalTokens used in Message Templates to be filtered…

DamienLaw:
… by TokenGroups.


DmitriyKulagin:
#6093

DmitriyKulagin:
@DamienLaw Thanks for the suggestion.

Issue #6089, opened 2022-01-20
Posted data to Order Details in Admin backoffice is not redirected


DamienLaw:
nopCommerce version: 4.50-rc

Steps to reproduce the problem: 

1. Place an order in store front.
2. Go to admin backoffice and view/edit the order.
3. Mark the order as Paid.
4. Refresh the page.

![image](https://user-images.githubusercontent.com/14930009/150398194-399cd79a-c5d9-4c87-9ab9-f09a47bdc88f.png)

In order to prevent resubmission of the same action, redirect back to the same page instead. This is known as the design pattern POST-REDIRECT-GET (PRG):

1. https://andrewlock.net/post-redirect-get-using-tempdata-in-asp-net-core/
2. https://en.wikipedia.org/wiki/Post/Redirect/Get

Taking an excerpt from the above.
![image](https://user-images.githubusercontent.com/14930009/150447793-d9722a6f-fe84-4722-9048-2c7487a50d2c.png)

The following actions are affected:

1. CancelOrder
2. CaptureOrder
3. MarkOrderAsPaid
4. RefundOrder
5. RefundOrderOffline
6. VoidOrder
7. VoidOrderOffline
8. ChangeOrderStatus
9. EditCreditCardInfo
10. EditOrderTotals
11. EditShippingMethod
12. EditOrderItem
13. ResetDownloadCount
14. ActivateDownloadItem

AndreiMaz:
Thanks a lot for the suggestion. But anyway it won't cause any issues because we validate whether an order has already been paid, cancelled, etc.

DamienLaw:
Thanks for taking the time to read this issue. Yes, I know the requests are idempotent because I'm able to read the source codes. However, I couldn't say the same to the clients that I deploy nopCommerce for. It is them that use nopCommerce the most. When they approached me with this question of whether it's safe or not to resubmit, I couldn't tell them until after I spent some time reading the source.

PRG is a widely adopted design pattern that has been around for quite some time which is not so much to safeguard against non-idempotent requests, but rather to prevent the "Confirm Form Resubmission" from appearing and making users guessing and wonder. The only time that this is acceptable is, when a form validation failed and required user's attention. Successful POST requests should be redirected to prevent confusion.

DamienLaw:
Actually there are a lot of POST controller actions already using PRG. For example,
[...]

PR #6094, opened 2022-01-23
#6089 Fixed some controller actions that are not consistent with the …

DamienLaw:
…rest of the controller actions employing POST-REDIRECT-GET (PRG) design pattern.


RomanovM:
@DamienLaw Thanks for your help.

Issue #6099, opened 2022-01-26
PayPal Standard to support additional fees in both forms of fixed fee and fee by percentage


DamienLaw:
nopCommerce version: 4.5

PayPal is charging merchants 3.9% of Order Total + MYR 2 for each transaction. Naturally, merchants would want to push this expense to the customers.

The current PayPal Standard plugin is only able to handle additional fee of either a fixed amount or fee by a certain percentage, but not both. Here's a suggestion to support both.

RomanovM:
Thanks for the suggestion, but we don't plan to implement it, we leave it for customization.
Moreover, when we developed PayPal Commerce integration in cooperation directly with PayPal team, they asked to remove any additional fees from the plugin, this was one of their requirement. So, perhaps we remove it from PayPal Standard plugin as well.

PR #6100, opened 2022-01-26
#6099 PayPal Standard to support additional fees in both forms of fix…

DamienLaw:
…ed amount and fee by percentage.


RomanovM:
Thanks. Leave this for customization.

Issue #5947, opened 2021-11-13
Search customers by "Registration" (from and to) and "Last activity" dates 


AndreiMaz:
Let's add two new date filters on the "Customers" page in admin area. Filter customers by `LastActivityDateUtc` property (from and to). The field names will be "Last activity from" and "Last activity to".
Originally suggested in https://www.nopcommerce.com/boards/topic/91632/search-for-active-customers-only

Also, add new other fields to filter customers - "Registration date from" and "Registration date to". They will use the `CreatedOnUtc` property

PR #6119, opened 2022-02-10
#5947 - 8 search customers by last activity and registration from and to dates

mhsjaber:


RomanovM (changes requested):
We need to exclude guests from the result since the filter is "by registration date", and the guests have no registration, but they also have `CreatedOnUtc`. So, let's do it right in the factory.

Issue #5702, opened 2021-05-19
Don't store a currency for the Crawler accounts


RomanovM:
nopCommerce version: 4.40

We should not store currency for Crawler customers (search engines). Always use the default currency for them

Many topics on our forums, e.g https://www.nopcommerce.com/boards/topic/46884/prevent-change-currency-from-google-crawlers, https://www.nopcommerce.com/boards/topic/90983/microdata-currency, https://www.nopcommerce.com/boards/topic/90841/googlebot-picking-up-incorrect-currency, etc.

Good explanation of the problem and suggestion to fix it here https://www.nopcommerce.com/boards/topic/90841/googlebot-picking-up-incorrect-currency#284671

PR #6122, opened 2022-02-10
#5702 Currency store restricted for the system customer

mhsjaber:


RomanovM (approved):

RomanovM:
@mhsjaber  There is a problem with the file line endings, so I merge changes directly to the develop.

Issue #29, opened 2016-01-14
Display "From" prices for products with attributes or combinations with price adjustments (configurable)


mariannk:
We have product boxes on catalog pages (category, manufacturer, etc). In some cases, we have product attributes with price adjustments. In some cases, we have attribute combinations with price adjustments.

So actually we should display the minimum possible price of a product. So if a product could have different prices (based on selected attributes), then we should display "From ..." price instead of the fixed base price (such as it's done now at https://demo.nopcommerce.com/desktops - just an example).

Also, let's make it configurable whether to display the base price or display the minimum possible price. Otherwise, it can affect the performance in some cases. Add this new setting to admin area > configuration > catalog pages > "Catalog pages" block > "Display "From" prices" setting (disabled by default)

AndreiMaz:
I was going to bring up something on the 3.4 roadmap, but because it was already looking awesome, I was going to wait until 3.5 before bringing this up. But seeing the statement "make it configurable because it can seriosly affect performance (disable by default) ", I will mention it now.

If we took a different architecture approach, many of these expensive lookups could be minimized. If we had an additional background job that ran off a queue, we could do very complex logic like this work item is requesting with no performance impact.

Just like cache is updated when we save a product in the admin, any changes could also be pushed to a stack (similar to what we are doing already with activity monitor). So, lets look at a practical scenario.

Store admin comes in and makes price adjustments to attribute options and maybe even updates a lot of attribute combinations. We write (using a queue paradigm) the events to a background work stack. So, an example row would be (ProductId, enum for price adjustment, etc). Within a min or two, the background job would pull off the stack and determine (From and To) prices based on changes, then write these 2 values right into 2 new columns in the product table and updates the cache as normal.

When this product is pulled for the Catalog for someone browsing, the From and To are now simple fields that have no "real-time" calculation needed.

This same architecture shift could be applied to many many scenarios in nopCommerce. The goal being, during user browsing of the site, avoid complex logic by just putting results as simple attributes, and leaving the calculation of those values to back-end services. 

As this work item is on the 3.4 roadmap now, I will make a note in the roadmap thread.

AndreiMaz:
One additional point to the above comment. Essentially this adds up to introducing the concept of "Calculated fields" populated by background jobs to make catalog browsing faster.

AndreiMaz:
[...]

PR #6126, opened 2022-02-13
#29 - 10 catalog pages display from prices for products with attributes with price adjustments configurable

mhsjaber:


RomanovM:
@mhsjaber There is a problem with the file line endings, so I merged changes directly to the develop. See this commit https://github.com/nopSolutions/nopCommerce/commit/514d783d778ad680829542830ea369e9139a1725

Issue #385, opened 2016-01-18
Add support for product URLs with a category in path (configurable)


mariannk:
This way, product URLs could be `/category-slug-name/product-slug-name`. If a product has several categories in the breadcrumb, then use only the last one (the most nested). Also, note that sometimes a product could be with any categories (in this case always use `/product-slug-name`)

Let's make it configurable - the current way (e.g. `/product-slug-name`) or a new one (`/category-slug-name/product-slug-name`).  Add an appropriate setting to the admin area

Some notes (to be investigated):
1. The workaround for that is for `GenericPathRoute` to map to `{*generic_name}` rather than {generic_name} I believe.
2. Please note after addition of such custom route ALL controllers and routes should be also registered. For example, \somecontroller\somemethod. Otherwise, customers will be redirected to this new {*generic_name} route handler. Also, think about the processing of 404 errors for such cases. Some investigation is required.
3. Each missing static resource (like /content/files/missingfile.pdf) is also caught by `GenericPathRoute`, it's not a big problem, but I'm not sure of the consequences. 

Also see - http://www.nopcommerce.com/boards/t/29637/performance-optimization-needed-nopcommerce-is-too-slow-.aspx?p=7#123632

rudgr:
vote++;

MikeDub:
I initially thought I wanted a /category/product route too.
However I remember that NopCommerce supports multiple categories, 
so would /Brand/Product be a more reliable suggestion?

Maybe ideally if we had a route such as this: /Brand/Category?/Product, where it would only display the category if it was the only category the product was in.

Thoughts?

PR #6137, opened 2022-02-19
#385 product url with category breadcrumb added

mhsjaber:


joekrauss:
Hello Friend, I’ve waited years to see this feature get added to nopCommerce . Looking over the code I see a possible issue for some store fronts. A quick background several SEO experts explain the need for this feature as it gives “SEO value” to products hierarchy. This comes from the category to product relationships. In our case as with many other store fronts there is no definition of default category.
We use weekly special categories today for showcasing our weekly promotions. Some deals categories are restricted to certain roles. Given we have products listed in multiple categories which category should be listed in the breadcrumb? I believe there should be a default category assignment to link products to specific categories for your logic and SEO benefits.

Thx

RomanovM:
Thanks, but we decided to implement it ourselves.

PR #6174, opened 2022-03-13
Refactored intent of variable in Setting Controller.

coderudit:
Refactored storeScope variable to activeStoreScopeConfiguration to clear it contains configuration.


RomanovM:
Thanks for your suggestion, but it's already clear, moreover we use such var name in a whole controller.

PR #6175, opened 2022-03-13
Refactor localization migration up method into multiple methods.

coderudit:
Extracted add, delete, rename locales method from Up method in LocalizationMigration.cs.


RomanovM:
@coderudit What is the purpose of such refactoring?

coderudit:
> @coderudit What is the purpose of such refactoring?

I have picked up this repository to target clean code.

Issue #5874, opened 2021-09-30
Facebook Pixel plugin. Use Facebook Conversions API


li5908:
The Facebook Pixel plugin (available out of the box). Let's use Facebook Conversions API

Also see here https://www.nopcommerce.com/boards/topic/92576/api-conversions-facebook-plugin

RomanovM:
See also here https://www.nopcommerce.com/boards/topic/92576/api-conversions-facebook-plugin

PR #6190, opened 2022-03-18
#5874 Facebook pixel conversions api functionality added

mhsjaber:


src/Plugins/Nop.Plugin.Widgets.FacebookPixel/Services/FacebookPixelService.cs
```diff
@@ -801,6 +904,391 @@ public async Task PrepareCompleteRegistrationScriptAsync()
...
+        public async Task SendPurchaseEventAsync(Order order)
+        {
+            await HandleFunctionAsync(async () =>
+            {
+                //check whether the purchase was initiated by the customer
+                var customer = await _workContext.GetCurrentCustomerAsync();
+                if (order.CustomerId != customer.Id)
+                    return false;
+
+                //prepare event object
```
RomanovM, line 995: Now preparing of event objects (for all events) is duplicated for the script and for the conversions API, I suggest to combine them, perhaps even using common models for this.

RomanovM (approved):

PR #6193, opened 2022-03-20
Refactored authroize method to new IAuthorizationService from IPermissionsService.

coderudit:
Authorize method is present in IPermissionsService which is giving permission to users. Authorize should have it's own class. Therefore a new interface IAuthorizationService and class AuthorizationService is extracted with Authorize method moved.
Extract class: IAuthorizationService, AuthorizationService
Move method: Authorize


RomanovM:
Thanks for the suggestion, but we leave the current approach.

Issue #6153, opened 2022-03-02
Move customer fields from GenericAttribute table to Customer table. Prepare the upgrade scripts.


RomanovM:
nopCommerce version: develop

In continuation of #4601.
We should prepare the SQL scripts (for each supported db provider) for the upgrade. It will be needed for users who will update db manually, without migration (for example, if there are many records in the `Customer` and `GenericAttribute` tables).

skoshelev:
Closed #6153

PR #6195, opened 2022-03-21
#6153 upgrade script added to move customer fields from GenericAttribute table to Customer table

mhsjaber:


skoshelev:
Hi @mhsjaber. Thank you for your help

Issue #5989, opened 2021-12-08
Use RenderPartial method on views 


RomanovM:
nopCommerce version: develop

Let's consider using of `RenderPartialAsync` methods on all views instead of `PartialAsync`.

From [MS docs](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-6.0):

> Since RenderPartialAsync streams rendered content, it provides better performance in some scenarios. In performance-critical situations, benchmark the page using both approaches and use the approach that generates a faster response.

exileDev:
Closed #5989

Details #6204

PR #6204, opened 2022-03-28
Replace PartialAsync with RenderPartialAsync

Zalhera:
Replace all appropriate calls to PartialAsync with RenderPartialAsync.


Zalhera:
Solves issue #5989

glhays:
Curious if this was considered in this proposal from issue #5989.
"benchmark the page using both approaches and use the approach that generates a faster response."


________________________________

exileDev:
I've made some tests to compare these changes with the develop branch. To summarize: there are no performance improvements, both in public and admin area.

Thank you for your work, @Zalhera

PR #6208, opened 2022-04-05
Ignored cancelled orders in discount coupon with N-Time use

fuadhasan28:


src/Libraries/Nop.Services/Discounts/IDiscountService.cs
```diff
@@ -209,7 +209,7 @@ public partial interface IDiscountService
         /// </returns>
         Task<IPagedList<DiscountUsageHistory>> GetAllDiscountUsageHistoryAsync(int? discountId = null,
             int? customerId = null, int? orderId = null,
-            int pageIndex = 0, int pageSize = int.MaxValue);
+            int pageIndex = 0, int pageSize = int.MaxValue, bool includeCancelledOrders = true);
```
RomanovM, line 212: The preferable way is to add new parameters before pagination ones. And don't forget to update summary comments of methods.

RomanovM (changes requested):

Issue #6182, opened 2022-03-15
Add reCAPTCHA to guest checkout 


AndreiMaz:
Let's add a setting indicating whether reCAPTCHA should be displayed during guest checkout (disabled by default). Let's display it on the last "Confirm order" step (if enabled).

Source:
1. https://www.nopcommerce.com/en/boards/topic/93742/recaptcha-one-page-checkout-how-do-i-do-it 
2. https://www.nopcommerce.com/en/boards/topic/46776/recaptcha
3. https://www.nopcommerce.com/en/boards/topic/35658/any-tips-for-implementing-no-captcha-recaptcha-in-nop#182428

ChrisClaude:
Is anyone working on this? I would love to pick it up.

AndreiMaz:
@ChrisClaude Thanks a lot for interest to contribute. But it's almost done - https://github.com/nopSolutions/nopCommerce/pull/6223

PR #6215, opened 2022-04-06
Captcha for guest checkout confirm

Zalhera:
Solves #6182


Zalhera:
Still working on the correct Captcha Validation on the Opc Checkout and adding the Captcha to the normal Checkout.

RomanovM:
Thanks for your contribution, it was partially used in another pr (#6223) for this issue.

PR #6216, opened 2022-04-07
GetStoreHost behind reverse proxy

ekjuanrejon:
When running NopCommerce behind a reverse proxy the image link and javascript links are not generated with https. this results in browser mixed content. 
In checkout page ajax calls fails to execute because of https/http mismatch. This will check if the request contains x-forwarded-proto header


ekjuanrejon:
May could have injected and check HostingConfig.UseProxy. In any case load balancers will forward those headers. 
`
 EngineContext.Current.Resolve<AppSettings>();
`

Issue #5969, opened 2021-11-24
Product back in stock notifications not triggered on import


skoshelev:
We should be updated the ImportManager.ImportProductsFromXlsxAsync method To call the notification service as it does in the Product controller.

Details are described on premium support ticket 16342

PR #6227, opened 2022-04-12
Back In Stock Notification on Excel Import

Zalhera:
Solves issue #5969


RomanovM:
@Zalhera Thanks for your help.

PR #6232, opened 2022-04-17
Add NopHostedService

ekjuanrejon:
- move plugin installation to hosted service
- move default permission to hosted service

- When a new container is created(scaling) at least the plugin is missing. This will install the plugin


ekjuanrejon:
Any reason why pr is not accepted?

RomanovM:
@ekjuanrejon Well, based on your previous PRs (e.g. #6216, #6222), which you immediately closed, I supposed that you just customize your solution in this way.

So, what exactly is the point of using a hosted service for installation?

ekjuanrejon:
When a nop docker image is built the plugin are inside the image but have to be installed. 

1. The plugin are first installed by the user through the InstallController
2. If you start with one container and then try to scale to 4 container, the new containers don't have the plugin installed.
3. This will always install the plugin inside the container

I can re-open the other prs. In my opinion the application should work seamlessly even if it is deployed behind a loadbalancer

RomanovM:
Thanks for your suggestion, we leave it for customization.

Issue #3651, opened 2019-04-01
Use single generic event for order status changes


ahmadkq:
nopCommerce version: 4.1

Steps to reproduce the problem: 
Only OrderCancelledEvent is available, let us add Pending, Processing & Completed events.

Only when order is canceled or completed, email notification available, let us add when order is processing.

astrumcs:
This idea has merit.  I believe that it would be a good idea to add a general event when there is any order status change.   Pass the old and new order status as an argument.

PR #6239, opened 2022-04-20
#3651 Added single generic event for order status changes

mhsjaber:


src/Libraries/Nop.Core/Domain/Orders/OrderStatusChangedEvent.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// Order status changed event
+    /// </summary>
+    public class OrderStatusChangedEvent
+    {
+        /// <summary>
+        /// Ctor
+        /// </summary>
+        /// <param name="order">Order</param>
+        public OrderStatusChangedEvent(Order order)
```
RomanovM, line 12: No need for this ctor.

src/Libraries/Nop.Core/Domain/Orders/OrderStatusChangedEvent.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Order
+        /// </summary>
+        public Order Order { get; }
+
+        /// <summary>
+        /// Previous order status
+        /// </summary>
+        public OrderStatus PrevOrderStatus { get; set; }
```
RomanovM, line 36: Let's name this `PreviousOrderStatus`

src/Libraries/Nop.Services/Orders/IOrderProcessingService.cs
```diff
@@ -18,7 +18,7 @@ public partial interface IOrderProcessingService
         /// </summary>
         /// <param name="order">Order</param>
         /// <returns>A task that represents the asynchronous operation</returns>
-        Task CheckOrderStatusAsync(Order order);
+        Task CheckOrderStatusAsync(Order order, bool notifyCustomer);
```
RomanovM, line 21: Not sure if this param is needed. In what cases can it be `false`?

src/Libraries/Nop.Services/Orders/OrderProcessingService.cs
```diff
@@ -2405,7 +2423,7 @@ public virtual async Task CancelOrderAsync(Order order, bool notifyCustomer)
             //Adjust inventory
             await ReturnOrderStockAsync(order, string.Format(await _localizationService.GetResourceAsync("Admin.StockQuantityHistory.Messages.CancelOrder"), order.Id));
 
-            await _eventPublisher.PublishAsync(new OrderCancelledEvent(order));
+            await _eventPublisher.PublishAsync(new OrderStatusChangedEvent(order, prevOrderStatus));
```
RomanovM, line 2426: This new event raises only when the order is canceled. What about all the other status changes?

RomanovM (changes requested):
I suggest splitting this pr into two, one for the new message template and related setting, and the other for the new event.

RomanovM:
@mhsjaber Merged directly in develop in [this commit](https://github.com/nopSolutions/nopCommerce/commit/2534fdb12dcf0d7565269ba46f7016a8b7a1eea5).

Issue #6188, opened 2022-03-17
Bump EasyPost-Official Package (Dependency)


Justintime50:
nopCommerce version: 
NA

Steps to reproduce the problem: 
The version of `EasyPost-Official` being used in this repo was unlisted on Nuget and should be bumped to the latest (v2.8.2 at the time of this writing) due to an incorrectly packaged version at v2.7.0 which is in use in this project. It should be as easy as a version swap as no breaking changes were introduced.

Justintime50:
We just released `v3.0.0` with various improvements; however, it's a large breaking change that would require a bit of refactoring of the EasyPost integration here.

PR #6240, opened 2022-04-21
Update EasyPost to 3.0

Zalhera:
Solves #6188

### DIFF
--- a/src/Plugins/Nop.Plugin.Shipping.EasyPost/Services/EasyPostService.cs
+++ b/src/Plugins/Nop.Plugin.Shipping.EasyPost/Services/EasyPostService.cs
@@ -417,12 +417,16 @@ namespace Nop.Plugin.Shipping.EasyPost.Services
                 options.currency = EasyPostDefaults.CurrencyCode;
 
             //create shipment
-            var shipment = await Shipment.Create();
-            shipment.to_address = addressTo;
-            shipment.from_address = addressFrom;
-            shipment.parcel = parcel;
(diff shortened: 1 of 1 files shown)


src/Plugins/Nop.Plugin.Shipping.EasyPost/Services/EasyPostService.cs
```diff
@@ -417,17 +417,15 @@ private async Task<Address> PrepareAddressFromAsync(GetShippingOptionRequest req
...
             //create shipment
-            var shipment = new Shipment
-            {
-                to_address = addressTo,
-                from_address = addressFrom,
-                parcel = parcel,
-                customs_info = customsInfo,
-                options = options
-            };
+            var shipment = await Shipment.Create();
```
nwithan8, line 420: This will not work in v3.0.0 of the EasyPost C# library. You need to pass the attributes in a dictionary as a parameter to the `Create` function.

Please refer to the Upgrade Guide here for more details: https://github.com/EasyPost/easypost-csharp/blob/master/UPGRADE_GUIDE.md#30-removes-non-static-create-functions

Zalhera: @nwithan8 Thank you for the hint, the shipment creation is now updated.

nwithan8:
Looks good to me! Thanks for updating!

RomanovM:
Thanks for your contribution.

PR #6264, opened 2022-05-25
Fix container path for app data

mitch-b:
When building the web application container, ensure it uses the proper path from documentation so that mounted paths work as expected.

See [forum post](https://www.nopcommerce.com/en/boards/topic/94314/docker-upgrade-installation-page#296108) by **giovannidebona**.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,8 @@ RUN apk add tzdata --no-cache
 COPY ./entrypoint.sh /entrypoint.sh
 RUN chmod 755 /entrypoint.sh
 
+WORKDIR /app
+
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @mitch-b. Thank you for your help. we'll update the docker image soon

D3bo:
I updated with the new tag 4.50.2 that you just released, but seems that the version number is wrong:

image

skoshelev:
hi @d3bo. I checked twice, all versions are set to 4.50.2

mitch-b:
Thanks @D3bo and @skoshelev - the `4.50.2` tag is working as expected today.

Issue #6266, opened 2022-05-25
issue - Map LocalizedProperty with foreign Id


SpeakOutLoud:
nopCommerce version: 4.60

Hi, this is my first issue ticket :).

What do you think about the `LocalizedProperty` table to create foreign id from 2 columns:
LocalKey and LocalKeyGroup.

Why I am mention this, if a client wants to host many products (tens of thousand) and would like to add multiple languages for those products, than the LocalKey and LocalKeyGroup as string would repeatedly be stored on the table and would indicate a performance issue for reading the data.

As a solution, if approved of course. I would suggest creating 2 other tables, that would host the LocalKeyGroup and LocalKey with a parented to LocalKeyGroup. Then from LocalizedProperty table we would just map the foriegn id from those 2 created tables, in this case we would reduce the size of the table in terms of space, and would improve the performance, instead of reading the same string multiple times, now it would read the foreign Id as int.

Thank You!

AndreiMaz:
Thanks a lot for your contribution. But we think that the current implementation works absolutely fine in 99.99% cases.

If someone experience any performance issues, we would recommend to check your pull request. Perhaps, he/she will manually use or merge it

PR #6267, opened 2022-05-25
Localization performance #6266

SpeakOutLoud:
Hi, as for my first PR here, I did this for fun so I can get a bit involved in the community :).
I have changed a bit of structure to improve the performance of LocalizedProperty.
Hope anyone likes it or even dislikes.

The issue ticket is #6266 

Thanks


AndreiMaz:
Thanks a lot for your contribution. But we think that the current implementation works absolutely fine in 99.99% cases.

If someone experience any performance issues, we would recommend to check your pull request. Perhaps, he/she will manually use or merge it

Issue #5928, opened 2021-11-01
Allow selecting default product sorting (currently by position only)


brakim:
nopCommerce version:  4.40

The setting **AllowProductSorting**  has unexpected behavior.

When the setting is disabled, the catalog does not honor the sorting options displayorder or active/inactive status, it just default to **Position**.

So if i want the default sorting to be A-Z i must allow the customer to change the order as well.

Original

```
public virtual async Task PrepareSortingOptionsAsync(CatalogProductsModel model, CatalogProductsCommand command)
        {
            //set the order by position by default
            model.OrderBy = command.OrderBy;
            command.OrderBy = (int)ProductSortingEnum.Position;

            //ensure that product sorting is enabled
            if (!_catalogSettings.AllowProductSorting)
                return;

            //get active sorting options
            var activeSortingOptionsIds = Enum.GetValues(typeof(ProductSortingEnum)).Cast<int>()
                .Except(_catalogSettings.ProductSortingEnumDisabled).ToList();
            if (!activeSortingOptionsIds.Any())
                return;

            //order sorting options
            var orderedActiveSortingOptions = activeSortingOptionsIds
                .Select(id => new { Id = id, Order = _catalogSettings.ProductSortingEnumDisplayOrder.TryGetValue(id, out var order) ? order : id })
                .OrderBy(option => option.Order).ToList();

            model.AllowProductSorting = true;
            command.OrderBy = model.OrderBy ?? orderedActiveSortingOptions.FirstOrDefault().Id;

            //prepare available model sorting options
            foreach (var option in orderedActiveSortingOptions)
            {
                model.AvailableSortOptions.Add(new SelectListItem
                {
                    Text = await _localizationService.GetLocalizedEnumAsync((ProductSortingEnum)option.Id),
                    Value = option.Id.ToString(),
                    Selected = option.Id == command.OrderBy
                });
            }
        }
```

Should maybe look like this

```
public virtual async Task PrepareSortingOptionsAsync(CatalogProductsModel model, CatalogProductsCommand command)
        {
            //get active sorting options
            var activeSortingOptionsIds = Enum.GetValues(typeof(ProductSortingEnum)).Cast<int>()
                .Except(_catalogSettings.ProductSortingEnumDisabled).ToList();

            //order sorting options
            var orderedActiveSortingOptions = activeSortingOptionsIds
[...]

PR #6279, opened 2022-06-06
#5928 - Allow selecting default product sorting added

fuadhasan28:


RomanovM (changes requested):
@fuadhasanbs23 No need to create a new setting here, let's just use the option with min `DisplayOrder`.

fuadhasan28:
@RomanovM previous commit reverted and proceed with the sorting option with min **DisplayOrder.** Can you please check and confirm?

RomanovM (approved):

Issue #5607, opened 2021-04-12
ACL rules for the multi-factor authentication


RomanovM:
nopCommerce version: 4.40

> I'm trying to figure out how to setup Multi-factor authentication for admins only and I'm having a hard time finding info on this. I did find the setting to enforce it for all users but also don't see how to setup a source for it.

Source: https://www.nopcommerce.com/boards/topic/90742/multi-factor-authentication-setup

So, we have ["ForceMultifactorAuthentication"](https://github.com/nopSolutions/nopCommerce/blob/master/src/Libraries/Nop.Core/Domain/Customers/MultiFactorAuthenticationSettings.cs#L28) setting, which requires all registered users to use multifactor authentication when enabled. It would be better to change this setting to permission (let's name it "Multifactor Authentication"), which will allow control of who needs multifactor authentication to be used.
In migration, we should apply the new permission to all roles (except Guests) if the setting was enabled, and only to Admin if disabled, then remove that setting.

PR #6280, opened 2022-06-06
#5607 - ACL rules for the multi-factor authentication added

fuadhasan28:


src/Libraries/Nop.Data/Migrations/UpgradeTo460/DataMigration.cs
```diff
@@ -220,6 +221,19 @@ T getAttributeValue<T>(IList<GenericAttribute> attributes, string key, Func<stri
                     }
                 );
             }
+
+            //#5607
+            var permissionRecordTypeTable = _dataProvider.GetTable<PermissionRecord>();
+
+            if (!permissionRecordTypeTable.Any(prt => string.Compare(prt.SystemName, "MultiFactorAuthentication", StringComparison.InvariantCultureIgnoreCase) == 0))
```
RomanovM, line 228: Wrong permission name.

src/Libraries/Nop.Data/Migrations/UpgradeTo460/DataMigration.cs
```diff
@@ -220,6 +221,19 @@ T getAttributeValue<T>(IList<GenericAttribute> attributes, string key, Func<stri
...
+                _dataProvider.InsertEntity(
+                    new PermissionRecord
+                    {
+                        SystemName = "ForceMultiFactorAuthentication",
+                        Name = "Public store. Force multi-factor authentication",
+                        Category = "PublicStore"
```
RomanovM, line 234: Let's define a new category (`Security`) since this permission applies to both the public store and the admin area.

src/Libraries/Nop.Services/Security/StandardPermissionProvider.cs
```diff
@@ -128,7 +129,8 @@ public virtual IEnumerable<PermissionRecord> GetPermissions()
                 EnableWishlist,
                 PublicStoreAllowNavigation,
                 AccessClosedStore,
-                AccessProfiling
+                AccessProfiling,
+                ForceMultiFactorAuthentication
```
RomanovM, line 133: No need to apply this permission to any roles by default.

RomanovM (changes requested):
@fuadhasan28 See the comments above and also look at this point in the issue
>In migration, we should apply the new permission to all roles (except Guests) if the setting was enabled, and only to Admin if disabled, then remove that setting.

RomanovM (changes requested):
@fuadhasan28 Still not implemented
>In migration, we should apply the new permission to all roles (except Guests) if the setting was enabled, and only to Admin if disabled, then remove that setting.

fuadhasan28:
> @fuadhasan28 Still not implemented
> 
> > In migration, we should apply the new permission to all roles (except Guests) if the setting was enabled, and only to Admin if disabled, then remove that setting.

@RomanovM Migration added based on the setting. Can you please check again?

src/Libraries/Nop.Data/Migrations/UpgradeTo460/DataMigration.cs
```diff
@@ -220,6 +222,64 @@ T getAttributeValue<T>(IList<GenericAttribute> attributes, string key, Func<stri
...
+                        Category = "Security"
+                    }
+                );
+
+                var isForceMultifactorAuthenticationEnabled = _dataProvider.GetTable<Setting>().Any(s => string.Compare(s.Name, "MultiFactorAuthenticationSettings.ForceMultifactorAuthentication", StringComparison.InvariantCultureIgnoreCase) == 0 && 
+                string.Compare(s.Value, "True", StringComparison.InvariantCultureIgnoreCase) == 0);
+
+                if (isForceMultifactorAuthenticationEnabled)
+                {
+                    //add it to all customer roles by default except guests role
```
RomanovM, line 245: Since only "Registered" role is checked [here](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web.Framework/Mvc/Filters/ForceMultiFactorAuthenticationAttribute.cs#L83), let's also add new permission only for "Registered" role, if the setting was enabled (not for all roles, as was required first).

RomanovM (changes requested):
@fuadhasan28 Still not implemented
> then remove that setting.

src/Libraries/Nop.Data/Migrations/UpgradeTo460/DataMigration.cs
fuadhasan28: @RomanovM resolve the feedback. Can you please check it?

RomanovM: @fuadhasan28 The model and locales are also no longer needed, let's remove them.

fuadhasan28: @RomanovM It has been removed. Can you please check?

RomanovM:
@fuadhasan28 Slightly changed the implementation and decided to keep both the setting and the new permission. Merged directly in develop in [this commit](https://github.com/nopSolutions/nopCommerce/commit/3c4c7f86c4d85feab9d7d9ed40f219ba24fc007b).

Issue #5770, opened 2021-07-21
Implement all the methods from the ValidatorExtensions class consistently


mariannk:
nopCommerce version: 4.40.4

Steps to reproduce the problem: 

1. Open the ValidatorExtensions class (Nop.Web.Framework.Validators)
2. Look at the IsPassword method which has a different implementation than other methods
3. We should implement all these methods one way, consistently.

exileDev:
Closed #5770

PR #6290, opened 2022-06-23
IsPassword method implemented

sohids:


RomanovM:
@sohids What is the purpose of these changes?

sohids:
> @sohids What is the purpose of these changes?

https://github.com/nopSolutions/nopCommerce/issues/5770

RomanovM:
@sohids Well, I see. But #5770 is just about refactoring and consistent implementation of `IsPassword` method with the others (using a separate Validator). No need to add new settings and copy the implementation of `IsUsername` method here. We just need to move method body in a separate validator.
Could you please update your pr?

Issue #6056, opened 2021-12-28
Replace sample pictures 


AndreiMaz:
Let's upgrade the slider graphics on the home page (Nivo Silder). I don’t care about the internal products so much, but the first thing you see are five year old phones.

 It has iPhone 6 and Nokia Lumia 1020!  Makes people wonder if the store is behind the times.

michalwitek1232:
@AndreiMaz is this thing getting deployed right now?
It might be reason for my issue https://github.com/nopSolutions/nopCommerce/issues/6286

DmitriyKulagin:
Closed #6056

Issue #5214, opened 2020-12-08
Update sample data in  the default theme


AndreiMaz:
Currently some sample data is outdated. The use of way outdated items in the demo products & categories give the impression that the product is outdated

We update have to it, make it more modern

iammourya:
I think the nop-template pavilion theme should be the one that should be adopted by the nopcommerce team as the default theme. Enough refinements are done on nop-template team that it would take many years to reach to the same level of "WOW" factor. I'm sure nopcommerce team can work out a deal with nop-template team.

sajasbana:
Hi 
It will be also great if admin template can be customizable like public store so will can be admin themes in market also

goransipic:
Will new theme have support for Bootstrap. Or it will be vanilla CSS. @AndreiMaz

cabbage89:
I think this is an option  ,  https://github.com/cabbage89/Nop.Plugin.Administration.ProAdminUI.Doc

glhays:
Is this available on the official nopcommerce marketplace?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

HarunKilic:
It would be awesome with more modern frontend like Blazor or React/Vue/Angular.

nathanfranke:
Dark theme when the browser requests it?

MeysamNaderi:
The base theme of NopCommerce should prioritize clean design, simple and easily editable code, an enjoyable user experience and SEO optimization. Hence, there's no need for the WOW theme. My recommendation is to create a new theme using a framework like Bootstrap, which will feature a clean and adaptable UI, and utilize AJAX for all functionalities, adaptive (& responsive) design with lower JS/CSS/HTML code.

PR #6299, opened 2022-07-08
#6056 - Upgrade nivoslider sample images with modern phone images

kazirahiv:
Upgraded NivoSlider plugin sample images with these:
banner1
banner2


RomanovM:
@kazirahiv Thanks for your suggestion. We are going to update our theme (see #5214) and then will change the pictures.
P.S. There are typos in the text on your images.

PR #6304, opened 2022-07-14
Allow VAT number to be entered in guest checkout added

mhsjaber:


src/Libraries/Nop.Core/Domain/Tax/TaxSettings.cs
```diff
@@ -107,6 +107,11 @@ public class TaxSettings : ISettings
         /// </summary>
         public bool EuVatEnabled { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether EU VAT (Europe Union Value Added Tax) for guest customers is enabled
+        /// </summary>
+        public bool GuestCustomerVatEnabled { get; set; }
```
RomanovM, line 113: Let's name the setting `EuVatEnabledForGuests`, for similarity with the rest.

src/Presentation/Nop.Web/Controllers/CheckoutController.cs
```diff
@@ -56,6 +58,8 @@ public partial class CheckoutController : BasePublicController
         private readonly IShippingService _shippingService;
         private readonly IShoppingCartService _shoppingCartService;
         private readonly IStoreContext _storeContext;
+        private readonly ITaxService _taxService;
+        private readonly TaxSettings _taxSettings;
```
RomanovM, line 62: We should add new fields alphabetically.

src/Presentation/Nop.Web/Views/Checkout/BillingAddress.cshtml
```diff
@@ -37,6 +37,22 @@
                             </div>
                         </div>
                     }
+                    @if (Model.GuestCustomerVatEnabled || Model.DisplayGuestCustomerVatWarning)
```
RomanovM, line 40: Let's display this field in the new address form only.

src/Presentation/Nop.Web/Views/Checkout/OpcBillingAddress.cshtml
```diff
@@ -11,6 +11,22 @@
             </p>
         </div>
     }
+    @if (Model.GuestCustomerVatEnabled || Model.DisplayGuestCustomerVatWarning)
```
RomanovM, line 14: Let's display this field in the new address form only.

src/Presentation/Nop.Web/Views/Checkout/BillingAddress.cshtml
```diff
@@ -37,6 +37,22 @@
...
+                        <div class="section vat-number">
+                            @if (Model.DisplayGuestCustomerVatWarning)
+                            {
+                                <label class="vat-number-warning">
+                                    @T("Checkout.VatNumber.Warning")
+                                </label>
+                            }
+                            @if (Model.GuestCustomerVatEnabled)
+                            {
+                                <label asp-for="VatNumber">@T("Checkout.VatNumber")</label>
```
RomanovM, line 51: Let's always display this label, and if the setting is disabled, then show a warning alongside.

src/Presentation/Nop.Web/App_Data/Localization/defaultResources.nopres.xml
```diff
@@ -16533,6 +16539,12 @@
   <LocaleResource Name="Checkout.UseRewardPoints">
     <Value>Use my reward points, {0} of {1} reward points ({2}) available for this order</Value>
   </LocaleResource>
+  <LocaleResource Name="Checkout.VatNumber">
+    <Value>VAT number</Value>
+  </LocaleResource>
+  <LocaleResource Name="Checkout.VatNumber.Warning">
+    <Value>VAT number can be entered and used only after registration.</Value>
```
RomanovM, line 16546: We could add a link to the customer info page here so that the user can go, register and immediately enter the VAT number.

RomanovM (changes requested):

PR #6308, opened 2022-07-19
add ellipses for pager

Olek-HZQ:
add pager ellipses


RomanovM:
Thanks for your suggestion. We'll leave this for customization.

PR #6311, opened 2022-07-25
Localized export/import added for Products, Manufacturers and Categories

mhsjaber:


src/Libraries/Nop.Services/ExportImport/Help/LocalizedPropertyByName.cs
```diff
@@ -0,0 +1,245 @@
...
+        private object _propertyValue;
+
+        /// <summary>
+        /// Ctor
+        /// </summary>
+        /// <param name="propertyName">Property name</param>
+        /// <param name="func">Feature property access</param>
+        /// <param name="ignore">Specifies whether the property should be exported</param>
+        /// <returns>A task that represents the asynchronous operation</returns>
+        public LocalizedPropertyByName(string propertyName, Func<T, L, Task<object>> func, bool ignore = false)
```
exileDev, line 24: Hi, @mhsjaber! I guess that we can add this ctor to the PropertyByName class instead of adding a new class. As a result, new duplicate methods will be unnecessary. For example, here we can add a new flag and update the current logic of the PropertyManeger class to use language for this instance. What do you think about it?

mhsjaber:
> Hii, @mhsjaber! Looks good. Can you update your PR based on the latest develop branch?

Updated

PR #6319, opened 2022-07-29
Teste failing

mgracietti:
Teste failing because extension of imagem file was incorrect (.jpeg instead of .jpg)


RomanovM:
@mgracietti Thanks, but already been fixed by [this commit](https://github.com/nopSolutions/nopCommerce/commit/4600a5ca600f85d64e5fbcf339862a0498227006).

PR #6329, opened 2022-08-05
Added user agent for Azure Traffic Manager and Azure AppService AlwaysOn as crawlers

szilardd:
### Azure Traffic Manager

When using [Azure Traffic Manager endpoint monitoring](https://docs.microsoft.com/en-us/azure/traffic-manager/traffic-manager-monitoring), if the Path is set to `/`, a guest Customer will be created every time the traffic manager calls the url. This can be disabled by configuring the user agent as a crawler:

```xml
<browscapitem name="Azure Traffic Manager Endpoint Monitor"> <!-- Azure Traffic Manager -->
  <item name="Crawler" value="true" />
</browscapitem>
```

### Azure App Service AlwaysOn

When using [Azure App Service Always On](https://docs.microsoft.com/en-us/azure/app-service/configure-common?tabs=portal#configure-general-settings) functionality, it will call the root url every 5 minutes, which also creates a guest Customer on each request. This can be disabled by configuring the `Always On` user agent as a crawler:

```xml
<browscapitem name="AlwaysOn"> <!-- Azure AppService AlwaysOn -->
  <item name="Crawler" value="true" />
</browscapitem>
```

### DIFF
--- a/src/Presentation/Nop.Web/App_Data/browscap.crawlersonly.xml
+++ b/src/Presentation/Nop.Web/App_Data/browscap.crawlersonly.xml
@@ -14004,4 +14004,10 @@
   <browscapitem name="Core_Class_HttpClient_Cached">
     <item name="Crawler" value="true" />
   </browscapitem>
+  <browscapitem name="Azure Traffic Manager Endpoint Monitor"> <!-- Azure Traffic Manager -->
+    <item name="Crawler" value="true" />
+  </browscapitem>
+  <browscapitem name="AlwaysOn"> <!-- Azure AppService AlwaysOn -->
+    <item name="Crawler" value="true" />
+  </browscapitem>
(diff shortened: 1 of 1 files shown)


skoshelev:
Thanks for the suggestion, however, we generate the browscap.crawlersonly.xml file based on data from the http://browscap.org/ site, this file can also be regenerated by the site administrator (the instruction is described in the headers of the browscap.crawlersonly.xml file itself). It follows that adding any data manually to this file is not a good idea. In addition, the entries that you propose to make are very specific. In 4.70 it will be possible to add custom entries through the additional.crawlers.xml file. So any user will be able to create their own list of additional records without the risk of losing them and the need to restore them every time the data is updated. However, we will add your recommendations to the documentation for deploying the site on the Azure.

PR #6373, opened 2022-10-06
Multiple pictures associated to product attribute value and product attribute combination

mhsjaber:


src/Libraries/Nop.Data/Migrations/UpgradeTo460/Domain/ProductAttributeCombination450.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Nop.Core;
+using Nop.Core.Domain.Catalog;
+
+namespace Nop.Data.Migrations.UpgradeTo460.Domain
+{
+    public class ProductAttributeCombination450 : ProductAttributeCombination
```
skoshelev, line 6: I don't like this idea, it's too complicated, it's better to leave this field in the original class and add a description in a comment that the field is no longer used, or move this field to a partial class and still add a comment that the field is left for backwards compatibility during migration

src/Libraries/Nop.Data/Migrations/UpgradeTo460/DataMigration.cs
```diff
@@ -5,19 +5,22 @@
...
+using Nop.Core.Domain.Media;
 using Nop.Core.Domain.Messages;
 using Nop.Core.Domain.ScheduleTasks;
 using Nop.Core.Domain.Security;
+using Nop.Data.Migrations.UpgradeTo460.Domain;
 
 namespace Nop.Data.Migrations.UpgradeTo460
 {
-    [NopMigration("2022-02-03 00:00:00", "4.60.0", UpdateMigrationType.Data, MigrationProcessType.Update)]
+    [NopMigration("2022-02-03 00:00:00", "DataMigration for 4.60.0", MigrationProcessType.Update)]
```
skoshelev, line 23: I don’t understand why these changes were made, especially since they break the logic of running this migration every time the debug mode is set in the studio (see https://github.com/nopSolutions/nopCommerce/blob/061ea8c69312bab8c7b08951d54858f23af964d5/src/Libraries/Nop.Data /Migrations/MigrationManager.cs#L125-L129)

src/Libraries/Nop.Services/ExportImport/ExportManager.cs
```diff
@@ -588,8 +590,12 @@ private async Task<int> ExportProductAttributesAsync(Product item, PropertyManag
                                 Quantity = pav.Quantity,
                                 IsPreSelected = pav.IsPreSelected,
                                 DisplayOrder = pav.DisplayOrder,
-                                PictureId = pav.PictureId
-                            });
+                                PictureIds = string.Join(",",
```
skoshelev, line 593: better to use ; as separator

src/Presentation/Nop.Web/Views/Product/_ProductAttributes.cshtml
```diff
@@ -492,6 +492,23 @@
                         if (data.pictureFullSizeUrl) {
                             $('#main-product-img-lightbox-anchor-@Model.Id').attr("href", data.pictureFullSizeUrl);
                         }
+                        @if (Model.DisplayAttributeCombinationImagesOnly) 
+                        {
+                            <text>
+                            $.each($('.thumb-item'), function(i, elem) { 
+                                var id = $(elem).data('pictureid');
+                                console.log(id)
```
skoshelev, line 500: please don't forget to remove the debug code

src/Libraries/Nop.Core/Domain/Catalog/ProductEditorSettings.cs
```diff
@@ -232,6 +232,11 @@ public class ProductEditorSettings : ISettings
         /// </summary>
         public bool AllowAddingOnlyExistingAttributeCombinations { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether to display attribute combination images only
+        /// </summary>
+        public bool DisplayAttributeCombinationImagesOnly { get; set; }
```
skoshelev, line 238: the setting is used only to control visibility on the product editing page, which is not entirely correct, since when it is disabled, all products for which the analogous property was set to true will retain their behavior, which can confuse users

src/Presentation/Nop.Web.Framework/Migrations/UpgradeTo460/LocalizationMigration.cs
```diff
@@ -285,6 +291,15 @@ public override void Up()
...
                 ["Admin.Configuration.Stores.Ssl.Updated"] = "The SSL setting has been successfully changed. Do not forget to synchronize the store URL with the current HTTP protocol.",
+
+                //1934
+                ["Admin.Catalog.Products.ProductAttributes.AttributeCombinations.Fields.Pictures"] = "Pictures",
+                ["Admin.Catalog.Products.ProductAttributes.AttributeCombinations.Fields.Pictures.Hint"] = "Choose pictures associated to this attribute combination. These pictures will replace the main product image when this product attribute combination is selected.",
+                ["Admin.Catalog.Products.ProductAttributes.Attributes.Values.Fields.Pictures"] = "Pictures",
+                ["Admin.Catalog.Products.ProductAttributes.Attributes.Values.Fields.Pictures.Hint"] = "Choose pictures associated to this attribute combination. These pictures will replace the main product image when this product attribute combination is selected.",
+                ["Admin.Configuration.Settings.ProductEditor.DisplayAttributeCombinationImagesOnly"] = "Display attribute combination images only",
```
skoshelev, line 306: No detailed description for customization (.Hint)

PR #6389, opened 2022-10-18
Topic/story 30637 test suite

addabis:
After a few rounds of fixes of Metalama.


addabis:
This was merging into a wrong repo, sorry for that.

PR #6390, opened 2022-10-19
Update Dockerfile

LaplancheMaxime:
Added port 80 as default

Currently no port is exposed by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,5 +97,7 @@ RUN chmod 755 /entrypoint.sh
 WORKDIR /app
 
 COPY --from=build /app/published .
+
+EXPOSE 80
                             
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @LaplancheMaxime. we did not indicate that the container is waiting on port 80, since this behavior seemed obvious to us. Tell me, please, if you have any additional arguments in favor of your changes, maybe I don’t quite understand the essence of the EXPOSE instruction. And shouldn't we also specify ports 5000 and 5001, which are opened by the dotnet Nop.Web.dll command?

LaplancheMaxime:
Hello, 

The EXPOSE command allows you to "open" the ports of the container. 
If they are not exposed, the container cannot communicate. 
What are the ports 5000 and 5001 used for?

PR #6416, opened 2022-11-15
Refactor customer addresses cache

holydk:


src/Libraries/Nop.Services/Common/Caching/AddressCacheEventConsumer.cs
```diff
@@ -11,13 +11,17 @@ namespace Nop.Services.Common.Caching
...
     {
         /// <summary>
-        /// Clear cache data
+        /// Clear cache by entity event type
         /// </summary>
         /// <param name="entity">Entity</param>
+        /// <param name="entityEventType">Entity event type</param>
         /// <returns>A task that represents the asynchronous operation</returns>
-        protected override async Task ClearCacheAsync(Address entity)
+        protected override async Task ClearCacheAsync(Address entity, EntityEventType entityEventType)
```
skoshelev, line 19: I don't quite understand why it is not necessary to update the cache when adding a new address?

src/Libraries/Nop.Services/Customers/Caching/CustomerAddressMappingCacheEventConsumer.cs
```diff
@@ -10,13 +10,19 @@ namespace Nop.Services.Customers.Caching
...
+        /// Clear cache by entity event type
         /// </summary>
         /// <param name="entity">Entity</param>
+        /// <param name="entityEventType">Entity event type</param>
         /// <returns>A task that represents the asynchronous operation</returns>
-        protected override async Task ClearCacheAsync(CustomerAddressMapping entity)
+        protected override async Task ClearCacheAsync(CustomerAddressMapping entity, EntityEventType entityEventType)
         {
-            await RemoveByPrefixAsync(NopCustomerServicesDefaults.CustomerAddressesPrefix);
+            await RemoveAsync(NopCustomerServicesDefaults.CustomerAddressesCacheKey, entity.CustomerId.ToString());
```
skoshelev, line 20: I see no reason to use ToString

src/Libraries/Nop.Services/Customers/Caching/CustomerCacheEventConsumer.cs
```diff
@@ -23,6 +23,23 @@ public async Task HandleEventAsync(CustomerPasswordChangedEvent eventMessage)
...
+        /// </summary>
+        /// <param name="entity">Entity</param>
+        /// <param name="entityEventType">Entity event type</param>
+        /// <returns>A task that represents the asynchronous operation</returns>
+        protected override async Task ClearCacheAsync(Customer entity, EntityEventType entityEventType)
+        {
+            if (entityEventType == EntityEventType.Delete)
+            {
+                await RemoveAsync(NopCustomerServicesDefaults.CustomerAddressesCacheKey, entity);
+                await RemoveByPrefixAsync(NopCustomerServicesDefaults.CustomerAddressesByCustomerPrefix, entity);
```
skoshelev, line 37: For this code to work properly, you must also change the CustomerAddressCacheKey key, since it is now erroneous

skoshelev:
Hi @holydk. I looked at the refactoring you suggested and left a couple of recommendations, please make the necessary changes

src/Libraries/Nop.Services/Common/Caching/AddressCacheEventConsumer.cs
holydk: When we add a new address, it is not yet associated with any customer

src/Libraries/Nop.Services/Customers/Caching/CustomerAddressMappingCacheEventConsumer.cs
holydk: This is a small optimization so we avoid extra boxing (int -> object).

skoshelev: Instead, convert to a string, I'm not even sure which is faster, in any case, such an entry looks redundant

src/Libraries/Nop.Services/Customers/Caching/CustomerCacheEventConsumer.cs
skoshelev: Yes that's right

src/Libraries/Nop.Services/Common/Caching/AddressCacheEventConsumer.cs
skoshelev: Perhaps, but I'm not completely sure that this is the case, especially considering that the standard behavior can be changed by some kind of plugin. I don't think the idea of removing the insert event is safe

holydk: It seems to me that we will gain more performance if we do not remove all cached addresses from all users each time an address is added.

src/Libraries/Nop.Services/Customers/Caching/CustomerCacheEventConsumer.cs
holydk: Fixed

src/Libraries/Nop.Services/Customers/Caching/CustomerAddressMappingCacheEventConsumer.cs
holydk: Returned as it was, in a single case it will not give any advantages

src/Libraries/Nop.Services/Common/Caching/AddressCacheEventConsumer.cs
skoshelev: I would agree with this thesis if it were not a potential problem, in such a case as this, it is better to delete the data than not to delete it at the right time

holydk: Ok, let's leave it as it was, probably this issue needs to be approached more comprehensively.

skoshelev:
Thank you @holydk

PR #6469, opened 2022-12-17
NopEngine methods are marked as virtual.

aboyko3:
This would make it possible to reuse the NopEngine in case it needs to be replaced with a custom/extended one.


skoshelev:
Hi, @aboyko3. Thank you for the offer, of course it is unlikely that someone will change something in this area, but let there be such an opportunity

PR #6478, opened 2022-12-21
Build - ignore Plugins folder for linux

deathgore:
Windows \Plugins\ is hardcoded in the initial dll scans.
When calling ClearPluginAssemblies from linux Plugins folder is not ignored.

Fixing this by replacing \ with Path.DirectorySeparatorChar


skoshelev:
Hi @deathgore. Thank you for your help

PR #6509, opened 2023-01-02
remove-blocking-call-to-session-cache-from-facebookpixel-plugin

Einarsson:
This change removes a blocking call to the session cache in facebook pixel plugin.

When calling `HttpContext.Session.Get` you have to perform the load beforehand explicitly using `LoadAsync()`, otherwise the call to the session cache will be synchronous, if you're using a distributed cache (for example Redis) this can create big bottlenecks and performance issues.

As described in this article: [Session and state management in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-7.0).

Blocking this call will cause thread starvation. This is what happened to our customers website if Redis response times temporarily increased.

Thread Count:
image

Page responsetime:
image

There are some more places where the session cache is called synchronously (Checkout, for example), but the impact is low. However, facebook pixel loads on every page and has massive impact on performance.

It might be advisable to wrap this functionality to enforce an async pattern globally in nopCommerce.

Cheers,
Emil Einarsson
CTO & senior developer, Majako
Web: [www.majako.se](https://www.majako.se/)

image

### DIFF
--- a/src/Libraries/Nop.Core/Http/Extensions/SessionExtensions.cs
+++ b/src/Libraries/Nop.Core/Http/Extensions/SessionExtensions.cs
@@ -36,6 +36,27 @@ namespace Nop.Core.Http.Extensions
             return value == null ? default : JsonConvert.DeserializeObject<T>(value);
         }
 
+        /// <summary>
+        /// Remove the given key from session if present.
+        /// </summary>
+        /// <param name="session">Session</param>
+        /// <param name="key">Key</param>
+        public static async Task RemoveAsync(this ISession session, string key)
+        {
+            await LoadAsync(session);
+            session.Remove(key);
+        }
+
+        /// <summary>
+        /// Remove all entries from the current session, if any. The session cookie is not removed.
+        /// </summary>
+        /// <param name="session">Session</param>
+        public static async Task ClearAsync(this ISession session)
+        {
+            await LoadAsync(session);
+            session.Clear();
+        }
+
         private static async Task LoadAsync(ISession session)
         {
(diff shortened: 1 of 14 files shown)


skoshelev:
Hi @Einarsson. Thank you very much for your analysis. I also studied this issue and agree that such optimization is necessary for the FacebookPixel plugin. However, I'm not sure that such an optimization will bring a significant increase in the nopCommerce core. However, I created a separate [ticket](https://github.com/nopSolutions/nopCommerce/issues/6520) for this task.

As for your code, you should take into account that the LoadAsync method can throw an exception when called directly and their handling falls on the shoulders of developers, let's add exception catching to avoid such a situation

Einarsson:
I suppose you could try-catch the call, however if it can't get the session, won't the code crash on the next line anyway when it tries again? If you're unable to access your Session the call should fail imo.

skoshelev:
This is not entirely true, if you look at the implementation of an asynchronous method and a synchronous one, you can see a number of differences due to which the asynchronous version may fail and the synchronous version that is called later will work fine and access to the session will be obtained without problems.

rickardvh:
@skoshelev I have modified the session extension methods according to [this article](https://bartwullems.blogspot.com/2019/12/aspnet-core-load-session-state.html) (with added error handling as requested), and replaced all uses of them solution-wide. Hopefully this should result in a more general performance improvement :)

skoshelev:
Hi, @rickardvh. Thank you very much for your help.

Issue #6520, opened 2023-01-09
Add async version of Get, Set and Remove methods for SessionExtensions


skoshelev:
When calling HttpContext.Session.Get you have to perform the load beforehand explicitly using LoadAsync(), otherwise the call to the session cache will be synchronous, if you're using a distributed cache (for example Redis) this can create big bottlenecks and performance issues.

see details: https://github.com/nopSolutions/nopCommerce/pull/6509#issue-1516466263

skoshelev:
closed by https://github.com/nopSolutions/nopCommerce/commit/598580c77415f602ccfce8d7e471274f4371f344

PR #6529, opened 2023-01-13
Issue 6520, add async version of Get, Set and Remove methods for SessionExtensions

duc-tv:
This is the fix for [issue #6520](https://github.com/nopSolutions/nopCommerce/issues/6520)

When calling HttpContext.Session.Get you have to perform the load beforehand explicitly using LoadAsync(), otherwise the call to the session cache will be synchronous, if you're using a distributed cache (for example Redis) this can create big bottlenecks and performance issues.

### DIFF
--- a/src/Libraries/Nop.Core/Http/NopDistributedSession.cs
+++ b/src/Libraries/Nop.Core/Http/NopDistributedSession.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Session;
+using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Logging;
+
+namespace Nop.Core.Http
+{
+    /// <summary>
+    /// An <see cref="ISession"/> backed by an <see cref="IDistributedCache"/>.
+    /// </summary>
+    public class NopDistributedSession : DistributedSession, ISession
+    {
+        private bool _loaded = false;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="NopDistributedSession"/>.
+        /// </summary>
+        /// <param name="cache">The <see cref="IDistributedCache"/> used to store the session data.</param>
+        /// <param name="sessionKey">A unique key used to lookup the session.</param>
+        /// <param name="idleTimeout">How long the session can be inactive (e.g. not accessed) before it will expire.</param>
+        /// <param name="ioTimeout">
+        /// The maximum amount of time <see cref="LoadAsync(CancellationToken)"/> and <see cref="CommitAsync(CancellationToken)"/> are allowed take.
+        /// </param>
+        /// <param name="tryEstablishSession">
+        /// A callback invoked during <see cref="Set(string, byte[])"/> to verify that modifying the session is currently valid.
+        /// If the callback returns <see langword="false"/>, <see cref="Set(string, byte[])"/> throws an <see cref="InvalidOperationException"/>.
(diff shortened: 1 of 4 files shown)


skoshelev:
Hi, @duc-tv. Thank you very much for your help, however your code doesn't solve any problem, it's basically just synchronous code wrapped in asynchronous notation. The essence of the ticket was to organize exactly the asynchronous loading of session data. details are described [here ](https://github.com/nopSolutions/nopCommerce/pull/6509 )and [here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-7.0# load-session-state-asynchronously)

duc-tv:
@skoshelev, I have updated the source code, pls check it for me.

src/Libraries/Nop.Core/Http/NopDistributedSession.cs
```diff
@@ -0,0 +1,83 @@
...
+                try
+                {
+                    LoadAsync().Wait();
+                }
+                catch
+                {
+                    //session will be loaded synchronously.
+                }
+
+                _loaded = true;
```
skoshelev, line 79: This code will make it impossible to load in the synchronous version, and therefore, if an error occurs in the asynchronous operation, the session data will not be loaded at all.

skoshelev:
Hi @duc-tv. Thank you for your help, the last option you suggested will work (if you'll move the _loaded = true instruction from the unsafe place), but it's redundant in our code. Recently another user suggested a better option, you can see his changes in [this commit](https://github.com/nopSolutions/nopCommerce/pull/6509/commits/d58b132a2a26dea1041b6f130d946ec330a1785e)

duc-tv:
> Hi @duc-tv. Thank you for your help, the last option you suggested will work (if you'll move the _loaded = true instruction from the unsafe place), but it's redundant in our code. Recently another user suggested a better option, you can see his changes in [this commit](https://github.com/nopSolutions/nopCommerce/pull/6509/commits/d58b132a2a26dea1041b6f130d946ec330a1785e)

Thanks for your reply.
However, access to the [ISession.IsAvailable](https://github.com/nopSolutions/nopCommerce/pull/6509/commits/d58b132a2a26dea1041b6f130d946ec330a1785e#diff-c0d6540b56097be581546b65181ea4cf2dbda0bdae351abfea390d8912f22120R41) property, I think that's not necessarily the optimal solution.
See the implement of the [ISession.IsAvailable](https://github.com/aspnet/Session/blob/9e4b6a5e4a8d4c789b627a37b0342dad45c87b6c/src/Microsoft.AspNetCore.Session/DistributedSession.cs#L78) property and [DistributedSession](https://github.com/aspnet/Session/blob/master/src/Microsoft.AspNetCore.Session/DistributedSession.cs),  if you access this property before the Session is loaded, the Session will be loaded synchronously. Therefore, in most cases, ISession.LoadAsync will never work and this issue is still unresolved.

src/Libraries/Nop.Core/Http/NopDistributedSession.cs
duc-tv: [NopDistributedSession._loaded] field is responsible for calling [DistributedSession.LoadAsync] or not, does not affect [DistributedSession._loaded], so when an error occurs, method [DistributedSession.Load] will still be called because [DistributedSession._loaded] = false.
If the naming of this variable confuses you, I will correct the source code.

-> I think in all cases the Session will be synchronized if the asynchronous load has failed

skoshelev:
@duc-tv. Yes, you are right, thanks for pointing out this feature, I [removed ](https://github.com/nopSolutions/nopCommerce/commit/7040681f47ded4a141333900b707a708b4bd33f2)the extra check

PR #6531, opened 2023-01-14
Caching improvements

rickardvh:
This pull request addresses a few performance issues with the current caching infrastructure:

- value factories are wrapped in Lazy instances to avoid spawning multiple acquisition threads performing the same work
- keys are stored in a trie (prefix tree) to optimise removing by prefix, bringing the time complexity of finding keys by prefix down from $O(n|k|)$ to $O(|k|)$, where $n$ is the number of keys and $|k|$ is the length of the prefix

The existing implementations of the distributed cache also suffers from a bug where prefixes are not case-sensitive while the actual cache keys are. This has been fixed so both are case-sensitive.

### New Redis-synchronised memory cache
Furthermore, a more performant implementation of a Redis-based cache is added to make distributed and scalable hosting feasible. This new RedisSynchronizedMemoryCache does not actually store cache entries on Redis, but uses Redis' pub/sub functionality to keep instances in sync. Although this pattern is easily extensible to storage and retrieval of values with in-memory L1 caching, the hybrid approach has the advantage of using almost no storage and not hinging on high throughput. Therefore, one can use a very small and cheap Redis server while still getting the performance of a regular memory cache.

Redis settings now include the option to set the instance name, which allows partitioning of a single Redis server between different applications (e.g. development and production environments).

### Support for background tasks
Finally, Locker implementations have been refactored to separate classes (due to the introduction of RedisSynchronizedMemoryCache, which is used by a regular MemoryCacheManager but needs to store task locks as Redis keys), and extended to include support for long-running background tasks. Background tasks can be run with a "heartbeat": a status flag that is periodically updated by the executing nopCommerce instance as long as the task is running. Other instances can observe this flag to avoid starting the same task, and the heartbeat will be automatically expired after a given timespan if the application is unexpectedly shutdown. Majako has two plugins using this functionality waiting to be published.

Best regards,

Rickard von Haugwitz
Majako
majako

### DIFF
--- a/src/Libraries/Nop.Core/Caching/MemoryCacheManager.cs
+++ b/src/Libraries/Nop.Core/Caching/MemoryCacheManager.cs
@@ -1,18 +1,17 @@
 ﻿using System;
-using System.Collections.Concurrent;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Primitives;
 using Nop.Core.Configuration;
+using Nop.Core.Infrastructure;
 
 namespace Nop.Core.Caching
 {
     /// <summary>
     /// Represents a memory cache manager 
     /// </summary>
-    public partial class MemoryCacheManager : CacheKeyService, ILocker, IStaticCacheManager
+    public partial class MemoryCacheManager : CacheKeyService, IStaticCacheManager
     {
         #region Fields
 
@@ -20,8 +19,7 @@ namespace Nop.Core.Caching
         private bool _disposed;
 
         private readonly IMemoryCache _memoryCache;
-
-        private static readonly ConcurrentDictionary<string, CancellationTokenSource> _prefixes = new();
+        private static readonly ConcurrentTrie<byte> _keys = new();
         private static CancellationTokenSource _clearToken = new();
 
         #endregion
@@ -42,7 +40,7 @@ namespace Nop.Core.Caching
         /// </summary>
         /// <param name="key">Cache key</param>
         /// <returns>Cache entry options</returns>
-        private MemoryCacheEntryOptions PrepareEntryOptions(CacheKey key)
+        private static MemoryCacheEntryOptions PrepareEntryOptions(CacheKey key)
         {
             //set expiration time for the passed cache key
             var options = new MemoryCacheEntryOptions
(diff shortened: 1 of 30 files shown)


src/Libraries/Nop.Core/Caching/MemoryCacheManager.cs
```diff
@@ -42,47 +40,36 @@ public MemoryCacheManager(AppSettings appSettings, IMemoryCache memoryCache) : b
...
+        private static void OnEviction(object key, object value, EvictionReason reason, object state)
         {
-            if ((key?.CacheTime ?? 0) <= 0 || data == null)
-                return;
-
-            _memoryCache.Set(key.Key, data, PrepareEntryOptions(key));
+            switch (reason)
+            {
+                // we clean up after ourselves elsewhere
+                case EvictionReason.Removed:
```
rickardvh, line 64: Note that a key might be removed by the Redis-synced cache and thus wouldn't be removed from _keys here. That would by no means be a big problem (worst case, RemoveByPrefix would try to remove a few nonexistent keys), but I'll do something about it anyway for the sake of cleanliness.

rickardvh: Resolved by commit [9f5b144](https://github.com/nopSolutions/nopCommerce/pull/6531/commits/9f5b1445f996e1d3daa8f130cbe16dd251e44f8a)

src/Libraries/Nop.Core/Caching/MemoryCacheManager.cs
```diff
@@ -117,15 +107,16 @@ public async Task<T> GetAsync<T>(CacheKey key, Func<Task<T>> acquire)
...
-            return result;
+            var value = await _memoryCache.GetOrCreate(
+                key.Key,
+                entry =>
+                {
+                    entry.SetOptions(PrepareEntryOptions(key));
+                    return new Lazy<Task<T>>(acquire, true);
+                }).Value;
+            if (value == null)
+                await RemoveAsync(key);
```
rickardvh, line 118: This is problematic - because the fact that caching null is disallowed is (ab-)used by services to avoid setting values, RedisSynchronizedMemoryCache unnecessarily spams update events for removing null values. I think a better pattern would be to allow and guard against the possibility of null being cached, and where necessary pass `acquire = null` (instead of `acquire = () => null`.

rickardvh: Resolved in commit [1570d47](https://github.com/nopSolutions/nopCommerce/pull/6531/commits/1570d47bbb26fca389f7b3416327317b82452d52)

skoshelev:
Hi, @rickardvh. I looked at your changes and made some adjustments. You did a great job, thank you very much for your work

cyates81:
So did this get included with the 4.7 release? I don't see it in the release notes.

glhays:
> So did this get included with the 4.7 release? I don't see it in the release notes.

I found it in the release notes for 4.70
https://www.nopcommerce.com/en/release-notes

cyates81:
oh, my bad! Excited to try it out, our load is too high to use redis without the L1. Thanks!

PR #6533, opened 2023-01-17
Use async main

rickardvh:
Since C# 7.1, main can be declared async (see [this article](https://learn.microsoft.com/en-us/archive/blogs/mazhou/c-7-series-part-2-async-main#the-async-main-method)). By using this, we can remove some Task.Wait calls and simplify startup.

Best regards,

Rickard von Haugwitz
Majako
majako


skoshelev:
Hi, @rickardvh. Thanks a lot for your work

PR #6536, opened 2023-01-18
Cache lookups instead of lists

rickardvh:
Cache categories, localized entities and URL records as lookups for faster retrieval.

Currently, CategoryService, LocalizedEntityService and UrlRecordService cache their respective entities as one big list, and every new lookup is performed by querying this list. Every such operation has a time complexity in $O(n)$, where $n$ is the total number of entities. Since the number of entities in these services can easily number in the thousands, in some cases hundreds of thousands, this quickly becomes a problem.

To address this issue, we instead cache these entities as lookups (specifically, lookup-like dictionaries, since actual lookups cannot be JSON-serialised), reducing retrieval complexity to $O(1)$ (subsequent filtering and sorting of the retrieved lists is negligible, as they are typically very small). Where applicable, the lookups are broken up into smaller pieces (e.g. storing localizations by language) to avoid added complexity from having to store nested lookups, and to reduce the size of the objects for faster serialisation and transfer in scenarios like distributed caching.

Best regards,

Rickard von Haugwitz
Majako
majako


skoshelev:
Hi, @rickardvh. Good work. Thank you.

PR #6550, opened 2023-01-25
Don't pass token to remove in DistributedCacheLocker

rickardvh:
- Don't pass the cancellation token to _distributedCache.RemoveAsync, as it may be in a cancelled state already and we want to remove the key regardless (the tests pass anyway because they use the memory distributed cache, which likely ignores the token)
- Now that we're using a synchronous callback, it is guaranteed to have finished executing when Timer.DisposeAsync has been called (note that the regular parameterless Dispose has no such guarantee), so we don't need the heartbeat token anymore


skoshelev:
Hi, @rickardvh. I agree that such a decision would be correct. Thank you for your help

Issue #6549, opened 2023-01-25
Icon moves when expanding Admin menu sections


mike-warren-dev:
**nopCommerce version:** 

4.70.0

**Steps to reproduce the problem:** 

1) Navigate to Admin page. 
2) Watching the caret icon closely, expand the 'Configuration' section. Note that the caret rotates, but does not move/translate:
![image](https://user-images.githubusercontent.com/90517350/214605047-a289defa-4c20-4cc9-81ab-6729590c2acb.png)

3) Keep the 'Configuration' section expanded
4) Watching the caret icon closely, expand the 'Settings' section. Note that the caret rotates and moves/translates down and to the right: 
![image](https://user-images.githubusercontent.com/90517350/214605766-45daebed-e239-4c8b-bc8f-2cb04f345d46.png)

I believe I understand the cause of the issue and would be happy to submit a PR.

AndreiMaz:
@mike-warren-dev Thanks a lot for reporting. Su,re you're more than welcome to submit a pull request

exileDev:
Closed #6549

PR #6554, opened 2023-01-27
change caret icon styling to prevent movement when clicked

mike-warren-dev:
This resolves issue #6549 . 

Previously, when clicking a child-level item in the Admin menu, the caret would move down and to the right: 

image

Now it stays put: 

image

**Changes:** 

Line 1354 -- Removed the **min-width** styling. This was being used to move child-items' carets left/inward, and was the cause of this issue. Removing this necessitates lines 1356-1358.

Line 1352 -- Narrowed the scope of the **padding-top** styling, which is needed for the 'circle' icon (this styling pushes it down toward the vertical center-line), but not the 'caret' icon (the styling pushes it down/away from the vertical center-line).

image

Lines 1356-1358 -- override AdminLTE.io styling to move child-level caret icons left/inward.

### DIFF
--- a/src/Presentation/Nop.Web/wwwroot/css/admin/styles.css
+++ b/src/Presentation/Nop.Web/wwwroot/css/admin/styles.css
@@ -1349,9 +1349,12 @@ input[type="text"].k-input {
   align-items: flex-start;
 }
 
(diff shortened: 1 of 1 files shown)


exileDev:
Thanks!

Issue #6572, opened 2023-02-05
ValidatePictureAsync throws an exception when decoding an SVG image


MarinaAndreeva:
nopCommerce version: 4.60

Steps to reproduce the problem: 

1. Go to _Catalog -> Products page_
2. Choose any product
3. Press _Edit_ button
4. Press _Upload files_ button and choose an SVG image
5. Press _Save_ (or don't - it doesn't matter)
6. Go to _System -> Log page_
7. There is a message with "Cannot decode picture binary (file name: filename)"

Full text:
```
System.ArgumentNullException: Value cannot be null. (Parameter 'codec')
   at SkiaSharp.SKBitmap.Decode(SKCodec codec)
   at SkiaSharp.SKBitmap.Decode(ReadOnlySpan`1 buffer)
   at SkiaSharp.SKBitmap.Decode(Byte[] buffer)
   at Nop.Services.Media.PictureService.ValidatePictureAsync(Byte[] pictureBinary, String mimeType, String fileName) in C:\directory\nopCommrece_clean\nopCommerce\src\Libraries\Nop.Services\Media\PictureService.cs:line 1066
```

If _Display order_ of uploaded SVG image isn't 0 (and this is not a single product image), then the above error will be repeated for all product images (any format, not just SVG).

parshan2:
As I investigated SkiaSharp can not decode an SVG file to a bitmap image (this is due to the nature of SVG format).
also resizing a SVG file is meaningless, and it most be saved on other picture formats after image size determined.
By the way I think the solution is to set `validateBinary=false` for `GetDownloadBitsAsync` method when inserting new SVG picture by `InsertPictureAsync` method.

DmitriyKulagin:
Closed #6572

PR #6580, opened 2023-02-08
avoid validation picture binary for svg images #6572

parshan2:


DmitriyKulagin:
@parshan2 Thanks for your suggestion, the solution is correct, but I simplified it a bit more.

Issue #6546, opened 2023-01-24
HTML code is showing in PDF when checkout attribute control type is File upload


Muditshah30:
nopCommerce version: 4.60.1

Steps to reproduce the problem: 
1. Redirect on Admin > Catalog > Attribute > Checkout attribute
2. Add new Checkout attribute with Control type is "File Upload"
3. Now, Add product in cart and upload file in Checkout attribute
4. Placed order and click on PDF invoice
5. Observe that anchor tag is shown in Order PDF

![image](https://user-images.githubusercontent.com/121164051/214257636-59902d8d-0bf5-4c9e-b04e-7b91a033646a.png)

DmitriyKulagin:
Closed #6546

PR #6583, opened 2023-02-08
HTML code convert to plain text in PDF when checkout attribute control type is File upload #6546

parshan2:
…l type is File upload #6546


DmitriyKulagin:
@parshan2  Thank you for your suggestion, it is generally correct, but in this case, you can only get by with this method - `_htmlFormatter.ReplaceAnchorTags`

Issue #6579, opened 2023-02-08
Remove unused services from AvalaraTaxManager class


skoshelev:
The AvalaraTaxManager class contains following unused services:

- IActionContextAccessor
- ICheckoutAttributeService
- ILocalizationService
- ISettingService
- ITaxPluginManager
- IUrlHelperFactory
- IWebHelper
- WidgetSettings

parshan2:
`ICheckoutAttributeService` is used in `CreateLinesForCheckoutAttributesAsync`

PR #6584, opened 2023-02-08
unused services removed from AvalaraTaxManager class #6579

parshan2:


RomanovM:
@parshan2 You should also bump a version in `plugin.json` file.

parshan2:
@RomanovM version bumped to 2.69

RomanovM:
@parshan2 Thanks.

Issue #6577, opened 2023-02-08
Remove unused IDownloadService from OrderController


skoshelev:
The OrderController class contains an unused IDownloadService service, it should be removed as unnecessary

PR #6585, opened 2023-02-08
unused IDownloadService removed from OrderController #6577

parshan2:


RomanovM:
@parshan2 Thanks for your help.

PR #6596, opened 2023-02-14
typo fix for Topic

parshan2:
Correction of the word [**_topic_**], which is typed as [**_opic_**] in most methods summary.


RomanovM:
@parshan2 Thanks.

PR #6616, opened 2023-02-25
Fix typos

successgo:


RomanovM:
@successgo Thanks for your help.

PR #6618, opened 2023-02-28
Undo null caching

rickardvh:
A bug seems to have been introduced regarding handling of null values in the cache, so we have made the following fixes:

- don't cache nulls in DistributedCacheManager
- the `Lazy<Task<T>>` instance from `_memoryCache.GetOrCreate` in MemoryCacheManager cannot be null, but the resulting `T` value might be - since we won't know until the task is awaited, we have to remove the key if the result is null
- as a result of the above, thread interleaving could cause a key to return a null result before it has been removed, so we need to deal with that possibility in the default value overload of `GetAsync`

I believe it would be much better if we did allow for the possibility of null caching and instead dealt with it outside the cache manager, though. Thoughts?

Best regards,
Rickard von Haugwitz
Majako [majako.net](https://www.majako.net/)

### DIFF
--- a/src/Libraries/Nop.Core/Caching/DistributedCacheManager.cs
+++ b/src/Libraries/Nop.Core/Caching/DistributedCacheManager.cs
@@ -177,7 +177,7 @@ namespace Nop.Core.Caching
                 {
                     item = (T)await lazy.Value;
 
-                    if (key.CacheTime == 0)
+                    if (key.CacheTime == 0 || item == null)
                         return item;
 
                     setTask = _distributedCache.SetStringAsync(
@@ -235,10 +235,10 @@ namespace Nop.Core.Caching
             
             try
             {
+                _ongoing.TryAdd(key.Key, lazy);
                 // await the lazy task in order to force value creation instead of directly setting data
                 // this way, other cache manager instances can access it while it is being set
                 SetLocal(key.Key, await lazy.Value);
-                _ongoing.TryAdd(key.Key, lazy);
                 await _distributedCache.SetStringAsync(key.Key, JsonConvert.SerializeObject(data), PrepareEntryOptions(key));
             }
             finally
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi @rickardvh. Thank you for your help. 

Historically, we didn't store the null value in the cache, since such a value usually indicates that there was some problem while reading the data, and the data needs to be re-read the next time it is accessed. If we abandon this idea now, we will have to carefully check everything outside of the CacheManager, which will lead to a large number of unidentified problems.

rickardvh:
> Hi @rickardvh. Thank you for your help.
> 
> Historically, we didn't store the null value in the cache, since such a value usually indicates that there was some problem while reading the data, and the data needs to be re-read the next time it is accessed. If we abandon this idea now, we will have to carefully check everything outside of the CacheManager, which will lead to a large number of unidentified problems.

I see. It makes sense, but if something *actually* went wrong, wouldn't throwing an exception be more "correct"? After all, a null might just indicate that the value is missing for a certain cache key, and finding that out could conceivably take some time. There are workarounds for this in places where a missing string is saved as the empty string instead of null, but we can't do that with most other types.

skoshelev:
Yes, of course, the exception is an excellent solution, however, we will again rest in the history of evolution of approaches to programming. Now the trend is such that it is better to raise the exception as early as possible. However, there were times when exceptions were considered a bad pattern and NULL returned in critical situations. To the fact that now it will take a lot of effort to make sure that the preservation of NULL in the cache will not lead to artifacts in work. I believe that such work needs to be done, but I consider labor costs unacceptable now

rickardvh:
> Yes, of course, the exception is an excellent solution, however, we will again rest in the history of evolution of approaches to programming. Now the trend is such that it is better to raise the exception as early as possible. However, there were times when exceptions were considered a bad pattern and NULL returned in critical situations. To the fact that now it will take a lot of effort to make sure that the preservation of NULL in the cache will not lead to artifacts in work. I believe that such work needs to be done, but I consider labor costs unacceptable now

Ah yes, I thought that might be the case, I was taught the same thing back in university. If you could point me to a place where this pattern is used in the current codebase, I can have a look. As it is, we are actually caching exceptions as well (in the memory cache), but on the other hand - the only code that can do anything about it is the caller, and the caller needs to know if something goes wrong in order to retry or whatever.

PR #6619, opened 2023-02-28
Optimise and streamline EntityRepository and ProductService

rickardvh:
I removed some unnecessary eager evaluations and turned for loops into pipelines to save execution time and memory, and changed lists into sets or dictionaries to avoid $O(n)$ lookups and make the code more intuitive.

Best regards,
Rickard von Haugwitz
Majako [majako.net](https://www.majako.net/)

### DIFF
--- a/src/Libraries/Nop.Data/EntityRepository.cs
+++ b/src/Libraries/Nop.Data/EntityRepository.cs
@@ -115,6 +115,28 @@ namespace Nop.Data
             return query.OfType<ISoftDeletedEntity>().Where(entry => !entry.Deleted).OfType<TEntity>();
         }
 
+        /// <summary>
+        /// Transactionally deletes a list of entities
+        /// </summary>
+        /// <param name="entities">Entities to delete</param>
+        protected virtual async Task DeleteAsync(IList<TEntity> entities)
+        {
+            using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+            await _dataProvider.BulkDeleteEntitiesAsync(entities);
+            transaction.Complete();
+        }
+
+        /// <summary>
+        /// Soft-deletes <see cref="ISoftDeletedEntity"/> entities
+        /// </summary>
+        /// <param name="entities">Entities to delete</param>
+        protected virtual async Task DeleteAsync<T>(IList<T> entities) where T : ISoftDeletedEntity, TEntity
+        {
+            foreach (var entity in entities)
+                entity.Deleted = true;
+            await _dataProvider.UpdateEntitiesAsync(entities);
+        }
+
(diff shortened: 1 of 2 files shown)


src/Libraries/Nop.Data/EntityRepository.cs
```diff
@@ -602,21 +603,27 @@ public virtual async Task DeleteAsync(IList<TEntity> entities, bool publishEvent
             if (entities == null)
                 throw new ArgumentNullException(nameof(entities));
 
-            if (entities.OfType<ISoftDeletedEntity>().Any())
```
skoshelev, line 605: I think it's probably better to leave one check, since TEntity will always be of the same type and check this every time for an entity in a loop - redundant code.

src/Libraries/Nop.Services/Catalog/ProductService.cs
```diff
@@ -240,8 +240,6 @@ protected virtual async Task<string> GetStockMessageForAttributesAsync(Product p
             if (!product.DisplayStockAvailability)
                 return string.Empty;
 
-            string stockMessage;
```
skoshelev, line 243: I do not agree with such changes, it seems to me that the readability of the method will suffer greatly, instead of one exit point, we have as many as six.

src/Libraries/Nop.Data/EntityRepository.cs
skoshelev: Sorry, I did not see the second commit right away, the question is no longer relevant

src/Libraries/Nop.Services/Catalog/ProductService.cs
rickardvh: Fair enough - I personally don't agree, since the reader will have to scroll down to the end of the method and try to follow the rest of it to see that nothing more was done after all (doing exactly that was what made me change it), but that's an age-old debate and purely a matter of personal preference ;) Early returns seemed to be preferred elsewhere in the code, too. I'll change it back :)

src/Libraries/Nop.Data/EntityRepository.cs
rickardvh: Yeah, the original loop with double type checking fooled me there - type constraints are very nice in this kind of situation.

src/Libraries/Nop.Services/Catalog/ProductService.cs
skoshelev: Thank you very much for understanding. I know that this is a discussion topic and glad that we did not go to disputes :)

rickardvh: It's your code ;)

PR #6627, opened 2023-03-03
Buffer redis messages

rickardvh:
We discovered a potential performance bottleneck in the Redis-synced cache where the instances could be flooded with key events (each of which spawns a new thread on the receiving side) when a lot of keys were deleted at once. To address this, we added message buffering to send updates in batch once per a set interval (default 500 ms). Setting the interval to 0 restores the original behaviour, but since instantaneous propagation cannot be guaranteed (or even achieved) in the first place, this should never be desirable.

Best regards,
Rickard von Haugwitz


skoshelev:
Hi @rickardvh. Thank you very much for your work

Issue #6167, opened 2022-03-10
Add Reply-To for Order Notification Email


RomanovM:
nopCommerce version: 4.50

> When I receive an order, sometimes I want to reply to the customer directly from the order notification email.  Rather than copy and paste the customers email address from the body of the email, it would be easier to click Reply and the customers email address is already in the To field.

Source: https://www.nopcommerce.com/boards/topic/93657/add-reply-to-for-order-notification-email

We could add a new property to `MessageTemplate` whether to fill in `ReplyTo` with a customer (or admin) email if applicable.
We already have `ReplyTo` parameters in `SendNotification` method, so we should pass them when this new property is set for the particular message template.

ChrisClaude:
If this task has no assignee. I would love to be assigned to it.

AndreiMaz:
@ChrisClaude Please feel free to make to make a pull request

micolf:
i have the same problem

micolf:
@RomanovM do you have find a solution?

parshan2:
I have filled out `replyToEmailAddress` and `replyToName` parameter of `SendNotificationAsync` method for most admin email messages.
@skoshelev any suggestion!

skoshelev:
Hi, @parshan2. I looked at your changes, and have a couple of points:

- part of the code is duplicated from method to method, it is more logical to group it into a separate method
- Initially, the task included a condition that it would be possible to configure filling in the fields for a response within a specific template, since often responses are not expected for email templates

parshan2:
Hi, @skoshelev . do you agree to add a new property named **`AllowToReply`** or **`AllowDirectReply`** to message templates.
When enabled it will allow admin to reply to the messages that is generated based on customer activity.

PS: What should this option mean for the emails that are sent to the customers?

![image](https://user-images.githubusercontent.com/10446385/220038965-0207a127-f350-4b8a-a02d-55991276474b.png)

micolf:
i resolved thank you

Il giorno lun 20 feb 2023 alle ore 14:30 Sergey Koshelev <
***@***.***> ha scritto:

parshan2:
@skoshelev filling `ReplyTo` fields with mail account data makes no sense. currently customer can reply to default sender that is the same as the sender mail account data.
Let me know if you have a better argument for that in mind.

I propose following methods to get ReplyToEmail and ReplyToName for most store owner templates:

        /// <summary>
[...]

PR #6637, opened 2023-03-07
#6167 AllowDirectReply option added to message templates

parshan2:


skoshelev:
Hi, @parshan2. Thanks for the help. I made a few minor adjustments. From the main: sql scripts were redundant, there is no need for them. In these types of migrations, we now apply the NopSchemaMigration attribute

PR #6642, opened 2023-03-08
Remove ConcurrentTrie

rickardvh:
When running a site under heavy traffic, we noticed that memory usage sometimes kept growing at a steady rate until we ran out of memory. The problem was eventually discovered to be the trie used by the cache key manager - it relies on keys following a hierarchical prefix structure, but my guess is that some plugin or something was using random keys, which caused memory overhead to grow uncontrollably. Since we cannot stop anyone from setting arbitrary keys, it is better that we simply switch back to using ConcurrentDictionaries. RemoveByPrefix will be an $O(n)$ operation instead of in $O(1)$, but it's fast enough that it's worth not using up all the memory for.

My apologies for not discovering this sooner - I was vaguely aware of the potential but couldn't create any problems during local load tests (with third-party plugins turned off...).

Best regards,
Rickard von Haugwitz


skoshelev:
Hi, @rickardvh. I'm also trying to figure out how to eliminate the problem, but I don't fully understand its causes. However, let's assume that we remake the tree in such a way that it will be formed not by characters, but by string. All our keys are formed according to this type part1.part2.....partN. therefore we can split the key by "." and form children from these fragments. Then if random keys come, they will not greatly aggravate the situation. Do you think this approach will solve the memory exhaustion problem?

PR #6644, opened 2023-03-10
Added support for multiple result sets in a stored procedure

kucukemre:
Added support for multiple result sets in a stored procedure.

### DIFF
--- a/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
+++ b/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
@@ -496,6 +496,25 @@ namespace Nop.Data.DataProviders
         }
 
         /// <summary>
+        /// Executes command using System.Data.CommandType.StoredProcedure command type and returns a result containing multiple result sets.
+        /// Saves result values for output and reference parameters to corresponding LinqToDB.Data.DataParameter object.
+        /// </summary>
+        /// <typeparam name="T">Result set type</typeparam>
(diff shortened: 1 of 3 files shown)


RomanovM:
@kucukemre Thanks for the suggestion, but what is the purpose of these changes? How it can be used?

kucukemre:
@RomanovM Hi, this adds a feature already available in LinqToDB. This method is useful to get multiple result sets from stored procedure. This is important to support in some cases due to the inefficiency of making multiple queries (especially when ids need to be passed to subsequent calls).

**Attribute approach**

A stored procedure is capable of returning multiple results sets in a given order. Using attributes we can specify these indexes on a result object like so:

`public class MyResultObject
    {
        [ResultSetIndex(0)] 
        public IEnumerable<A> A { get; set; }

        [ResultSetIndex(1)] 
        public IEnumerable<B> B { get; set; }

        [ResultSetIndex(2)] 
        publicIEnumerable<C> C { get; set; }

        [ResultSetIndex(3)] 
        public IEnumerable<D> D { get; set; }
    }`

We can then execute a new extension method in code that will fetch results for this:

`var result = await _dataProvider.QueryProcMultipleAsync<MyResultObject>("my_stored_procedure_name", parameters.ToArray());`

Each property in the ProductFeedResult object marked with the ResultSetIndex attribute will have its contents serialized using the item type of the property.

RomanovM:
@kucukemre Thanks for the clarification. Looks interesting but too specific, we'll leave this for customization.

Issue #6648, opened 2023-03-12
Set ShipSeparatelyOneItemEach setting to false on install


RomanovM:
nopCommerce version: develop

> It defaults to `True`.  (Which is odd, considering that the Hint on the Product field "_Ship Separately_" says "_... But notice that if the order includes **several items of this product, all of them will be shipped in single box**._")

Source: https://www.nopcommerce.com/boards/topic/96461/why-shipping-price-is-multiplied#304582

Initial issue: #2286

PR #6652, opened 2023-03-16
Set ShipSeparatelyOneItemEach setting to false on install #6648

parshan2:


RomanovM:
Thanks for your help.

PR #6661, opened 2023-03-29
Optimise ConcurrentTrie

rickardvh:
Due to space-complexity problems with the previous ConcurrentTrie, and the time-complexity issues with using a simple ConcurrentDictionary, I have completely rewritten ConcurrentTrie to optimise for both time and memory.

- Instead of a naïve trie, ConcurrentTrie is now implemented as a radix tree (see [https://en.wikipedia.org/wiki/Radix_tree](https://en.wikipedia.org/wiki/Radix_tree)).
- Due to the significant memory overhead caused by having a ConcurrentDictionary in each node, children are stored in regular dictionaries and locking is done outside.
- Instead of keeping one lock per node, the lock striping pattern is used with a fixed number (eight times the number of processor cores) of shared locks.
- Values are stored and manipulated in a lockless manner using volatile fields and the Interlocked class.
- String comparisons are performed with ReadOnlySpan to avoid allocating and copying substrings.

I had preferred to avoid making this kind of changes, as it severely impacts the readability of the code, but I recommend starting from the tests and benchmarks to review the code.

We have been running the code in production for almost a week already, and memory and CPU usage are both more stable than ever (see response times comparison below), and we have not encountered any unexpected behaviour so far. Memory usage grows in $O(n)$ for $n$ keys even with random UUID keys, and insertion, deletion and removal by prefix are all in $O(k)$ for key length $k$.

response time comparison

Since we recognise that maintaining this sort of code can be difficult (and due to the lack of existing open-source implementations), we plan to set up a public repository to release it separately. We will then update the file here with a link to that repository for issue reporting and feature requests.

Best regards,
Rickard von Haugwitz

### DIFF
--- a/src/Libraries/Nop.Core/Infrastructure/ConcurrentTrie.cs
+++ b/src/Libraries/Nop.Core/Infrastructure/ConcurrentTrie.cs
@@ -8,9 +8,9 @@ namespace Nop.Core.Infrastructure
     public partial class ConcurrentTrie<TValue>
     {
         #region Fields
-        private volatile TrieNode _root = new();
-        private readonly StripedReaderWriterLock _locks = new();
-        private readonly ReaderWriterLockSlim _structureLock = new();
+        protected volatile TrieNode _root = new();
+        protected readonly StripedReaderWriterLock _locks = new();
+        protected readonly ReaderWriterLockSlim _structureLock = new();
 
         #endregion
 
@@ -23,7 +23,7 @@ namespace Nop.Core.Infrastructure
         {
         }
 
-        private ConcurrentTrie(TrieNode subtreeRoot)
+        protected ConcurrentTrie(TrieNode subtreeRoot)
         {
             _root.Children[subtreeRoot.Label[0]] = subtreeRoot;
         }
@@ -486,7 +486,7 @@ namespace Nop.Core.Infrastructure
 
         protected class TrieNode
         {
-            private class ValueWrapper
+            protected class ValueWrapper
             {
                 public readonly TValue Value;
 
@@ -497,11 +497,11 @@ namespace Nop.Core.Infrastructure
             }
 
             // used to avoid keeping a separate boolean flag that would use another byte per node
-            private static readonly ValueWrapper _deleted = new(default);
+            protected static readonly ValueWrapper _deleted = new(default);
 
             public readonly string Label;
             public readonly Dictionary<char, TrieNode> Children = new();
-            private volatile ValueWrapper _value;
+            protected volatile ValueWrapper _value;
 
             public bool IsDeleted => _value == _deleted;
             public bool HasValue => _value != null && !IsDeleted;
--- a/src/Libraries/Nop.Core/Infrastructure/ConcurrentTrie.cs
+++ b/src/Libraries/Nop.Core/Infrastructure/ConcurrentTrie.cs
@@ -1,90 +1,307 @@
-﻿using System.Collections.Concurrent;
+﻿using System.Runtime.CompilerServices;
 
 namespace Nop.Core.Infrastructure
 {
     /// <summary>
-    /// A thread-safe implementation of a trie, or prefix tree
+    /// A thread-safe implementation of a radix tree
     /// </summary>
     public partial class ConcurrentTrie<TValue>
     {
         #region Fields
-
-        protected readonly TrieNode _root;
-        protected readonly string _prefix;
+        private volatile TrieNode _root = new();
+        private readonly StripedReaderWriterLock _locks = new();
+        private readonly ReaderWriterLockSlim _structureLock = new();
 
         #endregion
 
         #region Ctor
 
-        public ConcurrentTrie() : this(new(), string.Empty)
+        /// <summary>
+        /// Initializes a new empty instance of <see cref="ConcurrentTrie{TValue}" />
+        /// </summary>
+        public ConcurrentTrie()
         {
         }
 
-        protected ConcurrentTrie(TrieNode root, string prefix)
+        private ConcurrentTrie(TrieNode subtreeRoot)
         {
-            _root = root;
-            _prefix = prefix;
+            _root.Children[subtreeRoot.Label[0]] = subtreeRoot;
         }
 
         #endregion
 
         #region Utilities
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        protected static int GetCommonPrefixLength(ReadOnlySpan<char> s1, ReadOnlySpan<char> s2)
+        {
+            var i = 0;
+            var minLength = Math.Min(s1.Length, s2.Length);
+            for (; i < minLength && s2[i] == s1[i]; i++)
+                ;
+            return i;
+        }
+
         /// <summary>
-        /// Adds a node if key does not already exist.
+        /// Gets a lock on the node's children
         /// </summary>
-        /// <param name="key">The key</param>
-        /// <returns>The <see cref="TrieNode"/> item</returns>
-        protected virtual TrieNode GetOrAddNode(string key)
+        /// <remarks>
+        /// May return the same lock for two different nodes, so the user needs to check to avoid lock recursion exceptions
+        /// </remarks>
+        protected virtual ReaderWriterLockSlim GetLock(TrieNode node)
+        {
+            return _locks.GetLock(node.Children);
+        }
+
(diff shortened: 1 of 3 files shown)


skoshelev:
Hi @rickardvh. I am studying your implementation and I want to say that you have done a great job, but I have doubts about the effectiveness of this implementation. Of course, it is much better than the previous one, even if the previous one is remade as I suggested into string instead of characters. However, it seems to me that using ConcurrentDictionary is still preferable. Perhaps I do not fully understand all the nuances. Now I'm relying on the tests you suggested and nothing else. In view of the complexity of your development, I decided to add the ability to safely switch to using the same ConcurrentDictionary. To do this, I extracted an interface from your class.

```cs
public interface INopConcurrentCollection<TValue>
{	
	bool TryGetValue(string key, out TValue value);
	void Add(string key, TValue value);
	void Clear();
	IEnumerable<KeyValuePair<string, TValue>> Search(string prefix);
	void Remove(string key);
	bool Prune(string prefix, out INopConcurrentCollection<TValue> subCollection);
	IEnumerable<string> Keys { get; }
}
```

Created its implementation with ConcurrentDictionary

```cs
public partial class NopConcurrentCollection<TValue> : INopConcurrentCollection<TValue>
{
	protected readonly ConcurrentDictionary<string, TValue> _dictionary;
	
	protected NopConcurrentCollection(IEnumerable<KeyValuePair<string, TValue>> subCollection)
	{
		_dictionary = new ConcurrentDictionary<string, TValue>(subCollection);
	}

	public NopConcurrentCollection()
	{
		_dictionary = new ConcurrentDictionary<string, TValue>();
	}

	public bool TryGetValue(string key, out TValue value)
	{
		return _dictionary.TryGetValue(key, out value);
	}

	public void Add(string key, TValue value)
	{
		_dictionary[key] = value;
	}

	public void Clear()
	{
		_dictionary.Clear();
	}

	public IEnumerable<KeyValuePair<string, TValue>> Search(string prefix)
	{
		return _dictionary.Keys.Where(k => k.StartsWith(prefix))
			.Select(key => new KeyValuePair<string, TValue>(key, _dictionary[key]));
	}

	public void Remove(string key)
	{
		_dictionary.Remove(key, out _);
	}

	public bool Prune(string prefix, out INopConcurrentCollection<TValue> subCollection)
	{
		subCollection = new NopConcurrentCollection<TValue>(Search(prefix));

		try
		{
			foreach (var key in subCollection.Keys)
				Remove(key);
		}
		catch
		{
			return false;
		}

		return subCollection.Keys.Any();
	}

	public IEnumerable<string> Keys => _dictionary.Keys;
}
```
And conducted comparative tests, below are the results of one of them.

image
image

image
image

image
image

Below I have given my adaptation of your "tests" so that you can compare the results

```cs
[TestFixture]
public class ConcurrentTrieTests
{
	private void Profile(Action action)
	{
		var sw = new Stopwatch();
		var memory = GC.GetTotalMemory(true) / 1024.0 / 1024.0;
		sw.Start();
		
		action.Invoke();

		sw.Stop();
		var delta = GC.GetTotalMemory(true) / 1024.0 / 1024.0 - memory;
		Console.WriteLine("Elapsed teme: {0:F}s", sw.ElapsedMilliseconds / 1000.0);
		Console.WriteLine("Memory usege: {0:F}Mb", delta);
	}

	[Test]
	[TestCase(typeof(NopConcurrentCollection<byte>))]
	[TestCase(typeof(ConcurrentTrie<byte>))]
	//[Ignore("Not a test, used for profiling.")]
	public void Profile(Type oType)
	{
		var sut = Activator.CreateInstance(oType) as INopConcurrentCollection<byte>;
		sut.Should().NotBeNull();

		Profile(() =>
		{
			for (var i = 0; i < 1000000; i++)
				sut.Add(Guid.NewGuid().ToString(), 0);
		});
	}

	[Test]
	[TestCase(typeof(NopConcurrentCollection<int>))]
	[TestCase(typeof(ConcurrentTrie<int>))]
	//[Ignore("Concurrency tests are inherently flaky, and may give false positives. Enable manually when needed.")]
	public void DoesNotBreakDuringParallelAddRemove(Type oType)
	{
		var sut = Activator.CreateInstance(oType) as INopConcurrentCollection<int>;
		sut.Should().NotBeNull();

		Profile(() =>
		{
			Parallel.For(0, 1000, new ParallelOptions { MaxDegreeOfParallelism = 8 }, j =>
			{
				for (var i = 0; i < 1000; i++)
				{
					var s = $"{i}-{j}";
					sut.Add(s, i);
					sut.TryGetValue(s, out var value).Should().BeTrue();
					value.Should().Be(i);
					sut.Remove(s);
					sut.TryGetValue(s, out _).Should().BeFalse();
				}
			});
		});
		
		sut.Keys.Count().Should().Be(0);
	}

	[Test]
	[TestCase(typeof(NopConcurrentCollection<int>))]
	[TestCase(typeof(ConcurrentTrie<int>))]
	//[Ignore("Concurrency tests are inherently flaky, and may give false positives. Enable manually when needed.")]
	public void DoesNotBreakDuringParallelAddPrune(Type oType)
	{
		var sut = Activator.CreateInstance(oType) as INopConcurrentCollection<int>;
		sut.Should().NotBeNull();

		Profile(() =>
		{
			Parallel.For(0, 1000, new ParallelOptions { MaxDegreeOfParallelism = 8 }, j =>
			{
				for (var i = 0; i < 1000; i++)
				{
					var s = $"{j}-{i}";
					sut.Add(s, i);
				}
				sut.Prune($"{j}-", out var st);
				st.Keys.Count().Should().Be(1000);
			});
		});
		
		sut.Keys.Count().Should().Be(0);
	}
}
```

Perhaps these tests do not reflect those moments in which the efficiency of the tree will bring the main benefit in performance, or I made a mistake somewhere else

rickardvh:
@skoshelev oh, that's because I didn't write benchmarks for pruning on a full-size collection; that's where a *lot* of CPU time is spent when using a ConcurrentDictionary. I will make one for you so you can see the difference.

rickardvh:
@skoshelev Try this, for example:

``` cs
    [Test]
    [TestCase(typeof(NopConcurrentCollection<byte>))]
    [TestCase(typeof(ConcurrentTrie<byte>))]
    //[Ignore("Not a test, used for profiling.")]
    public void ProfilePrune(Type oType)
    {
        var sut = Activator.CreateInstance(oType) as INopConcurrentCollection<byte>;
        sut.Should().NotBeNull();
        // insert
        for (var i = 0; i < 100000; i++)
            sut.Add(Guid.NewGuid().ToString(), 0);

        Profile(() =>
        {
            // remove by prefix
            for (var i = 0; i < 100; i++)
                sut.Prune(Guid.NewGuid().ToString()[..5], out _);
        });
    }
```
It gives the following results on my computer:
```
ConcurrentTrieTests.ProfilePrune(Nop.Core.Infrastructure.NopConcurrentCollection`1[System.Byte]):
    Outcome: Passed
    Standard Output Messages:
    Elapsed time: 10,43s
Memory usage: 0,17Mb

    
ConcurrentTrieTests.ProfilePrune(Nop.Core.Infrastructure.ConcurrentTrie`1[System.Byte]):
    Outcome: Passed
    Standard Output Messages:
    Elapsed time: 0,00s
Memory usage: -0,00Mb
```
And that's with only 100k keys and 100 removals by prefix - during heavy traffic when each "add to cart" triggers multiple remove by prefix, the result can be devastating on a large store, as you can see in the graph I attached above.

skoshelev:
@rickardvh Great, now I understand. I will make some more adjustments to your code, I hope you will take them as a basis for your repository. (there are mostly cosmetics and unification)

rickardvh:
@skoshelev thanks, that sounds good! I will make more rigorous tests for that using property-based testing, and implement some .NET collection interfaces to make it more generally useful.

There is one minor issue I know about - I wouldn't call it a bug since it doesn't interfere with any functionality, but after running DoesNotBreakDuringParallelAddRemove under maximum concurrency, there are some empty nodes left in the tree. They don't use much memory and will be reused when new keys are inserted anyway, so compared to using a ConcurrentDictionary, which only ever extends the underlying array but never trims it, it's still much less of a problem. Prune or Clear don't have the same issue, either.

skoshelev:
Merged by https://github.com/nopSolutions/nopCommerce/commit/dc75fd71420b3239e10929394f96b26df11c0cbb

@rickardvh I think that a few empty keys that do not affect the algorithm of work are a small price to pay for all the benefits. Once again, many thanks for your contribution, I hope I did not break anything with my refactoring.

Issue #6669, opened 2023-04-12
Translation for Admin.Catalog.Products.SpecificationAttributes.NameFormat not available in default installation


brothers28:
nopCommerce version: 4.60.1

Steps to reproduce the problem: 

1. Install nopCommerce
2. Create a new product specification attribute group
3. Create a new product specification attribute within that group
4. Create or edit a product and add the new specification attribute --> it just shows "Admin.Catalog.Products.SpecificationAttributes.NameFormat" because the translation is missing

skoshelev:
Closed #6669

markusschnepf:
@skoshelev as I see in your commit [aecfa8fd699f0733331249b72f6e2a03bd922128](https://github.com/parshan2/nopCommerce/commit/aecfa8fd699f0733331249b72f6e2a03bd922128)
you added the ressource to the default installation but you did not add a migration/update function.

I updated from 4.50.x to 4.60.3 and we are stilling missing the ressource.
Yes, NOW I know that I can just add the ressource but I needed some time to debug until I found out, why grouped specification attributes do not show up when I want to associate them to products :D 

could you maybe fix for the new release 4.70?

Thanks a lot!

skoshelev:
Hi @markusschnepf. I suppose that this problem was fixed by this commit https://github.com/nopSolutions/nopCommerce/commit/8dac154fdc55bfe1df93469a07191bdaaffa2ba3. and version 4.60.3 should have already included it

PR #6672, opened 2023-04-12
#6669 added missing localiztion resource  for specification attribute…

parshan2:
this localization resource only added in UpgradeTo440


skoshelev:
Hi @parshan2. Thank you for your help, I closed the ticket by [this commit](https://github.com/nopSolutions/nopCommerce/commit/8dac154fdc55bfe1df93469a07191bdaaffa2ba3) There you can find a number of differences as your fix is not enough

PR #6675, opened 2023-04-14
Fix array indexing in StripedReaderWriterLock

rickardvh:
@skoshelev since I used a bit mask to index the locks array, we can't just allow any number of locks - only powers of 2 will work for that. It probably doesn't matter here since processor cores always (I think) come in powers of 2 and we have hardcoded the multiple to 8, but it's still wrong. If you insist on allowing arbitrary values of nLocks, we should remove _mask and replace `& _mask` with `% _locks.Length`, as I have done here. Modulus is a slightly slower operation than masking, and since we use powers of 2 anyway I thought we might as well make use of it, but to be fair the time saved is minimal (on the order of a ms or so per million operations) and is overshadowed by basically everything else we do.


skoshelev:
Hi @rickardvh. Thanks for the clarifications. Indeed, it seems to me that it is better to use this approach rather than bitwise operations, especially since the performance losses are minimal even at high loads.

PR #6685, opened 2023-04-28
Add new edit type - select

dbelouslv:
I needed to customize a new logic in the admin panel. I added a new column I can edit after the' Edit' button click. It was the type of picture. But when I edited I could change only as string, but I needed to choose a property from select. Thought that it can be useful for other people.

How it looks:

How to use it (Example):


jancusra:
this is really necessary! ... we also have your own version, which is working... would be very nice to see in core!

dbelouslv:
I updated the branch and resolved conflicts.

jancusra:
Hi, thanks, looks good! Is there some chance it will be merged by someone ?

PR #6687, opened 2023-05-01
Inherited from byte to take up less memory

dotnettashkent:


skoshelev:
Hi @dotnettashkent. Why did you choose these three enums out of 80. Have you considered any real resource savings when using byte as the base type? 

I don't see any practical value in changing the default int to byte based on the theoretical size alone, but if you have real measurements, I'll gladly reconsider my position

skoshelev:
@dotnettashkent Do you have specific values for how much memory will be saved by replacing at least these three enumerations? You propose to make not the most obvious changes to the code, which will impose a number of restrictions on us and cause changes in the structure of the database (since we store these values in the database for many entities), I would like to understand how many resources will be freed up.

If you do not have real measurements, then let's dream up. Let's imagine that at the same time we keep 1 million enum values in memory (I think this is a lot for a real site). The size difference between bytes and ints will be 24 bits. With simple calculations, we get a savings of 2.86 MB. It seems to me that this is not quite commensurate with the labor costs for such changes, what do you think?

RomanovM:
@dotnettashkent Thanks for the suggestion, but we don't plan to implement this.

PR #6695, opened 2023-05-05
Fix for discount amount calculation

kucukemre:
The discount amount is calculated incorrectly when there is more than one different tax rate.


RomanovM:
@kucukemre You are right. Thanks for your help.

PR #6708, opened 2023-05-19
Fix product manufacturers cache

kucukemre:
When a manufacturer is edited or deleted, the ProductManufacturersByProductCacheKey is not deleted from the cache. This fixes this error.


kucukemre:
Hi @skoshelev. The showHidden part of the ManufacturerService.GetProductManufacturersByProductIdAsync method will not work properly. When I select publish checkbox and edit the manufacturer or vice versa, the result of this method will always be same from cache.

skoshelev:
@kucukemre Okay, that makes sense. Thanks again for your help.

PR #6709, opened 2023-05-19
Update CheckoutController.cs

roberthej:
We need to save selected payment method before calculate Total of Cart

### DIFF
--- a/src/Presentation/Nop.Web/Controllers/CheckoutController.cs
+++ b/src/Presentation/Nop.Web/Controllers/CheckoutController.cs
@@ -1882,6 +1882,10 @@ namespace Nop.Web.Controllers
                         NopCustomerDefaults.UseRewardPointsDuringCheckoutAttribute, model.UseRewardPoints,
                         store.Id);
                 }
+                
+                //save selected payment method befor calculate total cart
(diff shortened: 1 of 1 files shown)


roberthej:
We need to save selected payment method before calculate Total of Cart

roberthej:
It will affect when the "_orderProcessingService.IsPaymentWorkflowRequiredAsync()" method calls the tax calculation "activeTaxProvider.GetTaxTotalAsync()" which is in the plugin "FixedOrByCountryStateZip" and cannot determine the additional fee which is retrieved using PaymentMethodSystemName, so it should be saved first.

skoshelev:
Hi @roberthej. Okay, I understand what you mean. But I think that everything is just the opposite, you do not need to save this parameter before calling IsPaymentWorkflowRequiredAsync. More than that, we need to fix the IsPaymentWorkflowRequiredAsync method so that it calls the GetShoppingCartTotalAsync method with the parameter usePaymentMethodAdditionalFee = false. Since if payment is not required, then there is no need to take into account the additional fee for using the payment method

roberthej:
Tnx @skoshelev

Issue #5950, opened 2021-11-15
Order page (admin area) cannot handle POST if you have a lot of products (data transferred)


sk5202:
nopCommerce version:  4.4.4

Steps to reproduce the problem: 

1) Add 160 or more than Unique product in the cart.

2) Placed one order with 160 product.

3) Go to /admin/Order/List and open your order which you placed

4) Change order status It will gives 400 bad request apart from this if you change order items field than also you will get same 400 error.

**Note : You have more than 150 unique product in cart in my case I've 160 products in the cart**  

Problem is if form have a lot data than it will gives 400 error that is my assumption.

skoshelev:
Hi @sk5202. We analyzed the problem, indeed, if there are so many items in the order, then there is a problem with the return of error 400. This is due to the limitations imposed on the form by the platform itself. We believe that this is part of the functioning of the security mechanism and have made a decision not to interfere with its work. If you really need to handle such large orders, then we recommend that you look towards the RequestFormLimits attribute or you can change the OrderModel and replace IList with a IPagedList

Closed #5950

sk5202:
@skoshelev  I've tried both way
1. RequestFormLimits attribute not working if i add in particular action or globally in web config
2. if i change IList to Ipagedlist than not able to create object.

Do you have any other idea ?

skoshelev:
@sk5202 Replacing IList with IPageList is not so simple, you need to redo the logic for creating the model, and the presentation layer will most likely have to be changed.

And for the RequestFormLimits attribute to work correctly, you also need the IgnoreAntiforgeryToken attribute without it may not work

sk5202:
@skoshelev  i've tried RequestFormLimits above action and also add IgnoreAntiforgeryToken  but won't work.

Is there anything i mistake i did ?

please check below snap..

Your help would be appreciated..!!

![image](https://user-images.githubusercontent.com/41533502/142242877-c3556fd6-2a1f-4d98-b051-9495e91422c8.png)


![image](https://user-images.githubusercontent.com/41533502/142242989-d1924ad5-6264-4ba0-a220-753818163d7d.png)

skoshelev:
Hi, @sk5202. Doesn't work, is it in the sense that it keeps showing error 400?

esusavila:
Hi.
This problem "Error 400. Bad request" also occurs in version 4.40
My NopCommerce client has a lot of orders with more than 160 items, and he can't change the status either.
How to solve?

PR #6725, opened 2023-06-06
#6723 fix for large forms causing 400 Bad Request

### LINKED ISSUE #6723: 400 bad request when trying to cancel/edit large order
nopCommerce version: 4.60.3

The order that I'm working with has 432 items.

When i try to cancel the order with the `cancel` button, or perform a manual status change, the change fails and the user receives a blank page. Dev tools and nop log show 400 bad request. 

Since the edit page is one big form that posts all input values regardless of what action is being taken on the order, and each order item has 6 attributes being posted, this results in 2,592 form fields being posted + approximately 25 other fields. So, the edit form for this order is submitting over 2,600 fields. This exceeds the default value of 1024 for the `RequestFormLimits` action attribute. I ran into a similar problem when developing a plugin for wholesale ordering.

The best fix, in my opinion would involve limiting the form fields to only submit the values to what is necessary for what is being changed.

The quick, less secure/robust fix would be adding `[DisableRequestSizeLimit]`, `[RequestFormLimits(ValueCountLimit = int.MaxValue, KeyLengthLimit = int.MaxValue)]` attributes to the `CancelOrder` action and other edit related actions which could suffer from this same issue.

I can certainly apply the attribute overrides to my local build to get it working, but I would like to see a more robust upstream fix for this.

danFbach:
This fixes #6723 and probably also related #5950.

As mentioned in #6723, I believe this should be reviewed for its potential for abuse - i.e., the submission of massive forms.

Also up for debate is the placement of the new code, although this seems to be the best place in my opinion.

### DIFF
--- a/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ServiceCollectionExtensions.cs
@@ -6,6 +6,7 @@ using FluentValidation.AspNetCore;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
@@ -92,6 +93,12 @@ namespace Nop.Web.Framework.Infrastructure.Extensions
             var pluginConfig = new PluginConfig();
             builder.Configuration.GetSection(nameof(PluginConfig)).Bind(pluginConfig, options => options.BindNonPublicProperties = true);
             mvcCoreBuilder.PartManager.InitializePlugins(pluginConfig);
+
+            //allow large form submissions
+            services.Configure<FormOptions>(options =>
+            {
+                options.ValueCountLimit = int.MaxValue;
+            });
 
             //create engine and configure service provider
             var engine = EngineContext.Create();


danFbach:
I have tried using a narrower approach, applying the attribute `[RequestFormLimits(ValueCountLimit = int.MaxValue)]` to both the OrderController as a whole, and to the CancelOrder action specifically.

When applied to the controller, the request appears to succeed and redirect to you the Edit page, but actually just executes the Edit action. It appears this fails because every sub-request (to components, widgets, etc) also exceeds the form limits, so presumably the FormValueRequired 'cancelorder' is lost...? 

When applied only to the CancelOrder action, you still receive the 400 bad request because when it is redirected to Edit action, the ValueCountLimit is exceeded for Edit action.

The *best* fix for this, would be to change this page so that you never submit so many form values, but require a rewrite of the views, actions and supporting js.

skoshelev:
Hi @danFbach. Thank you for your help. We are also investigating solutions to this problem, at the moment our best option is to add the RequestFormLimits attribute to the BaseAdminController so as not to change the global setting. On the one hand, this should solve all such problems in the administration area, and on the other hand, it will not allow attacking the public part of the site. But we are still looking for a more flexible or configurable solution.

danFbach:
@skoshelev I tried applying the RequestFormLimits attribute to the BaseAdminController and it had the same effect as when applied to the order controller. Something gets messed up with the routing and the FormValueRequired is missed - even though when I watch Request.Form while debugging, "cancelorder" is there.

skoshelev:
Hi @danFbach. Yes, that's right, one attribute is not enough and I couldn't find the source of the problem. However, we found a solution that works, see [this commit](https://github.com/nopSolutions/nopCommerce/commit/a481be440a8505a01a0ae0f379fb2362782cf130). Thanks again for your help

Issue #6734, opened 2023-06-14
Issue with shopping cart estimate shipping popup


duc-tv:
nopCommerce version: latest from develop

Steps to reproduce the problem: 
1. add any product that require a shipping method to the cart
2. go to the cart and open the estimate shipping popup, fill in the necessary information.
3. no shipping options will be displayed

-> the reason is due to a syntax error [here](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Presentation/Nop.Web/Views/Shared/Components/ShoppingCartEstimateShipping/Default.cshtml#L41): `return '@Html.Raw(Url.RouteUrl("EstimateShippingt"))?' + params;`

PR #6735, opened 2023-06-14
Issue 6734, fix the syntax error in shopping cart estimate shipping popup component

duc-tv:
this is a fix for #6734


RomanovM:
@duc-tv Thanks for your help.

Issue #231, opened 2016-01-15
Implement "3D Bin Packing" algorithm for shipping


mariannk:
All of the nopCommerce shipping algorithms have the same problem. They do not work for real world cases with multiple items of various sizes being added to the cart. 

To produce accurate box selection (and shipping estimates) will require implementing "3D Bin Packing" simulation code. Given a set of items of various sizes, and a set of various sized boxes, what is the how many boxes, and of what size, are required? Shipping estimates would then be much more accurate, as the box selection would simulate the actual packing of boxes. 

Useful links:
http://en.wikipedia.org/wiki/Packing_problem
http://en.wikipedia.org/wiki/Bin_packing_problem
http://www.nopcommerce.com/boards/t/9414/bugs-in-shipping-rate-calculation-important.aspx?p=2
http://www.nopcommerce.com/boards/t/17719/bug-in-shipping-size-calculation.aspx
etc (search forums)

Have a look at the following contributions:
https://hg01.codeplex.com/forks/grene/shippingimprovements
https://hg.codeplex.com/forks/grene/packedboxes

[AM] The upcoming version 2.70 already has cube root method of calculating dimensions. That's enough for now. Anyway it's much more accurate than the previous method.

keremdemirer:
I created a repo does the 3d packing stuff in javascript. It needs the box sizes as input to do calculations. I'm posting this because its hard find this algorithm in a c# similar language, can be useful.

https://github.com/keremdemirer/3dbinpackingjs

sumo300:
There are a few services now that offer APIs for this type of packing solution.  Sounds like a plug-in is needed!

https://www.3dbinpacking.com/products/fillContainer

AndreiMaz:
almost not possible to implement

PR #6772, opened 2023-07-17
Correct the dimensions of shipment items

holydk:
Hi to best team

### DIFF
--- a/src/Libraries/Nop.Services/Shipping/IShippingService.cs
+++ b/src/Libraries/Nop.Services/Shipping/IShippingService.cs
@@ -270,6 +270,15 @@ namespace Nop.Services.Shipping
         /// </returns>
         Task<decimal> GetAdditionalShippingChargeAsync(ShoppingCartItem shoppingCartItem);
 
+        /// <summary>
+        /// Gets dimensions of product with optional <paramref name="attributesXml"/>
+        /// </summary>
+        /// <param name="product">The product</param>
+        /// <param name="attributesXml">The product attributes contains values with type <see cref="AttributeValueType.AssociatedToProduct"/> in XML format (optional)</param>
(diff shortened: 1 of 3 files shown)


AndreiMaz:
Also see https://github.com/nopSolutions/nopCommerce/issues/231 . Anyway, it's not possible to implement it properly and we're leaving the current implementation to keep it simple

holydk:
@DmitriyKulagin I agree with you, but my solution does not change the dimensions calculation method (just refactoring), everything remains as it was. It should probably be left to plugins, besides most plugins have custom methods for calculating package sizes ([UPS](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Plugins/Nop.Plugin.Shipping.UPS/Domain/PackingType.cs), [FedEx](https://github.com/nopSolutions/FedEx-plugin-for-nopcommerce/blob/nopcommerce-4.60/Nop.Plugin.Shipping.Fedex/PackingType.cs)).

My solution includes several uses:
1. Added a new public method **GetProductDimensionsAsync** for calculating product sizes. This will make it easier to get product sizes in plugins. Currently the size of a single product [can be obtained like this](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Plugins/Nop.Plugin.Shipping.UPS/Services/UPSService.cs#L739C87-L739C87)
2. Displays product sizes including associated products on the shipment creation page. The fact is that on this page I have a widget for creating an order in the delivery service. Parcel sizes can be entered manually, and products with associated products display incorrect sizes, because of this, the manager may make a mistake in the calculation

However, the second point still remains unresolved.
@DmitriyKulagin @AndreiMaz So how do you like the idea of adding a nested table for the product containing the associated products with dimensions?
shipment-items-with-accosiated-products

PR #6786, opened 2023-07-25
RedisConnectionWrapper.FlushDatabaseAsync: do not execute "FLUSHDB" on replicas

kralikba:
We are running NopCommerce with a Redis Sentinel cluster.
We have been getting the following exception:

```
StackExchange.Redis.RedisCommandException: Command cannot be issued to a replica: FLUSHDB
```

The problem was that RedisConnectionWrapper.FlushDatabaseAsync was calling FlushDatabaseAsync on the replicas as well. 
I believe that the attached fix fixes the issue.


skoshelev:
Hi @kralikba. Thank you for your help.

PR #6800, opened 2023-07-28
gitignore fix for rider, added ".idea" folder to ignore list

Maruf61:
while using rider it adds ide related files under ".idea" which should be ignored, so i added it to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 
 # Build results
 
+.idea/
 .vs/
 [Dd]ebug/
 [Rr]elease/


RomanovM:
We don't use rider.

Maruf61:
you may not be using rider but someone else does, this is very simple PR that does not negatively affect anyone at all. it simply helps people that uses rider to easily contribute to project.

skoshelev:
Hi @Maruf61. Of course, this is a simple change and does not have any negative impact, but then we have to add specific folders for other development tools, of which there are quite a few. I agree with the @RomanovM. If you work in rider, you can add the required line to the global gitignor for your computer and not worry that extra files will end up in PR

PR #6822, opened 2023-08-18
Fix typo in "standart" in ShoppingCartService

Manuel-Innovapps:
Should say "standard" (with a d at the end)


AndreiMaz:
@Manuel-Innovapps Thanks a lot

Issue #6826, opened 2023-08-22
SaveOrderDetailsAsync performance


ahmadkq:
nopCommerce version: 

Steps to reproduce the problem: 

1) Don't call UpdateOrderAsync twice.
Remove await _orderService.UpdateOrderAsync(order); line 778 and make it inside the if statment as following:

**if (details.RedeemedRewardPointsAmount <= decimal.Zero)
            {
                await _orderService.UpdateOrderAsync(order);
                return order;
            }**

2) Remove await _customerService.UpdateCustomerAsync(details.Customer); as it is not needed

skoshelev:
Hi @ahmadkq and @Dhiru1010. Thanks for your suggestion and research. In this case, it is not enough to look only at performance, this is an integral mechanism, we save all the details of the order, so that we can then calculate and save its number. Suppose an exception occurs in the AddRewardPointsHistoryEntryAsync method or in the GetResourceAsync method, what will we get as a result? Formed order, but without a number. I do not think that one additional call to the database is a lot for our peace of mind.

At the same time, I agree that calling the UpdateCustomerAsync method in this method is not needed

sinanopcommerce:
> Hi @ahmadkq and @Dhiru1010. Thanks for your suggestion and research. In this case, it is not enough to look only at performance, this is an integral mechanism, we save all the details of the order, so that we can then calculate and save its number. Suppose an exception occurs in the AddRewardPointsHistoryEntryAsync method or in the GetResourceAsync method, what will we get as a result? Formed order, but without a number. I do not think that one additional call to the database is a lot for our peace of mind.
> 
> At the same time, I agree that calling the UpdateCustomerAsync method in this method is not needed

Is it possible to use transaction, commit, and rollback for order and order item creation? Because for the scenario I got order has been created without the product/order item. Also in some cases, we found a quantity mismatch. Suppose a product has 5 quantities and it orders 7 quantities. This scenario happens mostly at campaign time.

skoshelev:
Hi @sinanopcommerce. 
Theoretically, it is possible to create an order in a transaction, but this will require a significant change in logic. I don't think such changes are advisable at all. The problems that you described are not typical of our code and usually arise due to customization or incorrect operation of third-party plugins; in any case, each such problem must be dealt with individually

skoshelev:
Closed #6826

sinanopcommerce:
"The problems that you described are not typical of our code."
[...]

PR #6834, opened 2023-08-27
Issue #6826 SaveOrderDetailsAsync performance code modification 

Dhiru1010:
this is the fix for #6826 
1. Remove await _orderService.UpdateOrderAsync(order); line 778 and make it inside the if statment as following:
```
if (details.RedeemedRewardPointsAmount <= decimal.Zero)
{
await _orderService.UpdateOrderAsync(order);
return order;
}
```
3. Remove await _customerService.UpdateCustomerAsync(details.Customer); as it is not needed


RomanovM:
Thanks for your help, but probably there is a need for more changes, see [this comment](https://github.com/nopSolutions/nopCommerce/issues/6826#issuecomment-1695192001).

Issue #6847, opened 2023-09-08
Is there a base path change?


Alerinos:
Needs to put the system on a domain but as a path.
For example:
``domain.com/shop``
Unfortunately nginx settings:
```
	location /shop {
		proxy_pass http://192.168.1.36:5000;
		proxy_set_header Upgrade $http_upgrade;
		proxy_set_header Connection $connection_upgrade;
		proxy_cache off;
		proxy_http_version 1.1;
		proxy_buffering off;
		proxy_read_timeout 100s;
		proxy_set_header Host $host;
		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
		proxy_set_header X-Forwarded-Proto $scheme;
    }
```
Do not work because I still have a redirect to "no-found" or "install".
Is there any minimally invasive way (like config) to change pathbase?

RomanovM:
@Alerinos  Please post your questions on [our forums](https://www.nopcommerce.com/boards). We use GitHub only as an issue tracker.

Alerinos:
@RomanovM 
Thank you for contacting me, I believe it is a technical problem. I did not find such a ready-made solution and did it.
You should add ``<base href="~/path"/>` in the head and then ``application.UsePathBase("/path");`.
I think this should be the basic functionality that we can set in appsettings. If you agree then I can prepare a PR.

PR #6852, opened 2023-09-09
Allows you to make your own path base store. For example, domian.com/…

Alerinos:
Allows you to make your own path base store. 
For example ``domian.com/shop``

In the nginx configuration, just indicate:
```
location / {
```
at
```
location / shop {
```

Issue: https://github.com/nopSolutions/nopCommerce/issues/6847

### DIFF
--- a/src/Libraries/Nop.Core/Configuration/HostingConfig.cs
+++ b/src/Libraries/Nop.Core/Configuration/HostingConfig.cs
@@ -28,6 +28,11 @@
         /// <summary>
         /// Gets or sets addresses of known networks to accept forwarded headers from
         /// </summary>
-        public string KnownNetworks { get; protected set; } = string.Empty;
+        public string KnownNetworks { get; protected set; } = string.Empty;        
+        
+        /// <summary>
+        /// Ability to set a custom path base, for example domain.com/path/
+        /// </summary>
+        public string UsePathBase { get; protected set; } = "/";
     }
 }
--- a/src/Libraries/Nop.Core/Infrastructure/NopEngine.cs
+++ b/src/Libraries/Nop.Core/Infrastructure/NopEngine.cs
@@ -4,6 +4,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Nop.Core.Configuration;
 using Nop.Core.Infrastructure.Mapper;
 
 namespace Nop.Core.Infrastructure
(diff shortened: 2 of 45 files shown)


Alerinos:
In ``Admin/Log/List`` it is hardwired to generate a link to "See" when you click on the logs. This needs to be found and corrected.
image
Probably poorly constructed routing or ``/`` at the beginning.

Alerinos:
I fixed all the "Edit" buttons, there are other elements that need to be improved. If I find something I will upload the changes.
I think I'm helping.

Alerinos:
I fixed the tutorial with the inclusion of pathBase.
JS is problematic, I wanted to do nicer but couldn't do it

RomanovM:
@Alerinos Thanks for this suggestion. I think this is related only to nginx configuration, so we'll leave it for customization.

Alerinos:
@RomanovM I tried, but unfortunately it doesn't work because routing in ASP.net still returns 404. Also standard in HTML is to use base in the head. Every ASP.net template has a base.
I think you should include this PR or provide a guide on how to do it via nginx.

Alerinos:
@RomanovM I still have not received the information, if it is possible I would like you to point me to the documentation.

According to Microsoft documentation, there must be PathBase as well as Base itself in Head. I have the impression that my PR has not even been analyzed.

@AndreiMaz Can you take a look at this PR? I believe I am correct as well as I have corrected some errors. 
https://www.w3schools.com/tags/tag_base.asp
https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-7.0

Alerinos:
I found another error, the path of this address should be corrected. It does not include PathBase because it is built rigidly.
```
Error 404. The requested page (/scheduletask/runtask) was not found
```
https://github.com/nopSolutions/nopCommerce/blob/0e8a771ec9255c566dd5000fd6ea8a73c317fd2b/src/Libraries/Nop.Services/ScheduleTasks/ScheduleTaskRunner.cs#L149
https://github.com/nopSolutions/nopCommerce/blob/0e8a771ec9255c566dd5000fd6ea8a73c317fd2b/src/Presentation/Nop.Web.Framework/WebWorkContext.cs#L220
https://github.com/nopSolutions/nopCommerce/blob/0e8a771ec9255c566dd5000fd6ea8a73c317fd2b/src/Libraries/Nop.Services/ScheduleTasks/TaskScheduler.cs#L64

Alerinos:
Another problem is the specification attributes.
It is probably JS, when clicked deletes the path base from the URL. 

https://gyazo.com/c60a398ee692e98526e33a1136ab0fe6

Issue #6853, opened 2023-09-12
Add three options for gender (as per German laws)


AndreiMaz:
nopCommerce version: 4.60.4

Steps to reproduce the problem: 

There's a critical legal requirement in Germany concerning the gender selection. As per German law, when offering a choice for gender, there must be three options available: Männlich (Male), Weiblich (Female), and Divers or 
Neutral (Non-Binary). Given that this third option is a legal necessity, it is imperative that it becomes a standard feature in nopCommerce to ensure compliance for all shops operating in Germany. This inclusion should be made standard for all nopCommerce implementations to facilitate ease of use for users and to avoid potential legal implications.

Here is an example - https://registrierung.web.de/

Let's implement it the following way:
1. There is the "Customer form fields" block on the "Customer settings" page in admin area
2. When "'Gender' enabled" is ticked, we should display a nested setting with the "'Neutral' option enabled" (disabled by default)
3. When this option is enabled, then we'll display this third option during gender selection
P.S The hint will be "Set if you need three gender options available - Male, female, neutral (as per German laws)"

rizwan3d:
Hey @AndreiMaz ,
I've made some progress on it; could you take a look?
Thanks.

DmitriyKulagin:
Closed #6853

PR #6866, opened 2023-10-03
issue-6853-Add-Neutral-Gender #6853

rizwan3d:
Resolved: #6853


DmitriyKulagin:
@rizwan3d Thank you for your assistance. We accept this request.

We have added some localization and settings changes for those who would like to update the version of nopCommerce rather than install it again.

Issue #6906, opened 2023-11-03
Created indexes on foreign key constraints


AndreiMaz:
Currently a task to delete customers could be very slow. We've found out that indexes to child tables could significatly increase performance of this tasks. For example, there should be the index of [CustomerId] index in the [CustomerPassword] table. Useful links - https://stackoverflow.com/questions/10901299/delete-statement-in-sql-is-very-slow

Also see https://learn.microsoft.com/en-us/sql/relational-databases/tables/primary-and-foreign-key-constraints?view=sql-server-ver16#indexes-on-foreign-key-constraints

ChrisClaude:
Is there a guide on how to add indexes on the project?

skoshelev:
Hi @ChrisClaude. I'm very sorry, but we don't have instructions for such changes. In general, adding an entry to the /Installation/Indexes.cs file will solve the issue only at the installation stage. To add indexes to an already running site, you must add a new migration.

skoshelev:
Closed #6906

PR #6911, opened 2023-11-08
#6906 Add indexes on foreign key constraints

ChrisClaude:
Add indexes on Customer Id foreign key constraints

### DIFF
--- a/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
+++ b/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
@@ -45,11 +45,11 @@ namespace Nop.Data.Migrations.UpgradeTo470
                 .OnColumn(nameof(ExternalAuthenticationRecord.CustomerId)).Ascending()
                 .WithOptions().NonClustered();
 
-            IfDatabase(databaseType).Create.Index("IX_PrivateMessage_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(PrivateMessage)))
+            IfDatabase(databaseType).Create.Index("IX_PrivateMessage_FromCustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(PrivateMessage)))
                 .OnColumn(nameof(PrivateMessage.FromCustomerId)).Ascending()
                 .WithOptions().NonClustered();
 
-            IfDatabase(databaseType).Create.Index("IX_PrivateMessage_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(PrivateMessage)))
+            IfDatabase(databaseType).Create.Index("IX_PrivateMessage_ToCustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(PrivateMessage)))
                 .OnColumn(nameof(PrivateMessage.ToCustomerId)).Ascending()
                 .WithOptions().NonClustered();
 
--- a/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
+++ b/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
@@ -29,11 +29,11 @@ namespace Nop.Data.Migrations.UpgradeTo470
                 .OnColumn(NameCompatibilityManager.GetColumnName(typeof(CustomerCustomerRoleMapping), nameof(CustomerCustomerRoleMapping.CustomerId))).Ascending()
                 .WithOptions().NonClustered();
 
-            IfDatabase(databaseType).Create.Index("IX_CustomerCustomerRoleMapping_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(ForumTopic)))
+            IfDatabase(databaseType).Create.Index("IX_ForumTopic_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(ForumTopic)))
                 .OnColumn(nameof(ForumTopic.CustomerId)).Ascending()
                 .WithOptions().NonClustered();
 
-            IfDatabase(databaseType).Create.Index("IX_CustomerCustomerRoleMapping_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(ForumPost)))
+            IfDatabase(databaseType).Create.Index("IX_ForumPost_CustomerId").OnTable(NameCompatibilityManager.GetTableName(typeof(ForumPost)))
                 .OnColumn(nameof(ForumPost.CustomerId)).Ascending()
                 .WithOptions().NonClustered();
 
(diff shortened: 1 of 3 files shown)


skoshelev:
Hi @ChrisClaude. As discussed in the ticket, indexes must be created on a separate migration but also we should add them to the installation file. I’m also not sure that adding indexes to fields with external links is worth it for all databases since it seems that in MySql and Postgres foreign keys are already indexed by default, and if so, there is no need for an additional index

ChrisClaude:
Hi @skoshelev I will add the migrations in a different file and the installation file.

My understanding of your last sentence is that we shouldn't add indexes on foreign keys because they automatically get one. Am I right? However, the issue mentions `For example, there should be the index of [CustomerId] index in the [CustomerPassword] table. ` [CustomerId] is a FK

skoshelev:
No, not really. Indexes for FK need to be added only to the MS SQL database, the other two do this automatically and it is not entirely clear what will happen if there are several indexes on such a field.

ChrisClaude:
@skoshelev I have updated the code to target MS SQL only. Please take a look.

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+		/// </summary>
+		public override void Up()
+		{
+			const string databaseType = "sqlserver";
+
+			IfDatabase(databaseType).Create.Index("IX_ActivityLog_CustomerId").OnTable(nameof(ActivityLog))
+				.OnColumn(nameof(ActivityLog.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_CustomerCustomerRoleMapping_CustomerId").OnTable(nameof(CustomerCustomerRoleMapping))
```
skoshelev, line 27: For this table and its CustomerId field, you must use the NameCompatibilityManager, since they differ in the database and the code

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+
+			IfDatabase(databaseType).Create.Index("IX_RewardPointsHistory_CustomerId").OnTable(nameof(RewardPointsHistory))
+				.OnColumn(nameof(RewardPointsHistory.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_NewsComment_CustomerId").OnTable(nameof(NewsComment))
+				.OnColumn(nameof(NewsComment.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_CustomerAddressMapping_CustomerId").OnTable(nameof(CustomerAddressMapping))
```
skoshelev, line 91: As with CustomerCustomerRoleMapping, you must use NameCompatibilityManager for both the table and the CustomerId field

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+
+			IfDatabase(databaseType).Create.Index("IX_CustomerPassword_CustomerId").OnTable(nameof(CustomerPassword))
+				.OnColumn(nameof(CustomerPassword.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_ExternalAuthenticationRecord_CustomerId").OnTable(nameof(ExternalAuthenticationRecord))
+				.OnColumn(nameof(ExternalAuthenticationRecord.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_PrivateMessage_From_CustomerId").OnTable(nameof(PrivateMessage))
```
skoshelev, line 47: As with CustomerCustomerRoleMapping, you must use NameCompatibilityManager for the table

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+				.OnColumn(nameof(PrivateMessage.FromCustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_PrivateMessage_To_CustomerId").OnTable(nameof(PrivateMessage))
+				.OnColumn(nameof(PrivateMessage.ToCustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_ForumSubscription_CustomerId").OnTable(nameof(ForumSubscription))
```
skoshelev, line 55: As with CustomerCustomerRoleMapping, you must use NameCompatibilityManager for the table

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+
+			IfDatabase(databaseType).Create.Index("IX_ActivityLog_CustomerId").OnTable(nameof(ActivityLog))
+				.OnColumn(nameof(ActivityLog.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_CustomerCustomerRoleMapping_CustomerId").OnTable(nameof(CustomerCustomerRoleMapping))
+				.OnColumn(nameof(CustomerCustomerRoleMapping.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_ForumTopic_CustomerId").OnTable(nameof(ForumTopic))
```
skoshelev, line 31: As with CustomerCustomerRoleMapping, you must use NameCompatibilityManager for the table

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -0,0 +1,104 @@
...
+				.OnColumn(nameof(ForumTopic.CustomerId)).Ascending()
+				.WithOptions().NonClustered();
+
+			IfDatabase(databaseType).Create.Index("IX_ForumPost_CustomerId").OnTable(nameof(ForumPost))
```
skoshelev, line 35: As with CustomerCustomerRoleMapping, you must use NameCompatibilityManager for the table

skoshelev:
Hi @ChrisClaude. Thanks, I looked at the code and left comments regarding the use of NameCompatibilityManager. You can find all the replacement rules for the main code in the [BaseNameCompatibility](https://github.com/nopSolutions/nopCommerce/blob/347adb86b82fe89d1a0c730db5ad51051849e2c9/src/Libraries/Nop.Data/Mapping/BaseNameCompatibility.cs) class

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: I have updated the code to use NameCompatibilityManager

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: NameCompatibilityManager

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: I have updated the code to use NameCompatibilityManager

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: I have updated the code to use NameCompatibilityManager

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: I have updated the code to use NameCompatibilityManager

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
ChrisClaude: I have updated the code to use NameCompatibilityManager

ChrisClaude:
@skoshelev Thanks for the comments. I have updated the code to use NameCompatibilityManager where required. Please review the code again.

skoshelev:
Hi @ChrisClaude. I looked at the changes, and almost everything is good :)
Only for all indexes in which you have used NameCompatibilityManager now one name "IX_CustomerCustomerRoleMapping_CustomerId" makes it impossible to start the migration

ChrisClaude:
@skoshelev I have updated the names of the indexes

skoshelev:
Hi @ChrisClaude. Looks like we did it. Many thanks for the help

PR #6920, opened 2023-11-14
Make ConcurrentTrie.Search not require exact hit

rickardvh:
`ConcurrentTrie.Search` previously required the argument key to exactly match an existing key in the tree. This didn't affect any code in the core codebase, but it was unintuitive behaviour. This fix changes the behaviour to conform with the `Prune` method, so that it returns all key-value pairs starting with the supplied prefix; see changes to `ConcurrentTrieTests.CanSearch`.

Best regards,

Rickard von Haugwitz
Majako [https://majako.net/](majako.net) / [https://majako.se/](majako.se)


rickardvh:
Bah, sorry for the extra commits - some git submodule changes sneaked in unseen.

skoshelev:
Hi @rickardvh. Thank you, the new behavior is indeed more expected and predictable

PR #6921, opened 2023-11-14
Use cached customer roles to avoid SQL joins

rickardvh:
Recently, we had been experiencing thread-pool starvation on the (MS)SQL server during periods of unusually high traffic, resulting in extremely long response times and even timeouts. These incidents were found to have had a strong correlation with spikes in CustomerRole select queries. As you can see below, during the month leading up to the deployment of this fix, this query was the most time-consuming database operation, being responsible for nearly a third of all query time; it has since been eliminated completely from the list.

image

Since all customer roles are already cached and no customer is likely to ever have more than a few roles, we could completely forego the expensive table join and limit ourselves to a single select query on CustomerRole on startup.

There is also no compelling reason to cache active and inactive roles separately - even if each customer were mapped to hundreds if not thousands of roles (which seems incredibly far-fetched), filtering them in memory wouldn't add much complexity.

The site has been exceptionally stable since we deployed this fix.

Best regards,

Rickard von Haugwitz

### DIFF
--- a/src/Libraries/Nop.Services/Customers/CustomerService.cs
+++ b/src/Libraries/Nop.Services/Customers/CustomerService.cs
@@ -1184,6 +1184,19 @@ namespace Nop.Services.Customers
         }
 
         /// <summary>
+        /// Gets a dictionary of all customer roles mapped by ID.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation and contains a dictionary of all customer roles mapped by ID.
+        /// </returns>
+        public virtual async Task<IDictionary<int, CustomerRole>> GetAllCustomerRolesDictionaryAsync()
+        {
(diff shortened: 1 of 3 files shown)


AndreiMaz:
@rickardvh Thanks a lot for your contribution! Minor suggestion: I think we can make the "GetAllCustomerRolesDictionaryAsync" method protected and remove it from from "ICustomerService"

skoshelev:
Hi @rickardvh. I made the edits that Andrei talked about and made a couple more minor changes. Many thanks for your help

rickardvh:
@AndreiMaz @skoshelev Great! I made it protected originally anyway, but thought that ID-keyed dictionaries are way more useful than lists for cached stuff, so perhaps others might want to use it, too. Clean and uncluttered interfaces are definitely more important, though.

PR #6923, opened 2023-11-15
Optimise caching in GetByIds

rickardvh:
EntityRepository.GetByIdsAsync has a couple of problems:

- Each new list of IDs (by set equality) will trigger a new database query, even if all entities are already cached individually or as a subset of another `.byids.` entry; GetById similarly cannot reuse previously cached entities from GetByIds, and there is no upper limit to how many combinations of entities can be stored in the cache
- The same entity may be cached multiple times under different keys, and each copy will be a new object rather than a reference to the same object in memory

When using a memory cache, it makes more sense to cache each entity individually to maximise reuse, while still batching queries for missing entities in GetByIds - since a memory cache is basically a big dictionary, it is extremely fast to check whether a key exists. On the other hand, if a (true) distributed cache is being used, it's more reasonable to use the existing implementation, as the overhead of asking the cache for individual entities would easily be larger than simply fetching everything from the database.

The difference this change made for us in memory usage can be seen in the charts below, of maximum memory usage across all instances of one app during September and from late October until now:

memory-sep

memory-oct-nov

Before the change, memory routinely kept climbing above 80 %, but afterwards it has rarely crossed 60 %, only briefly spiking during deployments.

### DIFF
--- a/src/Libraries/Nop.Data/EntityRepository.cs
+++ b/src/Libraries/Nop.Data/EntityRepository.cs
@@ -2,6 +2,7 @@
 using System.Transactions;
 using Nop.Core;
 using Nop.Core.Caching;
+using Nop.Core.Configuration;
 using Nop.Core.Domain.Common;
 using Nop.Core.Events;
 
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @rickardvh. Thanks for the analysis and good implementation

PR #6924, opened 2023-11-16
Improve upon LocalizedEntity and UrlRecord caching

rickardvh:
By precomputing only the values we are interested in and removing unnecessary nested cache calls, we can avoid double-caching localised entities and URL records without sacrificing speed. Since there may be a lot of both localised entities and URL records, even the cache keys themselves take unnecessary space in memory and extra time on eviction.

Best regards,

Rickard von Haugwitz

### DIFF
--- a/src/Libraries/Nop.Services/Localization/LocalizedEntityService.cs
+++ b/src/Libraries/Nop.Services/Localization/LocalizedEntityService.cs
@@ -170,27 +170,28 @@ namespace Nop.Services.Localization
         /// </returns>
         public virtual async Task<string> GetLocalizedValueAsync(int languageId, int entityId, string localeKeyGroup, string localeKey)
         {
+            if (_localizationSettings.LoadAllLocalizedPropertiesOnStartup)
+            {
+                //value tuples aren't json-serializable by default, so we use a string key
+                static string formatKey(string keyGroup, string key, int id) => $"{keyGroup}:{key}:{id}";
+
+                var localizedValues = await _staticCacheManager.GetAsync(
+                    _staticCacheManager.PrepareKeyForDefaultCache(NopLocalizationDefaults.LocalizedPropertyLookupCacheKey, languageId),
+                    async () => (await GetAllLocalizedPropertiesAsync(languageId))
+                        .GroupBy(p => formatKey(p.LocaleKeyGroup, p.LocaleKey, p.EntityId))
+                        .ToDictionary(g => g.Key, g => g.First().LocaleValue));
+
+                return localizedValues.TryGetValue(formatKey(localeKeyGroup, localeKey, entityId), out var localeValue)
+                    ? localeValue
+                    : string.Empty;
+            }
+
+            //gradual loading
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi @rickardvh!
I’m not entirely sure that the cache access nested in the anonymous method for GetAll could somehow affect performance, however, I must admit that your implementation is more understandable. 

Thank you very much for your work.

rickardvh:
> Hi @rickardvh! I’m not entirely sure that the cache access nested in the anonymous method for GetAll could somehow affect performance, however, I must admit that your implementation is more understandable.
> 
> Thank you very much for your work.

@skoshelev You're right, that part won't affect the execution speed (precomputing what values to store might, but the difference should be small - mostly more readable). The main difference is the number of keys in the cache, since we previously ended up storing every key twice (once in the cached dictionary, once individually in the cache). For us, that ended up being well over one million unnecessary keys. It's still not a *lot* in terms of memory usage, but well worth getting rid of. Also, since the memory cache itself doesn't support removing by prefix (recall that we prune the keys from the key manager), even though removing one by one from a dictionary is fast, it's still a lot of keys to be iterated over.

rickardvh:
@skoshelev There is one potential problem, though, that I realised just now. Moving the block outside of the cache getter and directly using the dictionary makes sense most of the time, except in one situation - if you are updating a lot of localised entities at once. We have a custom import plugin that does just that (thousands of updates running nonstop over up to an hour or longer), and since the (very big) cached dictionary is constantly refreshed by getting a huge db table, it has a noticably negative impact on performance while the import is ongoing. Previously, unrelated precached keys were not affected, but now they are.

Now, the question is whether this case is unusual enough that we should handle it ourselves, or if we should move the blocks back inside the cache getters. The ideal solution would be to directly update the dictionary on changes, but in my opinion that would deviate too much from the established pattern in the cache management, and increase the potential for bugs by adding more specialised code to event handlers.

Sorry it took us so long to identify this issue, the conditions are quite rare and difficult to isolate.

skoshelev:
Hi @rickardvh. I think that this is a very rare problem, and it seems to me that it is better to solve it by bypassing caching altogether, or maybe even better by bypassing the ORM. Plus, with so many locales, it would be better to disable the LoadAllLocalizedPropertiesOnStartup setting. I think that this case should not be considered as a potential problem.

rickardvh:
> Hi @rickardvh. I think that this is a very rare problem, and it seems to me that it is better to solve it by bypassing caching altogether, or maybe even better by bypassing the ORM. Plus, with so many locales, it would be better to disable the LoadAllLocalizedPropertiesOnStartup setting. I think that this case should not be considered as a potential problem.

@skoshelev OK! Well, we'll fix it ourselves, and if you encounter problems of this sort in the future, you know what might cause it and hopefully how to fix it.

rickardvh:
@skoshelev just a little update about this to show the effect it seems to have on memory usage. I didn't have time for a proper solution (suppressing cache events until the import is done) and just moved the blocks back inside the cache getters to add a level of protection. This is the memory usage chart for the past month, and it's pretty clear when that change was deployed...

image

Note that I restored previous code, so the past two weeks is what it looked like before the change in this pull request.

PR #6925, opened 2023-11-16
Optimize SortCategoriesForTreeAsync

rickardvh:
The time complexity of CategoryService.SortCategoriesForTreeAsync is currently $O(n^2)$ for $n$ categories. This is unnecessary, since we can use a lookup with $O(1)$ access instead of filtering a list in $O(n)$ on each recursion.

Although we can no longer sort entries all at once in SQL (potentially shifting execution time to the app server, if the database is hosted separately), it is much cheaper to sort many small lists instead of a single large one (unless the category tree is completely flat, in which case it's the same).

Best regards,

Rickard von Haugwitz


skoshelev:
Hi @rickardvh. Thanks, this is great work

Issue #6940, opened 2023-12-05
Add more database indexes (performance)


AndreiMaz:
nopCommerce version: 4.60.5

We can improve performance by adding more database indexes:
1. [Product].[Name]
2. [Forums_Topic].[Subject]

curllog:
Hi @AndreiMaz  @skoshelev  i have this [pull request ](https://github.com/nopSolutions/nopCommerce/pull/6944)can you please review it ? thanks

curllog:
@skoshelev  thanks for merge but i think we need to remove NonClustered() because it's sql server specefic [https://github.com/fluentmigrator/fluentmigrator/blob/b2158a93315968d3037d20de881a99333da9c942/src/FluentMigrator.Extensions.SqlServer/SqlServer/SqlServerExtensions.cs#L66](https://github.com/fluentmigrator/fluentmigrator/blob/b2158a93315968d3037d20de881a99333da9c942/src/FluentMigrator.Extensions.SqlServer/SqlServer/SqlServerExtensions.cs#L66) what do you think?

PR #6944, opened 2023-12-05
Add database index for product.name Forums_Topic.Subject

curllog:
Hi this closes [https://github.com/nopSolutions/nopCommerce/issues/6940](https://github.com/nopSolutions/nopCommerce/issues/6940)

### DIFF
--- a/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
+++ b/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
@@ -150,8 +150,7 @@ namespace Nop.Data.Migrations.UpgradeTo470
                     .OnColumn(nameof(Product.Name)).Ascending() 
                     .WithOptions().NonClustered();
 
-            if (!Schema.Table(NameCompatibilityManager.GetTableName(typeof(ForumTopic)))
-                    .Index("IX_Forums_Topic_Subject").Exists())
+            if (!Schema.Table(NameCompatibilityManager.GetTableName(typeof(ForumTopic))).Index("IX_Forums_Topic_Subject").Exists())
                 Create.Index("IX_Forums_Topic_Subject")
                     .OnTable(NameCompatibilityManager.GetTableName(typeof(ForumTopic)))
                     .OnColumn(nameof(ForumTopic.Subject)).Ascending()
--- a/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
+++ b/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
@@ -11,7 +11,7 @@ using Nop.Data.Mapping;
 
 namespace Nop.Data.Migrations.UpgradeTo470
 {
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi @curllog. These changes are not sufficient because the Installation/Indexes.cs migration is only performed during a clean installation. It is also necessary to add new indexes to the file https://github.com/nopSolutions/nopCommerce/blob/develop/src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs and change the date in the migration to the current one

curllog:
@skoshelev  I made the changes you requested hope it will be correct, Thanks

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Nop.Data.Migrations.UpgradeTo470
 {
-    [NopSchemaMigration("2023-11-09 00:00:02", "AddIndexesMigration for 4.70.0")]
+    [NopSchemaMigration("2023-12-06 19:53:00", "Add Index For [Product].[Name]-[Forums_Topic].[Subject]")]
```
skoshelev, line 14: I think that the description of migration should not be changed, let it remain the original one.

src/Libraries/Nop.Data/Migrations/UpgradeTo470/AddIndexesMigration.cs
```diff
@@ -143,6 +143,20 @@ public override void Up()
                 IfDatabase(databaseType).Create.Index("IX_ProductReview_CustomerId").OnTable(nameof(ProductReview))
                     .OnColumn(nameof(ProductReview.CustomerId)).Ascending()
                     .WithOptions().NonClustered();
+
+            if(!Schema.Table(nameof(Product)).Index("IX_Product_Name").Exists())
+                IfDatabase(databaseType).Create.Index("IX_Product_Name")
```
skoshelev, line 148: There is no need to check the database type; these indexes must be added to all supported DBMSs

curllog: @skoshelev Done

skoshelev:
@curllog Thank you for your help

skoshelev:
I don’t think that we need to remove this, I believe that for those servers that do not support this concept, this parameter will be ignored, and for MsSql we need this index to be NonClustered

curllog:
> I don’t think that we need to remove this, I believe that for those servers that do not support this concept, this parameter will be ignored, and for MsSql we need this index to be NonClustered

@skoshelev  I agree with you  and i think this is (and should be) handled in upstream library(fluentmigrator) since we are using database-agnostic namespace for migrations so do you recommend to open new issue and delete all IfDatabase(databaseType) from index migrations to 4.70? because it's fluentmigrator responsibility to check for supported migrations

skoshelev:
no, we will not remove the IfDatabase check since indexes are created there that are not found only in MsSql, but which are present in two other DBMSs by default (but we do not know the names of these indexes, and in order not to index the same thing twice, we simply ignore all DBMS except MsSql)

PR #6949, opened 2023-12-10
Clearing known proxies and networks of their defaults

chrisdrobison:
I've been struggling, along with many others with getting the SSL configuration going, especially using some external load balancer like the AWS ALB. This change should address the problems related to getting that working with other things. According to the docs (https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-8.0), KnownHeaders and KnownProxies comes with some default values. Having those defaults in there means that if the incoming proxy connection doesn't match a localhost network, the ForwardedHeadersMiddleware will exclude itself from processing the request. That means that SSL won't work correctly leading to all the issues that have been reported in the forums. This change allows for KnownProxies and KnownNetworks to be completely overridden, removing the default values. Specifying a null value in the JSON will leave the default. Specifying a string (including empty) will clear the default values before adding to the list.


chrisdrobison:
If you're interested, this would be good to put onto 4.60.x.

exileDev:
Hi! We added a similar thing to this commit https://github.com/nopSolutions/nopCommerce/commit/bc643dd32f145bebfe94068e9cd18fea0ae2b0fc . I think that's enough

Issue #6719, opened 2023-05-31
Issues from Rider Solution-Analysis


manuel3108:
nopCommerce version: develop

Steps to reproduce the problem: 
1. Download and install Rider
2. Wait a long time until the analysis has been completed.

Continuation of #6355 
Rider is currently reporting 313 errors, most of them seem accurate in my opinion. 4 of those errors are in c# files, the other ones in cshtml files. Currently i mainly see two sources for those errors:

---

**View 'x' cannot be found (Errors in cshtml files)**
1. https://github.com/nopSolutions/nopCommerce/blob/d738e3cc5c98e9d7ecef9559ab1f7aec18e4f5ae/src/Presentation/Nop.Web/Controllers/NewsController.cs#L225
2. https://github.com/nopSolutions/nopCommerce/blob/d738e3cc5c98e9d7ecef9559ab1f7aec18e4f5ae/src/Presentation/Nop.Web/Controllers/BlogController.cs#L239
3. https://github.com/nopSolutions/nopCommerce/blob/d738e3cc5c98e9d7ecef9559ab1f7aec18e4f5ae/src/Presentation/Nop.Web/Controllers/NewsController.cs#L225
4. https://github.com/nopSolutions/nopCommerce/blob/d738e3cc5c98e9d7ecef9559ab1f7aec18e4f5ae/src/Presentation/Nop.Web/Controllers/ProductController.cs#L360

All of those errors report back that the appropriate views cannot be found.

---

**Cannot resolve symbol 'item' (most of the cshtml errors)**
These errors appear to happen on nearly every `Html.LocalizedEditorAsync` Call.
Here is one of the many examples: https://github.com/nopSolutions/nopCommerce/blob/8070bc2dadeb85366705dc946efda8b4a58e2158/src/Plugins/Nop.Plugin.Payments.CheckMoneyOrder/Views/Configure.cshtml#L16

Currently, we usually have something like
```cs
@(await Html.LocalizedEditorAsync<ConfigurationModel, ConfigurationModel.ConfigurationLocalizedModel>("plugin-settings-localized",
     @<div>
....
```

something like this could do the trick (although untested). See `(item) =>` at the end:
```cs
@(await Html.LocalizedEditorAsync<ConfigurationModel, ConfigurationModel.ConfigurationLocalizedModel>("plugin-settings-localized", (item) =>
     @<div>
....
```

---

[...]

PR #6968, opened 2023-12-20
Fix @await syntax in cshtml files

curllog:
Hi  @AndreiMaz  this change will close this [issue](https://github.com/nopSolutions/nopCommerce/issues/6969) and @exileDev partially related to [this](https://github.com/nopSolutions/nopCommerce/issues/6719#issuecomment-1591001379)


exileDev:
Thanks!

PR #6971, opened 2023-12-21
Refactor and clean up UrlRecordService

rickardvh:
I cleaned up UrlRecordService by getting rid of unicode string parsing and locks, and simplified name formatting by replacing while/replace with regex.

However, I must ask what the rationale behind initialising `_seoCharacterTable` like that is, instead of just directly assigning it as readonly? I assume it is for the init method to be possibly overridden in a subclass, but if that is really necessary, wouldn't it be better to break it out into its own singleton class? That could then be modified and replaced in the DI container if necessary, without a subclass changing the behaviour of its superclass by changing a static member. My IDE is complaining that "Non-constant fields should not be visible", and I am inclined to agree.

Best regards,

Rickard von Haugwitz,
Majako [majako.se](https://majako.se)

### DIFF
--- a/src/Libraries/Nop.Services/Seo/UrlRecordService.cs
+++ b/src/Libraries/Nop.Services/Seo/UrlRecordService.cs
@@ -1,4 +1,5 @@
 ﻿using System.Text;
+using System.Text.RegularExpressions;
 using Nop.Core;
 using Nop.Core.Caching;
 using Nop.Core.Domain.Localization;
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @rickardvh. I looked at the changes, and overall I liked everything.
As for filling out _seoCharacterTable, I think that a lot here depends on remnants of the past. I don’t like the way it was or the way it is now. Carrying this data into a separate class is an interesting solution, but in this case, it is redundant. It seems to me that it’s best to just make a static constructor and fill this dictionary. This way we will solve the issue with the warning “Non-constant fields should not be visible”, and at the same time leave the opportunity to change something from the child class

Issue #405, opened 2016-01-18
Write more unit tests


mariannk:
Write more unit tests:
Nop.Services.Tests. Write tests for services
Nop.Web.MVC.Tests. Write tests for validators (admin area)
Nop.Web.MVC.Tests. Write tests for controllers
Nop.Web.MVC.Tests. Write tests for classes located into Nop.Web.Framework projects

Useful links:
http://stephenwalther.com/blog/archive/2011/12/22/asp-net-mvc-selenium-iisexpress.aspx
http://richarddingwall.name/2009/08/19/asp-net-mvc-tdd-and-fluent-validation/
http://nopcommerce.codeplex.com/SourceControl/network/Forks/kchugh/kchugh/contribution/1815

abdullohmuhamedov:
for which one unit test needs to be written?

skoshelev:
Ideally, we need tests for all services and all factories

PR #6982, opened 2024-01-03
More unit tests for vendor and Topic service

richasinghthakur40:
More unit tests #405
Added Testcases for Vendor Service.
Added Testcases for Topic Service.


skoshelev:
Hi, @richasinghthakur40. Thank you for your desire to help us with code coverage with tests. We do not use mocks in our tests; instead, you can get the desired interface from the IoC container by calling the GetService method (Just inherit your class from ServiceTest class)

PR #6985, opened 2024-01-03
unittest for cms

Kumudhini:
More unit tests for cms #405 
These are the Test Case for CMS 

src/Tests/Nop.Tests/Nop.Services.Tests/Cms/WidgetPluginInterfaceTests.cs
src/Tests/Nop.Tests/Nop.Services.Tests/Cms/WidgetPluginManagerInterfaceTests.cs
src/Tests/Nop.Tests/Nop.Services.Tests/Cms/WidgetPluginManagerTests.cs


skoshelev:
Hi, @Kumudhini. I'm very sorry, I don't quite understand the application value of your tests, what kind of code do they test? We specifically moved away from Mocks in testing by introducing an IoC container to test the final implementation and not interface signatures.

PR #6989, opened 2024-01-05
Test cases Added for Address Service

Dakshi507:


skoshelev:
Hi @Dakshi507. Thank you for your desire to help us with code coverage with tests. We do not use mocks in our tests; instead, you can get the desired interface from the IoC container by calling the GetService method

PR #6990, opened 2024-01-05
AutoUpdate the product total

mahendragowdabs94:
#2090 Automatically update products fields (total) when editing order items (price, qty).

this is the update code:
orderItem.PriceInclTax = unitPriceInclTax * quantity;
orderItem.PriceExclTax = unitPriceExclTax * quantity;


RomanovM:
This is not as trivial as it seems. Perhaps a lot more changes will be needed here. In any case, it’s easier for our team to implement this than to describe all the details for third-party developers.

PR #6994, opened 2024-01-08
Added tags implementation for Topics

zaidanashraf17:
Add tags implementation for Topics.
Just as you have TAGS available for the catalog (products), I have extended for TOPICS.

### DIFF
--- a/src/Libraries/Nop.Core/Domain/Catalog/TopicTag.cs
+++ b/src/Libraries/Nop.Core/Domain/Catalog/TopicTag.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Nop.Core.Domain.Localization;
+using Nop.Core.Domain.Seo;
+
+namespace Nop.Core.Domain.Catalog;
+public class TopicTag: BaseEntity, ILocalizedEntity, ISlugSupported
+{
+    /// <summary>
+    /// Gets or sets the name
+    /// </summary>
+    public string Name { get; set; }
+}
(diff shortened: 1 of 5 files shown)


src/Presentation/Nop.Web/Models/Catalog/PopularTopicTagsModel.cs
```diff
@@ -0,0 +1,86 @@
...
+    return GetFontSize(topicTag.TopicCount, mean, stdDev);
+}
+
+#endregion
+
+#region Properties
+
+public int TotalTags { get; set; }
+
+public List<TopicTagModel> Tags { get; set; }
```
apurvai007, line 83: I noticed that you are using `List<TopicTagModel>` for the Tags property. Can we use `IList<TopicTagModel>` instead?

src/Libraries/Nop.Core/Domain/Catalog/TopicTag.cs
apurvai007, line 1: can we add XML comments to document the purpose and usage of the class as well , like you did for property?
 Just like we are using for all catalogs see ProductTags.cs for example.

src/Presentation/Nop.Web/Models/Catalog/PopularTopicTagsModel.cs
apurvai007, line 1: can we add comments to document the purpose and usage of the class and its members?

src/Presentation/Nop.Web/Models/Catalog/TopicTagModel.cs
apurvai007, line 1: can we add comments to document the purpose and usage of the class and its properties?

src/Presentation/Nop.Web/Models/Catalog/TopicTagModel.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Nop.Web.Framework.Models;
+
+namespace Nop.Web.Models.Catalog;
+
+public partial record TopicTagModel : BaseNopEntityModel
+{
+    public string Name { get; set; }
+
+    public string SeName { get; set; }
```
apurvai007, line 9: can we use a more descriptive name for the SeName property?

apurvai007 (changes requested):

Issue #6978, opened 2023-12-28
Add LanguageId field to NewsletterSubscription


RomanovM:
nopCommerce version: 4.60

>I was very surprised to see that nopcommerce doesn't store the language (a.e. current UI language) of the user subscribing to the newsletter. I own a shop in a contry with multiple user languages. Informing the user in a language he can read is a very importent sales aspect - and multilanguage newsletters look horrible and uncustomized. So this greate feature is nearly useless.

Source: https://www.nopcommerce.com/boards/topic/98426/newsletter-subscriber-language

An important addition. It works fine for registered customers. But we also should support guest users who are not logged in and use the newsletter box. I think an extra property should be added to `NewsLetterSubscription`. So besides the `StoreId,` the `SubscriptionLanguageId` can be stored here as well. Let's ensure that it's synchronized for registered customers (they have the current language ID stored in database)

DmitriyKulagin:
Closed #6978

PR #7003, opened 2024-01-12
Adding Subscription Language

prasannapolina:
#6978 Add LanguageId field to NewsletterSubscription

1.Added new language field "Subscription Language" in the NewsLetterSubscription table.
2.To store the data in the new field for both guest and subscribed user made changes in the src/Presentation/Nop.Web/Controllers/NewsletterController.cs file

### DIFF
--- a/src/Libraries/Nop.Core/Domain/Messages/NewsLetterSubscription.cs
+++ b/src/Libraries/Nop.Core/Domain/Messages/NewsLetterSubscription.cs
@@ -19,6 +19,8 @@ public partial class NewsLetterSubscription : BaseEntity
     /// Gets or sets a value indicating whether subscription is active
     /// </summary>
     public bool Active { get; set; }
+    public int SubscriptionLanguage { get; set; }
+    
 
     /// <summary>
     /// Gets or sets the store identifier in which a customer has subscribed to newsletter
--- a/src/Presentation/Nop.Web/Controllers/NewsletterController.cs
+++ b/src/Presentation/Nop.Web/Controllers/NewsletterController.cs
@@ -61,10 +61,14 @@ public partial class NewsletterController : BasePublicController
         {
             email = email.Trim();
             var store = await _storeContext.GetCurrentStoreAsync();
-            var subscription = await _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync(email, store.Id);
             var currentLanguage = await _workContext.GetWorkingLanguageAsync();
+            var subscription = await _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync(email, store.Id);
+            //subscription.SubscriptionLanguage = currentLanguage.Id;
             if (subscription != null)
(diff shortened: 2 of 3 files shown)


apurvai007:
**Question** : To complete this request, we need to alter the table `NewsSubscription` and add a new column `SubscriptionLanguage` in `schemaMigration` , correct?

skoshelev:
Hi @prasannapolina. 
@apurvai007 is right, just adding a field to the entity is not enough as that will only work for a clean install. We still need to add it to the migration. Also in the note to the ticket, it is written that it is necessary to take into account the user’s ability to change the language while using the site. That is, the new field should change after Customer.LanguageId.

src/Libraries/Nop.Core/Domain/Messages/NewsLetterSubscription.cs
```diff
@@ -19,6 +19,8 @@ public partial class NewsLetterSubscription : BaseEntity
     /// Gets or sets a value indicating whether subscription is active
     /// </summary>
     public bool Active { get; set; }
+    public int SubscriptionLanguage { get; set; }
```
skoshelev, line 22: Better to use the LanguageId as a name. And also we need a doc string to the new field.

src/Presentation/Nop.Web/Controllers/NewsletterController.cs
```diff
@@ -61,10 +61,14 @@ public virtual async Task<IActionResult> SubscribeNewsletter(string email, bool
         {
             email = email.Trim();
             var store = await _storeContext.GetCurrentStoreAsync();
-            var subscription = await _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync(email, store.Id);
```
skoshelev, line 64: I think that changes in these three lines are not necessary

src/Presentation/Nop.Web/Controllers/NewsletterController.cs
```diff
@@ -61,10 +61,14 @@ public virtual async Task<IActionResult> SubscribeNewsletter(string email, bool
...
             email = email.Trim();
             var store = await _storeContext.GetCurrentStoreAsync();
-            var subscription = await _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync(email, store.Id);
             var currentLanguage = await _workContext.GetWorkingLanguageAsync();
+            var subscription = await _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync(email, store.Id);
+            //subscription.SubscriptionLanguage = currentLanguage.Id;
             if (subscription != null)
+                
             {
+                subscription.SubscriptionLanguage = currentLanguage.Id;
```
skoshelev, line 70: These changes are not needed since there is no entity saving in this block.

DmitriyKulagin:
Hi @prasannapolina 
We decided to implement this task on our own, since solving it requires several more changes than are proposed here. Still, thank you for your help.

Issue #7006, opened 2024-01-14
Error when Manufacturer is null


AxelAndersen:
https://github.com/nopSolutions/nopCommerce/blob/2c130c60018be960808382496e1aeeeb0a9acc6e/src/Presentation/Nop.Web/Controllers/CatalogController.cs#L493

AxelAndersen:
Maybe return false in top of method if Manufacturer is null

PR #7010, opened 2024-01-14
#7006 - Fixed error when Manufacturer is null

AxelAndersen:
Returning false if Manufacturer do not exist.


RomanovM:
@AxelAndersen Thanks for your help.

PR #7016, opened 2024-01-16
fix (unittests) #6948 Resolve issues with tests creating a stack overflow. Upgrade to NUnit4

### LINKED ISSUE #6948: After switching to the .NET 8 infrastructure, some unit tests stopped running
nopCommerce version: develop

Some tests stopped running and crashed with a `Stack overflow` error.
image

[This ticket](https://developercommunity.visualstudio.com/t/Upgrading-to-version-1780-and-higher-c/10525329) is also sent to support Microsoft. 

The problem occurs only when SDK 8.0.100 is installed on the machine. Old versions (on .NET 7) also break in tests.

Related: https://github.com/nopSolutions/nopCommerce/issues/6664

jafin:
fix (Tests): Fix Stackoverflow in tests and other issues.

- Update to NUnit 4
- Register missing services
- Change scope of IDataProviderManager (causing a stackoverflow in net8)
- Reduce state over tests by changing OneTimeSetup/Teardown to Setup/Teardown. TearDown
- Reduce coupling to DI for resolving Validators, Construct in test with dependencies (still via DI for now)
- Reduce the GetService Magic where it will construct a dependency even when it can't resolve it.  Either be explicit or register it.
- Some tests were using the implmenetation class not the interface so DI was not working (but the GetService magic filled the gap when the DI could not resolve from container.)

Resolves #6948

### DIFF
--- a/src/Tests/Nop.Tests/BaseNopTest.cs
+++ b/src/Tests/Nop.Tests/BaseNopTest.cs
@@ -83,6 +83,7 @@ using Nop.Web.Framework.Models;
 using Nop.Web.Framework.Mvc.Routing;
 using Nop.Web.Framework.Themes;
 using Nop.Web.Framework.UI;
+using Nop.Web.Framework.WebOptimizer;
 using Nop.Web.Infrastructure.Installation;
 using SkiaSharp;
 using IAuthenticationService = Nop.Services.Authentication.IAuthenticationService;
@@ -99,6 +100,8 @@ public partial class BaseNopTest
     {
         SetDataProviderType(DataProviderType.Unknown);
     }
+
+    public ServiceProvider ServiceProvider => _serviceProvider;
 
     private static void Init()
     {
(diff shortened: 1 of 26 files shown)


skoshelev:
Hi @jafin. Many thanks for your help. I don't completely agree with your opinion about the logic that was in GetService, but if it gave you a feeling of magic, then it may be difficult for other people to understand it at all. We always try to make our code as clear and readable as possible

jafin:
@skoshelev apologies, it was just that it was doing something 'unexpected' ie Resolving from a DI, but then if it couldn't do that it was instantiate the object itself (bypassing the DI).  I think this kind of code can lead to errors.  And in a few cases where the tests where using concrete classes over interface and getting resolved, can lead to hard to troubleshoot bugs. ?  But certainly if there is benefit there please revert those changes or I can submit a revert PR.     

Also it was probably an error to include all the changes in the one PR. Initial goal was just to get the tests to operate again

PR #7046, opened 2024-01-28
refactor (jQuery): Update to non-deprecated ready signature

jafin:
-per https://api.jquery.com/ready/ only one ready signature is not deprecated in jQuery 3+

> jQuery offers several ways to attach a function that will run when the DOM is ready. All of the following syntaxes are equivalent:
> 
> $( handler )
> $( document ).ready( handler )
> $( "document" ).ready( handler )
> $( "img" ).ready( handler )
> $().ready( handler )
> As of jQuery 3.0, **only the first syntax is recommended; the other syntaxes still work but are deprecated**.


exileDev:
Thanks!

Issue #7048, opened 2024-01-29
Fix incorrect documentation in HtmlFormatter


jared-chevalier:
In `Nop.Services.HtmlFormatter`, the documentation comment for the utility method `EnsureOnlyAllowedHtml` includes an incorrect description for the return value.

The method signature is `protected static string EnsureOnlyAllowedHtml(string text)`, and the method implementation returns the validated text with all invalid tags removed. However, the documentation indicates that the method returns a `bool` value to indicate the validity of the original text.

Actual documentation: `<returns>True - if the text contains only valid tags, false otherwise</returns>`.

Expected documentation: `<returns>Validated text with all invalid tags removed</returns>`.

PR #7049, opened 2024-01-29
#7048 Fix incorrect documentation in HtmlFormatter

jared-chevalier:
### Issue #7048

In `Nop.Services.HtmlFormatter`, the documentation comment for the utility method `EnsureOnlyAllowedHtml` includes an incorrect description for the return value.

The method signature is `protected static string EnsureOnlyAllowedHtml(string text)`, and the method implementation returns the validated text with all invalid tags removed. However, the documentation indicates that the method returns a `bool` value to indicate the validity of the original text.

Actual documentation: `<returns>True - if the text contains only valid tags, false otherwise</returns>`.

Expected documentation: `<returns>Validated text with all invalid tags removed</returns>`.

### Modifications

- Modify `EnsureOnlyAllowedHtml` `text` parameter description for consistency with same parameter in other methods.
- Modify `EnsureOnlyAllowedHtml` return description for correctness.
- Ensure that "HTML" is written in all capitals.

### DIFF
--- a/src/Libraries/Nop.Services/Html/HtmlFormatter.cs
+++ b/src/Libraries/Nop.Services/Html/HtmlFormatter.cs
@@ -28,8 +28,8 @@ public partial class HtmlFormatter : IHtmlFormatter
     /// <summary>
     /// Ensure only allowed HTML tags
     /// </summary>
-    /// <param name="text">Text to check</param>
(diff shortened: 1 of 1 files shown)


RomanovM:
@jared-chevalier Thanks for your help.

jared-chevalier:
@RomanovM You're welcome, happy to help.

PR #7058, opened 2024-02-05
Refactorint to work with Chart.js 4.4

peysche:
Hi,

feel free to use the code and change from Chart.js 2 to Chart.js 4

I needed a V4 feature for my need so i had to upgrade this 2 Views ..

Thx


peysche:
in Addition the .JS file of Chart JS has tobe updated 
https://cdnjs.cloudflare.com/ajax/libs/Chart.js/4.4.1/chart.umd.min.js

AndreiMaz:
https://www.nopcommerce.com/en/boards/topic/98658/jquery-400-upgrade#311095

exileDev:
Thank you! We've done it [here](https://github.com/nopSolutions/nopCommerce/commit/615800db5d1d0f8e2813e14d5452d55d4f469dda)

PR #7068, opened 2024-02-14
Update GoogleAnalyticsHttpClient to be ThreadSafe

deathgore:
Because HttpClient is a shared object it produces the following exception:
System.InvalidOperationException: This instance has already started one or more requests. Properties can only be modified before sending the first request.
   at System.Net.Http.HttpClient.set_BaseAddress(Uri value)
   at Nop.Plugin.Widgets.GoogleAnalytics.Api.GoogleAnalyticsHttpClient.RequestAsync(EventRequest request, GoogleAnalyticsSettings googleAnalyticsSettings)
   at Nop.Plugin.Widgets.GoogleAnalytics.EventConsumer.ProcessOrderEventAsync(Order order, GoogleAnalyticsSettings googleAnalyticsSettings, String eventName)

The PR removes the Set BaseAddress and Timeout from the HttpClient inside the RequestAsync method.

Instead of HttpClient now GoogleAnalyticsHttpClient has dependency on IHttpClientFactory and httpClient instance is created using  _clientFactory.CreateClient(nameof(GoogleAnalyticsHttpClient));

BaseAddres is remove, and it is set on HttpRequestMessage instance

### DIFF
--- a/src/Plugins/Nop.Plugin.Widgets.GoogleAnalytics/Api/GoogleAnalyticsHttpClient.cs
+++ b/src/Plugins/Nop.Plugin.Widgets.GoogleAnalytics/Api/GoogleAnalyticsHttpClient.cs
@@ -13,15 +13,15 @@ public class GoogleAnalyticsHttpClient
 {
     #region Fields
 
-    protected readonly HttpClient _httpClient;
+    private readonly IHttpClientFactory _clientFactory;
 
     #endregion
 
     #region Ctor
 
-    public GoogleAnalyticsHttpClient(HttpClient httpClient)
+    public GoogleAnalyticsHttpClient(IHttpClientFactory clientFactory)
     {
-        _httpClient = httpClient;
+        _clientFactory = clientFactory;
     }
 
     #endregion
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi @deathgore. Many thanks for your help. I looked at the changes you propose and do not find any practical sense in them. The thing is that when we register our HttpClient through the IServiceCollection.AddHttpClient<T> call, [this code](https://github.com/dotnet/dotnet/blob/759130b3f38eeefd59424a0ab32262903c2a4024/src/runtime/src/libraries/Microsoft.Extensions.Http/src/DependencyInjection/HttpClientBuilderExtensions.cs#L386) will eventually run. It shows that internally IHttpClientFactory is used to create a named HttpClient.
I'm having a hard time pinpointing your problem with InvalidOperationException, since GoogleAnalyticsHttpClient is not the only class created in the same way. Even in the nopCommerce core code, there is StoreHttpClient and NopHttpClient. The same as in many of our plugins. And we have not encountered such a problem before.
In theory, this problem can arise if several GoogleAnalyticsHttpClient.RequestAsync methods are called in one request without waiting for completion (for example, the await instruction is missed). However, in the latest version of the plugin, there is only [one place ](https://github.com/nopSolutions/nopCommerce/blob/develop/src/Plugins/Nop.Plugin.Widgets.GoogleAnalytics/Events.cs#L160)that calls this method and it is done correctly.
Please check exactly how this method is called in your code, and if they are also formatted correctly, then create an issue so that we can deal with the problem. Please be sure to include your nopCommerce version and a list of steps to reproduce the issue.

Issue #7076, opened 2024-02-16
Optimize address handling


professor-k:
nopCommerce version: 4.60.6

Steps to reproduce the problem: 
Create an order with same billing and shipping address. In Adresss table you'll get two extra copies of your address.

I understand the logic behind those copies (permanent copies for order history out of user's reach), yet I think the way it is currently implemented is significantly excessive. Most users will have more orders than changes to their address, so this whole duplication system gives no benefit in their case. In case there is pickup address, it's going to be copied thousand times too. For store that operates for a while, those addresses build up into significant numbers.

Looking through the code, I can immediately suggest to skip creation of extra copy if shipping and billing is same, that will reduce duplicates by half for majority of users. That's a cheap fix, but it only reduces problem.

More fundamental (and considerably more breaking) solution would be to treat addresses as immutable, update = creation of new address. This way there is going to be no need to create duplicates at all.

AndreiMaz:
It's by design so they could be edited separately in admin area after an order is placed. We're leaving it for customization if some merchants want to chnage this behavior

professor-k:
For the record: Still, it's pretty inefficient design. 99% of addresses are never changed, but to support that odd 1% we spend trice the space on exact duplicates, which builds up to one of the biggest tables in the database. 

Yes, I understand that db space is cheap nowadays. Still it's not free, both literally (at least in case of MS SQL server) and performance-wise. I took inventory on our store's website, and discovered that there are plenty of waste it top tables. Again, I'm only sharing my findings on one particular store's database that lasted for decade, YMMW.

So, Customers table takes the lead with piles of long time forgotten guests that once added something to their cart. Scheduled task similar to one there is for guests can deal with that.

Then there is surprisingly huge Customer_CustomerRole_Mapping, which along with the indexes takes way more space that I would expect. Given that only role for basically every customer is either Guest or Registered, that leads to question why those two even are roles, not just a bit flag in Customer table itself? There is performance gain hidden in this one too. 

And then there are Addresses table which is bloated 3 times with duplicates. Most obvious waste, but probably not the easiest to get rid of. Can be dealt with by making copies only in case of editing.

I suppose I can continue, but it's getting way less obvious and less impressive from here on.

PR #7077, opened 2024-02-16
Don't duplicate shipping address for order if it's same as billing

professor-k:
Low hanging fruit for #7076.


AndreiMaz:
It's by design so they could be edited separate in admin area afte an order is placed

PR #7080, opened 2024-02-18
fix (StorePickup/Validation): Change to non-deprecated method in FluentValidation.

jafin:
Change to non-deprecated `PrecisionScale` method in `FluentValidation`.
See https://github.com/FluentValidation/FluentValidation/issues/2030 for reason.


skoshelev:
Hi, @jafin. Thanks for the help

Issue #7115, opened 2024-03-21
Update deprecated packages


danzuep:
nopCommerce version: 
- 4.70.8

Steps to reproduce the problem: 
- Download source, then check NuGet packages.

AndreiMaz:
Duplicates https://github.com/nopSolutions/nopCommerce/issues/196

Issue #196, opened 2016-01-15
Periodically update all third-party packages


mariannk:
1. NuGet
2. javascript libraries
3. user agent strings (now moved to http://udger.com/). So we cannot use them anymore
4. MaxMind GeoIP.dat - http://www.maxmind.com/en/geolocation_landing и https://www.maxmind.com/en/opensource

ngtrian:
TinyMCE 5
https://www.tiny.cloud/docs/release-notes/release-notes50/

ecommerce-tester:
iTextSharp is EOL, and has been replaced by iText 7.
Can we updated from iTextSharp to iText 7?

AndreiMaz:
> iTextSharp is EOL, and has been replaced by iText 7.
> Can we updated from iTextSharp to iText 7?

Thanks. Moved to https://github.com/nopSolutions/nopCommerce/issues/4622

exileDev:
Data providers: 734e2c63171ef22b3679fb07c9d5bd80c26788b1

sateesh-m-pg:
Any plans to update to latest Bootstrap version as well?

PR #7116, opened 2024-03-21
Update deprecated packages

danzuep:
Update deprecated packages for #7115


AndreiMaz:
Thanks a lot for your contribution. But we update third-party libraries before the release. We'll do it as part of https://github.com/nopSolutions/nopCommerce/issues/196

PR #7120, opened 2024-03-23
Refactor: split the InstallationService into three parts

danzuep:
The InstallationService is massive, but since it's already a partial class, splitting it in three makes it easier to work with. An alternative way of doing this would be to inject two other services, but the original authors can decide on that later.


danzuep:
21 hours ago some new changes were made to the Installation Service that aren't in this pull request, so I'll close this.

Issue #7141, opened 2024-04-17
Use referrer url to find the correct breadcrumb instead of the first productcategory foun


lintho:
nopCommerce version: 4.6

Steps to reproduce the problem: 
navigate to a product with several product categories
first productcategory added on the product is whats being used to generate breadcrumb
the breadcrumb should instead be generated by the previous path the user was on (the category)

RomanovM:
Duplicate of #5776

PR #7142, opened 2024-04-17
use referrer url to generate a more correct breadcrumb

lintho:
…category id to generate breadcrumb

#7141


RomanovM:
Thanks for your suggestion. We'll leave this for customization.

PR #7164, opened 2024-05-01
Possibility to do not select 'ShippingMethod' to filter in 'ShippingByWeightByTotal' plugin

Dev-Sina:
Some times, it's need to filter shipping rates with all shipping methods, not only one.


Dev-Sina:
Suppose that you want to know for a specific state, which shipping methods are active and how they are configured to calculate the shipping rate. Now, you should filter country, state and shipping method and write them down and change the shipping method and write them down all on a paper! But if you could select all shipping method, you will do that easily.

skoshelev:
Hi, @Dev-Sina. Thank you for your help.

PR #7184, opened 2024-05-11
Preserve inline edit field values on backend validation errors

bhautikpatel1907:
In the Admin area, added functionality to retain inline edit field values in the grid when backend validation errors occur during the update process. Previously, upon encountering validation errors, the grid would refresh, causing users to lose all entered field values. This enhancement ensures a smoother editing experience for administrators by maintaining their input even in error scenarios.

Changes Made:
=> Table.cshtml : - Added logic to show update and cancel buttons without grid refresh upon AJAX response in `updateRowData_` function

[12.05.2024 02_44.webm](https://github.com/nopSolutions/nopCommerce/assets/104594350/0c55bcba-a60e-44df-a5a0-d6691eca7f10)

### DIFF
--- a/src/Presentation/Nop.Web/Areas/Admin/Views/Shared/Table.cshtml
+++ b/src/Presentation/Nop.Web/Areas/Admin/Views/Shared/Table.cshtml
@@ -270,6 +270,11 @@
             }
 
             function updateRowData_@(tableName)(currentCells, data, nameData) {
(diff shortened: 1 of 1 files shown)


DmitriyKulagin:
The current implementation does not provide for data validation depending on its type. All setting values ​​are of string type, so there will be no validation errors.

Issue #7202, opened 2024-05-23
Copying a product with attribute combinations results in incorrect StockQuantityHistory entry


GordonBlahut:
I believe there is a bug in the `ProductCopyService.CopyAttributesMappingAsync` method at the call to `_productService.AddStockQuantityHistoryEntryAsync` on lines 371-373.

The `combinationId` parameter receives `combination.Id`, but I believe that `combinationCopy.Id` would be more appropriate. Currently, the resulting entry in the `StockQuantityHistory` table references the product copy's id but the original comination's id and this causes issues when trying to reconcile inventory history for products with attribute combinations.

https://github.com/nopSolutions/nopCommerce/blob/98adab92f9c17f23b7d719a130497c518b7d234c/src/Libraries/Nop.Services/Catalog/CopyProductService.cs#L341-L373

I would also argue there's a valid use case for wanting to copy a product and attribute combinations *without copying any inventory information* but that's a separate issue.

parshan2:
You are absolutely right. This issue also causes an exception error when copying products with attribute combinations. I have also fixed two other parameter issues in the AddStockQuantityHistoryEntryAsync method call. I have just created an equivalent pull request for this.

skoshelev:
Closed #7202

PR #7204, opened 2024-05-26
Fixed the issue with StockQuantityHistory for product attribute combinations when copying a product with combinations. Resolves issue #7202.

parshan2:
Fixed the issue with StockQuantityHistory for product attribute combinations when copying a product with combinations. Resolves issue #7202.


parshan2:
it also create the message with the wrong product id

skoshelev:
Hi @parshan2. Thank you for your help

PR #7240, opened 2024-06-19
Feature/bitcoin payment plugin

pa0101:
- Built a payment plugin for the shopping cart. No BTC payment gateway is implemented yet, I'll need to do more research on how it works and also learn/understand payment gateway compliance to ensure legal and regulatory adherence.
- So far, I am able to build and run the application with my changes and display some text.


AndreiMaz:
@pa0101 Thanks a lot for your contribution. But we do not plan to include new payment plugins out of the box. Please feel free to implement and submit it to our marketplace with other [payment plugins](https://www.nopcommerce.com/en/extensions?category=payment-modules).

PR #7254, opened 2024-07-09
Update Methods

KhanbalaRashidov:


AndreiMaz:
@KhanbalaRashidov Thanks a lot for youe contribution. It looks good. But we won't merge it just because it'll make changes tracking more difficult for developers (upgrading from previous versions of nopCommerce)

KhanbalaRashidov:
Thank you

PR #7283, opened 2024-08-07
Fix db call in PermissionService.cs to async

Shane32:


skoshelev:
Hi @Shane32. Thank you very much for your help

PR #7284, opened 2024-08-07
Refactor SplitCamelCaseWord method

holydk:
```

BenchmarkDotNet v0.14.0, Windows 11 (10.0.22631.3007/23H2/2023Update/SunValley3)
Unknown processor
.NET SDK 8.0.303
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2


```
| Method | StringToSplit        | Mean       | Error     | StdDev    | Gen0     | Gen1   | Allocated  |
|------- |--------------------- |-----------:|----------:|----------:|---------:|-------:|-----------:|
| New    | This(...)Case [1785] |   4.076 μs | 0.0279 μs | 0.0261 μs |   0.7095 | 0.0153 |   11.66 KB |
| Old    | This(...)Case [1785] | 154.659 μs | 1.4227 μs | 1.3973 μs | 255.1270 | 5.8594 | 4170.13 KB |

### DIFF
--- a/src/Libraries/Nop.Core/CommonHelper.cs
+++ b/src/Libraries/Nop.Core/CommonHelper.cs
@@ -1,6 +1,7 @@
 ﻿using System.ComponentModel;
 using System.Globalization;
 using System.Net;
+using System.Text;
 using System.Text.RegularExpressions;
 using Nop.Core.Infrastructure;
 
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @holydk! This method converts values ​​from enumerations if they do not have a locale in the DB. It would be interesting to see the difference in performance when parsing, say, the ReadonlyCheckboxes value.

holydk:
Hi @skoshelev. I just saw [Added new ACL rules](https://github.com/nopSolutions/nopCommerce/commit/40692e04e9cf3dc6ad2148008576ef78626221fb) commit added that changed the method and decided to do a little optimization.

skoshelev:
Hi @holydk. I understand, but I think that such optimization is unnecessary, the code readability will deteriorate and the real performance will not change. In addition, then, logically, it will be necessary to rewrite another 34 places where the Aggregate method is used. But it seems to me that in this case, readability is still a priority.

PR #7288, opened 2024-08-08
Fix CI tests due to concurrency problems

Shane32:
When running tests within a GitHub Action runner, tests intermittently fail with `ObjectDisposedException` within the database provider.  This appears to be a concurrency-related problem.  Configuring NUnit to only run a single test at a time fixes the problem.


skoshelev:
Hi @Shane32. Thank you for your help

Issue #7289, opened 2024-08-09
VendorModel: Set PictureModel in PrepareVendorModelAsync


OzanOzcelikk:
nopCommerce version: 4.70

Steps to reproduce the problem: 

In VendorModel, there is a PictureModel property; however, it is not being set in the PrepareVendorModelAsync method. As a result, when users attempt to use the vendor's picture on pages like Views/Catalog/Vendor.cshtml, the picture is unavailable and returned as null. This means the vendor's image cannot be displayed or used on these pages.

AndreiMaz:
https://github.com/nopSolutions/nopCommerce/pull/7291

exileDev:
Closed #7289

AndreiMaz:
See related bug - https://github.com/nopSolutions/nopCommerce/issues/7289

PR #7291, opened 2024-08-09
Set `PictureModel` in `PrepareVendorModelAsync` method

OzanOzcelikk:
This pull request addresses issue #7289 where the `PictureModel` property in `VendorModel` was not being set in the `PrepareVendorModelAsync` method. With this update, `PictureModel` is now properly set, ensuring that vendor pictures are available and not null on pages like `Views/Catalog/Vendor.cshtml`.

**Changes Made:**
- Updated the `PrepareVendorModelAsync` method to set the `PictureModel` within `VendorModel`.
- Ensured that `PictureModel` is no longer null and accessible for display on relevant pages.

**nopCommerce version:**
4.70

In VendorModel, there is a PictureModel property; however, it is not being set in the PrepareVendorModelAsync method. As a result, when users attempt to use the vendor's picture on pages like `Views/Catalog/Vendor.cshtml`, the image is unavailable and returned as null. This update ensures that users can access and display the vendor's picture if they wish to include it.

**Related Issue:**
This pull request fixes issue #7289.


exileDev (approved):

exileDev:
Thanks!

PR #7323, opened 2024-08-27
When NumberOfProductTags is set 0, don't show popular tags widget on …

miritreus:
…catalog pages.

Used same logic as in ManufacturerNavigationViewComponent. NumberOfProductTags property is only used in PopularProductTagsViewComponent, so no sideeffects elsewhere when set to 0 to hide it.


AndreiMaz:
@miritreus Please do it the same way as https://github.com/nopSolutions/nopCommerce/blob/82ede01dacc0d8dd85ec72e5f5bebf7ecb7b8e84/src/Presentation/Nop.Web/Components/ManufacturerNavigationViewComponent.cs#L21C9-L21C68:

if (_catalogSettings.ManufacturersBlockItemsToDisplay == 0) 
and not < 1

PR #7329, opened 2024-09-06
use modern google maps icon in admin area

danFbach:
superficial, i know, but the current icon is probably a decade old at this point.
this updates it to the modern variant, taken directly from maps.google.com


RomanovM:
@danFbach Thanks. There is another icon like this below on this page, could you please replace it as well?

danFbach:
@RomanovM got it. Sorry, was porting from changes I made in 4.6 where only one icon exists.

PR #7344, opened 2024-09-22
Changed the CreateDataConnection signature to allow Transactions with Rollback and Commit options

TekyaygilFethi:
Hello,

In this PR, I have changed the CreateDataConnection signature on the BaseDataProvider class and added the function signature to the INopDataProvider interface to make it accessible from the project.

### Why?

Because while adding the orders and order items relevant to them, I needed to have a transaction block to rollback the changes when any exception occurred. I couldn't find any solution to that, so I have exposed the protected method as public to create a DataConnection object through the DataProvider and used it explicitly to create transactions and allow multiple operations to share the same DataConnection object.

Here, you can an example to my case:

```cs
 public class MyService : IMyService
 {
     private readonly INopDataProvider _nopDataProvider;
     private readonly IShoppingCartItemApiService _shoppingCartItemApiService;
     private readonly IDTOHelper _dtoHelper;

     public MyService(INopDataProvider nopDataProvider,
         IShoppingCartItemApiService shoppingCartItemApiService,
         IDTOHelper dtoHelper)
     {
         _nopDataProvider = nopDataProvider;
         _shoppingCartItemApiService = shoppingCartItemApiService;
         _dtoHelper = dtoHelper;
     }

     public async Task CreateOrder(int customerId)
     {
         // If you want to read data from the database, you can use the following code via using the _nopDataProvider object.
         var user = await _nopDataProvider.GetTable<Customer>().SingleOrDefaultAsync(_ => _.Id == customerId) ?? throw new ArgumentException("User not found");

         //Prepare data START
         var shoppingCartItemsToBePurchased = await _shoppingCartItemApiService.GetShoppingCartItemsV2(customerId);

         var shoppingCartItemsDtos = await shoppingCartItemsToBePurchased
                                     .SelectAwait(async shoppingCartItem => await _dtoHelper.PrepareShoppingCartItemDTOAsyncV2(shoppingCartItem))
                                     .ToListAsync();

         var order = new Order
         {
             CustomerId = customerId,
             BillingAddressId = 1,
             ShippingAddressId = 1,
             OrderGuid = Guid.NewGuid(),
             ShippingStatus = "not_shipped",
             PaymentStatus = "pending",
             OrderTotal = 10000
         };

         var orderItems = shoppingCartItemsDtos.Select(sci => new OrderItem
         {
             OrderItemGuid = Guid.NewGuid(),
             ProductId = sci.ProductId.Value,
             Quantity = sci.Quantity.Value,
             PriceInclTax = sci.Price,
             UnitPriceInclTax = sci.ProductDto.Price
         }).ToList();
         //Prepare data END

         // Create the DataConnection object via using same data provider
         var dataConnectionObject = _nopDataProvider.CreateDataConnection();

         // Create transaction from the data connection object. You may choose any isolation level as per your requirement
         using var transaction = await dataConnectionObject.BeginTransactionAsync(IsolationLevel.ReadCommitted);

         try
         {
             // Perform the insert operations
             var orderId = await dataConnectionObject.InsertWithInt32IdentityAsync(order);
             orderItems.ForEach(each => each.OrderId = orderId);
             await dataConnectionObject.BulkCopyAsync(orderItems);

             // Commit the transaction
             await transaction.CommitAsync();
         }
         catch (Exception)
         {
             // Rollback the transaction if any exception occurs
             await transaction.RollbackAsync();
             throw;
         }
     }
 }
```

With this way, developers may be able to create DataConnection objects and make them shareable through the domain entities to allow transactions. The code is tested and works as expected.

### DIFF
--- a/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
+++ b/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
@@ -27,7 +27,7 @@ public abstract partial class BaseDataProvider
     /// <summary>
     /// Creates the database connection
     /// </summary>
-    protected virtual DataConnection CreateDataConnection()
+    public virtual DataConnection CreateDataConnection()
     {
         return CreateDataConnection(LinqToDbDataProvider);
     }
@@ -37,7 +37,7 @@ public abstract partial class BaseDataProvider
     /// </summary>
     /// <param name="dataProvider">Data provider</param>
     /// <returns>Database connection</returns>
-    protected virtual DataConnection CreateDataConnection(IDataProvider dataProvider)
+    public virtual DataConnection CreateDataConnection(IDataProvider dataProvider)
     {
         ArgumentNullException.ThrowIfNull(dataProvider);
 
--- a/src/Libraries/Nop.Data/DataProviders/MySqlDataProvider.cs
+++ b/src/Libraries/Nop.Data/DataProviders/MySqlDataProvider.cs
@@ -26,7 +26,7 @@ public partial class MySqlNopDataProvider : BaseDataProvider, INopDataProvider
     /// <summary>
     /// Creates the database connection
     /// </summary>
-    protected override DataConnection CreateDataConnection()
+    public override DataConnection CreateDataConnection()
     {
         var dataContext = CreateDataConnection(LinqToDbDataProvider);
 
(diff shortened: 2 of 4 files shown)


skoshelev:
Hi, @TekyaygilFethi. Thank you very much for your suggestion.

We intentionally hid access directly to receiving DataConnection, since its incorrect use can lead to problems that will be very difficult to find.

You can implement IDataProvider for your tasks, where you can easily access the necessary properties.

You can also use the [TransactionScope](https://learn.microsoft.com/en-us/dotnet/api/system.transactions.transactionscope?view=net-8.0) class, as in [this](https://github.com/nopSolutions/nopCommerce/blob/release-4.70.0/src/Libraries/Nop.Data/EntityRepository.cs#L130) method

PR #7352, opened 2024-09-30
Various fixes for mysql-docker-compose.yml

os11k:
Let's use MySQL 8, additionally added volume for MySQL for persistent data and restart always for nopcommerce container

### DIFF
--- a/mysql-docker-compose.yml
+++ b/mysql-docker-compose.yml
@@ -3,17 +3,24 @@ services:
     nopcommerce_web:
         build: .
         container_name: nopcommerce
+        restart: "always"
         ports:
             - "80:80"
         depends_on:
             - nopcommerce_database
+        volumes:
+          - nopcommerce_data:/app
     nopcommerce_database:
-        image: "mysql:latest"
+        image: "mysql:8.0-debian"
         container_name: nopcommerce_mysql_server
         command: "--default-authentication-plugin=mysql_native_password"
         restart: "always"
         environment:
           MYSQL_ROOT_PASSWORD: "nopCommerce_db_password"
+        volumes:
+          - mysqldb:/var/lib/mysql
+
 
 volumes:
-  nopcommerce_data:            
+  nopcommerce_data:
+  mysqldb:


AndreiMaz:
@os11k Why should it be 8.0 and not the latest?

os11k:
Hi @AndreiMaz! First of all, the latest version is 9.0 now and it will change overtime quite often... Version 9.0 isn't even LTS and was just recently released. I think it's always a good idea to include some kind of version number because there are usually significant changes between versions. In this case, for example, the default-authentication-plugin has been removed from MySQL 9 and 8.4. Personally, I think we should stick with 8.0 for now.

skoshelev:
Hi @os11k!

I understand that these changes may be important and necessary for you. However, all the configuration files for docker-compose are provided exclusively in the minimal configuration for test deployment. I am sure no one will use the default file to deploy production infrastructure. For example, I would change the default password. Therefore, changing and adding other parameters should no longer cause difficulties.

Thank you very much for your suggestions, but I think they will be excessive for our task.

os11k:
Hi @skoshelev , main problem is that current docker-compose doesn't work at all... You can't run it, mysql doesn't starts it restarts in the loop because it doesn't know about parameter "default-authentication-plugin", sorry if I didn't explain well from first try. If it would start I would not probably even considered to create PR.

skoshelev:
@os11k, Thank you for reporting the issue, I have created a task accordingly.

os11k:
I would like to add that it will be nice if you can add persistence for MySQL container and nopcommerce_web container, because in reality even in dev environment you do need data persistent... Additionally probably you need to add `restart: "always"` for nopcommerce_web container, because after initial setup is done, it restarts nopcommerce_web service and without `restart: "always"` it doesn't starts at all and you finish basically broken state... 

Long story short, my small changes fixes all of those small things, what it really helps even if you run this in development only...

Not sure why you so against those changes and not sure about your process, but simple merge of my changes will fix that, without need to open any issue.

@AndreiMaz @skoshelev

skoshelev:
@os11k Your changes may be useful or may not, it all depends on the usage scenario. Again, we deliberately did not overload the configuration, leaving it as simple as possible.

Also, if you run the site through Visual Studio, you'll have to start the site manually, each time after restarting through UI. We deliberately do not enable permanent restart so that the user does not miss the moment when something goes wrong with the container at the debugging stage.

Limiting a specific version of MySql will only delay the solution of the problem, but will not solve it, so I created a task to find a permanent solution.

os11k:
Again we can discuss MySQL version separately and I will explain it below, but for dev environment you must have persistent data, and if you don't why you even created `nopcommerce_data:` volume if it is not used anywhere, that doesn't look great... Additionally `restart: "always"`, will not hurt to nopcommerce_web container container, why you don't want it there if it is already for MySQL container.

Additionally I don't get your argument about overloading the configuration, it is just adds 4 lines of code and makes completely different impression for your software, because now it all works, from first start.  But if you don't want that, fine for me.

Regarding MySQL version, I'm not MySQL admin, but there is a reason why other projects do tie their project to specific version, just because every version provides some breaking changes which will not be compatible with older version and it is always a good design to have some major version configured. Using "latest" version of MySQL container is just bad solution.

P.S. I'm fine if you don't accept my changes and frankly I don't really care. Problem is that one user who wanted to use your solution came to me because he wasn't able to use this docker-compose file out of the box and he was not so techie as me, I helped him to fixed that and I made as less changes to your file as possible and I hoped that it would help others, that why I created this PR. But seems you are not appreciating any PRs, you don't appreciate other works, because I did spend time on this, I did tried to run and make it work, I did test it. This in my opinion is not a great look for any open source project.

Have a nice day, bye!

PR #7362, opened 2024-10-07
Pending Approve code customized

GOSRRD:
Checking


AndreiMaz:
Thanks a lot! But we do not plan to add new order statuses out of the box and leave it for customization

PR #7367, opened 2024-10-12
Isssue#7339- (SOLVED) FooterModel.IsHomePage always returns false if SeoFriendlyUrlsForLanguagesEnabled

Aarize29:
Issue#7339
Earlier Equals method was used which definitely cannot return true if the urls are different. Instead of Equals I have used StartsWith method and swap the place of _webHelper.GetThisPageUrl(false) and _webHelper.GetStoreLocation()


exileDev:
In your case IsHomePage always returns true, right?

PR #7454, opened 2024-12-12
fix 7453 Update CatalogModelFactory.cs

### LINKED ISSUE #7453: nop 4.80 all subcategories load wrong image
nopCommerce version: 4.80

all subcategories load wrong image. problem found, PR inbound.

danFbach:
fixes #7453 
currently loads parent category picture.
Fixes reference so that child category picture is loaded instead of parent.


skoshelev:
Hi @danFbach. 
Thank you very much for your help

PR #7466, opened 2024-12-19
Sendinblue/Brevo Marketing Automation error

jigarsangoi:
Hello,

When we enabled guest checkout and start using sendinblue or brevo then we start getting exception in logs with error message "Marketing Automation error". This exception raised because email is empty in customer table when user have role "guest" & not email registered in system. 
To handle guest customers exception we can take email from billing address in OrderCompleted event when customer's email is empty. 
Please apply solution in next version if you approve PR


jigarsangoi:
Please delete this pull request. Its not created properly. Sorry for trouble.

PR #7480, opened 2025-01-02
Update PictureService.cs - NullReferenceException remedy

danFbach:
This is an edge case - my plugin generates invoices in a background task, in which the HttpContext is null, which produces `System.NullReferenceException: Object reference not set to an instance of an object.` 
`HttpContext` should have a null check just like `Request` does and this should not have any negative side effects, since null is just converted to `string.Empty` anyway.


danFbach:
It looks like inside of `WebHelper` this check is performed using the internal`IsRequestAvailable()` method on many of the helper functions.

skoshelev:
Hi, @danFbach!
many thanks for your contributions.

Issue #7471, opened 2024-12-29
Installation (sample data). Some category pictures are missing


AndreiMaz:
nopCommerce version: 4.80.1

Steps to reproduce the problem: 
1. Install nopCommerce
2. The following category pictures are missing: clothing, shoes, cell-phones, camera-photo

Check log for more details: 
Error text: Cannot decode picture binary (file name: clothing)
Full message
System.ArgumentNullException: Value cannot be null. (Parameter 'codec')
   at SkiaSharp.SKBitmap.Decode(SKCodec codec)
   at SkiaSharp.SKBitmap.Decode(ReadOnlySpan`1 buffer)
   at SkiaSharp.SKBitmap.Decode(Byte[] buffer)
   at Nop.Services.Media.PictureService.ValidatePictureAsync(Byte[] pictureBinary, String mimeType, String fileName)

PR #7482, opened 2025-01-03
In ValidatePictureAsync I added condition to check if image is not nu…

NishantKumar-2005:
…ll if it is not null then it convert the jpeg file to jpg file SKBitmap.Decode method might not recognize the file format due to incorrect MIME type or corrupted image data.
fixes : #7471 
image


skoshelev:
Hi @NishantKumar-2005.

I have analyzed your code. I am very sorry, even if I fix everything according to your screenshot (since the code suggested in the PR does not compile), it will not solve the main problem.

Anyway, thank you for your active participation in the development.

Issue #6874, opened 2023-10-11
Advanced newsletter


AndreiMaz:
**Multiple newsletter lists**
Currently our customers have only one option - subscribe or unsubscribe. 

Let's allow a store owner to add multiple newsletter subscription options instead of the single default option enabling customers to choose which newsletters they are subscribed to. This way a customer will be asked to subscriber to multiple lists. For example, "Partner newsletter","Special Offers", "Promo newsletter", etc. All lists will be managed by a store owners. So he will be able to create as many lists as he wants. And in this case (if multiple lists are configured) a customer will see all of them on the registration page or on the customer info page. So he'll be able to choose (with checkboxes) what newsletter lists he wants to subscribe to. Also add an option to choose "All newsletter lists"

All campaigns will also have an option to choose "newsletter list" that will be used. Also a store owner will be able to choose "All newsletter lists" if he wants to send this campaign to all customers (that has at least one list chosen). Also we have to support it in our email plugins (e.g. SendInBlue)

We should allow to filter by newsletter lists in admin area > promotions > Newsletter subscribers page. Also display this column in the grid (not editable). Support it in export/import.

Allow a store owner to configure lists (Name, Ticked by default or not). Once it's done we can remove appropriate customer form fields ('Newsletter' enabled, Newsletter ticked by default)

Original issue in the open-source repository: https://github.com/nopSolutions/nopCommerce/issues/2126

**Newsletter popup**
Currently we allow customers to subscribe on the newsletter block (displayed in the footer), during registration or on the customer info page. We should help store owners to reach to as many potential customers as possible. And we the most efficient approach is newsletter popups.

RomanovM:
See also:
#274
#591
#657
#2126
#4301

DmitriyKulagin:
Closed #6874

PR #7488, opened 2025-01-09
Add newsletter subscription data to Excel export

aniket-sainii:
This pull request adds the functionality to include the newsletter subscription status for each store in the Excel export of customer data. It ensures that for each store, the customer's subscription status is dynamically included in the export based on the existing logic in the system.

Changes:
1. Fetches all stores:
    - The stores are fetched using the ( _storeService.GetAllStoresAsync ) method.
2. Dynamically fetches subscription data:
    - The customer's subscription status is fetched for each store using the 
      ( _newsLetterSubscriptionService.GetNewsLetterSubscriptionByEmailAndStoreIdAsync) method.
    - The subscription status is then displayed as a column for each store in the Excel export.

### DIFF
--- a/src/Libraries/Nop.Services/ExportImport/ExportManager.cs
+++ b/src/Libraries/Nop.Services/ExportImport/ExportManager.cs
@@ -2034,7 +2034,8 @@ public partial class ExportManager : IExportManager
 
             return stateProvince?.Name ?? string.Empty;
         }
-
+        // Fetch all stores from the service
+        var stores = await _storeService.GetAllStoresAsync();
         //property manager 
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @aniket-sainii! Thank you for being so helpful. 
I looked at the implementation, I have no objections. I only left a small comment regarding the comments accompanying the code. 

However, since we added new data to the file for export, we need to do an import of this data as well

AndreiMaz:
Let's wait until the following work item is completed - https://github.com/nopSolutions/nopCommerce/issues/6874

PR #7504, opened 2025-01-21
Optimized the SettingsService by removing a redundant repository call

atiq-bs23:
unnecessary repository call removed


skoshelev:
Hi, @atiq-bs23. 

Many thanks for your help

PR #7507, opened 2025-01-22
Use ToListAsync for async query execution

atiq-bs23:
Replaced the synchronous `ToList()` method with `ToListAsync()` in the GetProductManufacturersByProductIdAsync method within ManufacturerService

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/ManufacturerService.cs
+++ b/src/Libraries/Nop.Services/Catalog/ManufacturerService.cs
@@ -418,7 +418,7 @@ public partial class ManufacturerService : IManufacturerService
             query = query.Where(pm => manufacturersQuery.Any(m => m.Id == pm.ManufacturerId));
         }
 
-        return await _staticCacheManager.GetAsync(key, query.ToList);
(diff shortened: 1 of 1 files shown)


src/Libraries/Nop.Services/Catalog/ManufacturerService.cs
```diff
@@ -418,7 +418,7 @@ join m in _manufacturerRepository.Table on pm.ManufacturerId equals m.Id
             query = query.Where(pm => manufacturersQuery.Any(m => m.Id == pm.ManufacturerId));
         }
 
-        return await _staticCacheManager.GetAsync(key, query.ToList);
+        return await _staticCacheManager.GetAsync(key, query.ToListAsync);
```
skoshelev: In this case, it is better to use the async () => await query.ToListAsync() construction.

atiq-bs23: @skoshelev already done

skoshelev: Thank you very much for your work

PR #7509, opened 2025-01-23
Optimize Reward Points Queries with Asynchronous Execution

atiq-bs23:
Refactored query executions in `RewardPointService` to use asynchronous operations (`ToListAsync`) instead of synchronous calls (`ToList`). This improves performance when processing reward points.


skoshelev:
Hi, @atiq-bs23.

Thank you for your work

PR #7512, opened 2025-01-24
Refactor IsEntityAclMappingExistAsync for Async Query Execution

atiq-bs23:
Updated `IsEntityAclMappingExistAsync` method to use query.AnyAsync for improved async query execution.

### DIFF
--- a/src/Libraries/Nop.Services/Security/AclService.cs
+++ b/src/Libraries/Nop.Services/Security/AclService.cs
@@ -69,8 +69,8 @@ public partial class AclService : IAclService
         var key = _staticCacheManager.PrepareKeyForDefaultCache(NopSecurityDefaults.EntityAclRecordExistsCacheKey, entityName);
 
         var query = from acl in _aclRecordRepository.Table
-                    where acl.EntityName == entityName
-                    select acl;
+            where acl.EntityName == entityName
+            select acl;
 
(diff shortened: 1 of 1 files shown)


src/Libraries/Nop.Services/Security/AclService.cs
```diff
@@ -72,7 +72,7 @@ protected virtual async Task<bool> IsEntityAclMappingExistAsync<TEntity>() where
             where acl.EntityName == entityName
             select acl;
 
-        return await _staticCacheManager.GetAsync(key, query.Any);
+        return await _staticCacheManager.GetAsync(key, ()=>query.AnyAsync());
```
skoshelev, line 75: In this case, it is better to use the async () => await query.AnyAsync() construction.

atiq-bs23: @skoshelev updated

skoshelev:
Hi @atiq-bs23. Thank you very much for your work and such promptness in making edits. However, your last commit included unnecessary changes in the markup, please roll them back so that I can merge your code into the core.

atiq-bs23:
@skoshelev I just ran the visual studio's default code cleanup. Should I revert it?

image

atiq-bs23:
@skoshelev Code is now updated. Thanks

Issue #7520, opened 2025-01-27
Search paging doesn't work


RomanovM:
nopCommerce version: 4.80.2

> If you go to the advanced search page and Search keyword:"gift",  4 come back.  If you then change the "Display per page" from 6 to 3, it comes back with "No products were found that matched your criteria."
> There are no errors in the System > Log.

Source: https://www.nopcommerce.com/boards/topic/101773/page-search-error

Probably related to #7308

exileDev:
Closed #7520

exileDev:
Closed #7520

PR #7523, opened 2025-01-28
#7520 fixed problem of 'Search paging doesn't work'

qasas145:
fixing #7520 problem , where any change in the page size results no products appear , and this is because we were sending post request and the data in the post request be in the body so here 
`        var isSearchTermSpecified = _httpContextAccessor.HttpContext.Request.Query.ContainsKey("q");
`
we try to get the search key from the query and this is not valid with post request so we must check first whether request method was post or get 
and if it was post then we must search from the `q` in the form not query like this 
`var request = _httpContextAccessor.HttpContext.Request;

        bool isSearchTermSpecified = false;
        if (request.Method == HttpMethod.Post.Method)

            isSearchTermSpecified = request.Form.ContainsKey("q");

        else if (request.Method == HttpMethod.Get.Method)
            isSearchTermSpecified = request.Query.ContainsKey("q");
`


RomanovM:
@qasas145 Thanks. We've fixed it the same way, see https://github.com/nopSolutions/nopCommerce/commit/0a97f00fbd0f8f2aacfdf3dd90c08cd5c85140c1

qasas145:
> @qasas145 Thanks. We've fixed it the same way, see [0a97f00](https://github.com/nopSolutions/nopCommerce/commit/0a97f00fbd0f8f2aacfdf3dd90c08cd5c85140c1)

Ok, thank you

Issue #7515, opened 2025-01-24
Add Name Search to Product Attribute Admin Page


chrisjainsley:
The Product Attributes admin page doesn’t have a search box to filter by name. This is a big issue for stores with lots of attributes (e.g., thousands). Scrolling through or manually filtering takes forever.

For stores with tons of attributes, this will save a lot of time and frustration.

Is there a specific reason this hasn’t been added yet? Was it left out by design, or has it just never come up?

I’m willing to implement this and submit a PR if this works for you!

exileDev:
Closed #7515

PR #7524, opened 2025-01-28
#7515 allowing search feature by 'Name' on the product attribute page…

qasas145:
… (admin area)
#7515  allowing search feature by 'Name' on the product attribute page

- I have updated the `List.cshtml` to add a search box so that we can filter product attributes by name 
- and also updated `IProductAttributeService.cs` to include a `searchModelName` as an optional parameter in the method `GetAllProductAttributesAsync`
- also updated `ProductAttributeService.cs` to use this `searchModelName` in the `GetAllProductAttributesAsync` method
- and also a added `SearchModelName` property to `ProductAttributeSearchMode.csl` so that we can filter product attribute by it

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/IProductAttributeService.cs
+++ b/src/Libraries/Nop.Services/Catalog/IProductAttributeService.cs
@@ -27,13 +27,14 @@ public partial interface IProductAttributeService
     /// <summary>
     /// Gets all product attributes
     /// </summary>
+    /// <param name="searchModelName">search model name</param>
     /// <param name="pageIndex">Page index</param>
     /// <param name="pageSize">Page size</param>
     /// <returns>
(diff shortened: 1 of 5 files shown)


sateeshmunagala:
It would be good If the search feature is available for Specification attribute options as well. 
We have 3000 specification attribute options . So search feature is very helpful.
Request you consider creating a PR for the same

qasas145:
> It would be good If the search feature is available for Specification attribute options as well. We have 3000 specification attribute options . So search feature is very helpful. Request you consider creating a PR for the same

Ok thank you

qasas145:
> So search feature is very helpful
see this :
https://github.com/nopSolutions/nopCommerce/pull/7546
this a PR in which i added searching by group name and also by specification name

PR #7526, opened 2025-01-29
IsCustomerShoppingCartEmpty method in ShoppingCartService with async database query execution

atiq-bs23:
Refactored the `IsCustomerShoppingCartEmpty` method in `ShoppingCartService` to use `AnyAsync` for async database query execution

### DIFF
--- a/src/Libraries/Nop.Services/Orders/ShoppingCartService.cs
+++ b/src/Libraries/Nop.Services/Orders/ShoppingCartService.cs
@@ -204,7 +204,7 @@ public partial class ShoppingCartService : IShoppingCartService
     /// <returns>Result</returns>
     protected virtual async Task<bool> IsCustomerShoppingCartEmptyAsync(Customer customer)
     {
-        return !(await _sciRepository.Table.AnyAsync(sci => sci.CustomerId == customer.Id));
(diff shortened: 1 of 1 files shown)


src/Libraries/Nop.Services/Orders/ShoppingCartService.cs
```diff
@@ -202,9 +202,9 @@ protected virtual async Task<bool> ShoppingCartItemIsEqualAsync(ShoppingCartItem
     /// </summary>
     /// <param name="customer">Customer</param>
     /// <returns>Result</returns>
-    protected virtual bool IsCustomerShoppingCartEmpty(Customer customer)
+    protected virtual async Task<bool> IsCustomerShoppingCartEmptyAsync(Customer customer)
     {
-        return !_sciRepository.Table.Any(sci => sci.CustomerId == customer.Id);
+        return !(await _sciRepository.Table.AnyAsync(sci => sci.CustomerId == customer.Id));
```
skoshelev, line 207: In this case and in similar calls below, parentheses are not needed, without them, the code will look neater

atiq-bs23: @skoshelev updated

skoshelev:
@atiq-bs23, Thank you very much for your work

PR #7527, opened 2025-01-29
Refactor to use ToDictionaryAsync for async execution

atiq-bs23:
Refactored the `ImportResourcesFromXmlAsync` method in `LocalizationService` to use `ToDictionaryAsync` for efficient async query execution and dictionary creation. Removed unnecessary sorting

### DIFF
--- a/src/Libraries/Nop.Services/Localization/LocalizationService.cs
+++ b/src/Libraries/Nop.Services/Localization/LocalizationService.cs
@@ -503,11 +503,8 @@ public partial class LocalizationService : ILocalizationService
         if (xmlStreamReader.EndOfStream)
             return;
 
-        var lsNamesList = new Dictionary<string, LocaleStringResource>();
-
-        foreach (var localeStringResource in _lsrRepository.Table.Where(lsr => lsr.LanguageId == language.Id)
-                     .OrderBy(lsr => lsr.Id))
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi, @atiq-bs23. Thank you very much for analyzing our code.
Unfortunately, in this case, we cannot apply the ToDictionaryAsync method. If you pay attention to the original code, you will see that we sort locales by identifier. This is done in case there are identical locales in the DB (yes, in the current nopCommerce implementation it is almost impossible, but in previous versions, it could happen, plus locales can be loaded into the database from external sources). When using ToDictionaryAsync, we will get an exception that the key already exists, as soon as we come across the first repetition of the locale.

atiq-bs23:
@skoshelev Got it. Thanks for the clarification

atiq-bs23:
@skoshelev 
What about async execution like below code?

```
var locales = await _lsrRepository.Table
    .Where(lsr => lsr.LanguageId == language.Id)
    .OrderBy(lsr => lsr.Id)
    .ToListAsync();

foreach (var localeStringResource in locales)
{
    lsNamesList[localeStringResource.ResourceName.ToLowerInvariant()] = localeStringResource;
}
```

skoshelev:
@atiq-bs23 Yes, it will be possible to use it

atiq-bs23:
@skoshelev Please review now. Thanks

skoshelev:
@atiq-bs23 Thanks again

PR #7536, opened 2025-01-30
7529 - Infinite recursion in product import

fuadwasi:
Fix the method SplitProductFile in ImportManager. 
image

### DIFF
--- a/src/Libraries/Nop.Services/ExportImport/ImportManager.cs
+++ b/src/Libraries/Nop.Services/ExportImport/ImportManager.cs
@@ -2042,7 +2042,7 @@ public partial class ImportManager : IImportManager
         var metadata = await PrepareImportProductDataAsync(workbook, languages);
         var defaultWorksheet = metadata.DefaultWorksheet;
 
-        if (_catalogSettings.ExportImportSplitProductsFile && metadata.CountProductsInFile > _catalogSettings.ExportImportProductsCountInOneFile)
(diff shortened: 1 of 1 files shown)


src/Libraries/Nop.Services/ExportImport/ImportManager.cs
```diff
@@ -1454,14 +1454,15 @@ protected virtual IList<string> SplitProductFile(IXLWorksheet worksheet, ImportP
         var endCell = metadata.Properties.Max(p => p.PropertyOrderPosition);
 
         var filePaths = new List<string>();
+        var exportImportProductsCountInOneFile = Math.Max(_catalogSettings.ExportImportProductsCountInOneFile, 1);
```
skoshelev, line 1457: I don't understand why these changes are needed, even if we imagine that the user enters a value less than 1 in the ExportImportProductsCountInOneFile setting, then splitting the file by one product does not seem like a reasonable idea.

fuadwasi: I added this to prevent the infinite recursion in the worst case. Can set a Static minimum value and use that.

skoshelev: So, is this protection against someone entering a value less than 1? If so, this is excessive reinsurance, I don't see any point in someone entering a value less than 500.

fuadwasi: Ok @skoshelev I have removed that checking. Can you please check now?

Issue #7545, opened 2025-02-02
Allow to search by "Name" on the specification attribute page (admin area)


qasas145:
The Speficaition Attributes admin page doesn’t have a search box to filter by name. This is a big issue for stores with lots of attributes (e.g., thousands). Scrolling through or manually filtering takes forever.

For stores with tons of attributes, this will save a lot of time and frustration.

See PR : #7546 (origially suggested to search in groups as well. but it'll be OK for have searching by "name" only

AndreiMaz:
Origially suggested to search in groups as well. but it'll be OK for have searching by "name" only

qasas145:
> Thanks a lot for suggestion. Origially suggested to search in groups as well. but it'll be OK for have searching by "name" only

Ok, I have updated my PR : #7546 so that we can search only by specification attribute name

exileDev:
Closed #7545

PR #7546, opened 2025-02-02
Allowing search by 'Specification attribute name' in specification attributes page (Admin area)

qasas145:
#7545 i have added this feature by these steps 

- update `GetSpecificationAttributeGroupsAsync` in the `ISpecificationAttributeService.cs` file to have this signature `GetSpecificationAttributeGroupsAsync(int pageIndex = 0, int pageSize = int.MaxValue, string specificationName = "")` and this is to enable search by group name which contains this specification attribute 
- update `GetSpecificationAttributesByGroupIdAsync` to `GetSpecificationAttributesByGroupIdAsync(int? specificationAttributeGroupId = null, string name = "")` to enable searching by attribute name in the group 
- update `PrepareSpecificationAttributeGroupListModelAsync` in the `SpecificationAttributeModelFactory` 
- update `List.cshtml` to include search box 
- update `SpecificationAttributeSearchModel.cs` to include add searchName property 
- update `SpecificationAttributeGroupSearchModel.cs` to include specification attribute search property 

**So this will be the page** 

**You can type the attribute name and it will fetch the attribute and the group of it , and if it has no group then it appear in the default group**

**_And you don't have to click on the control icon to display the result attributes , but the child table of specification attributes will open auto, so that you will see the attribute result directly_** 

**See images below**

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/ISpecificationAttributeService.cs
+++ b/src/Libraries/Nop.Services/Catalog/ISpecificationAttributeService.cs
@@ -25,13 +25,12 @@ public partial interface ISpecificationAttributeService
     /// </summary>
     /// <param name="pageIndex">Page index</param>
     /// <param name="pageSize">Page size</param>
-    /// <param name="gName">group name</param>
-    /// <param name="sName">specification attribute name name</param>
+    /// <param name="specificationName">specification attribute name name</param>
     /// <returns>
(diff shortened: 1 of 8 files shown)


qasas145:
@RomanovM ,I Have updated this is PR , so we can search only by specification attribute name

PR #7547, opened 2025-02-03
Optimize Get All Language Query by Applying Store Mapping at Database Level

atiq-bs23:
This PR improves language retrieval by applying store mapping constraints directly in the database query. This eliminates in-memory filtering.

### DIFF
--- a/src/Libraries/Nop.Services/Localization/LanguageService.cs
+++ b/src/Libraries/Nop.Services/Localization/LanguageService.cs
@@ -81,20 +81,18 @@ public partial class LanguageService : ILanguageService
 
         var languages = await _staticCacheManager.GetAsync(key, async () =>
         {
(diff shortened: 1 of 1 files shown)


exileDev:
In fact, initially languages are cached in the GetAllAsync method, and then using the store ID

atiq-bs23:
@exileDev Thanks for your input! Actually, caching happens in `GetAllLanguagesAsync` using `_staticCacheManager.GetAsync`. In GetAllAsync, since getCacheKey is null by default, it directly queries the DB:
```
if (getCacheKey == null)
    return await getAllAsync();
```
So, the cache is applied at the higher level, not inside `GetAllAsync`.

atiq-bs23:
@skoshelev I really love contributing here. You guys are very positive and truly appreciate contributors' efforts. It’s always a great experience.

PR #7554, opened 2025-02-05
Optimize Product Tag Mappings Deletion in ProductTagService

atiq-bs23:
Refactored `UpdateProductTagsAsync` in `ProductTagService` to improve performance by batch deleting product tag mappings. This reduces multiple DB calls to a single query, enhancing efficiency.


skoshelev:
Hi @atiq-bs23. 
I looked at the commits, thank you very much for the suggestion, I like this optimization. But there are a couple of points:
We try not to use HashSet where it is not really necessary, it is better to use IList in this code.
You changed the input parameter of the method from productTagId to productTagIds, but did not change the documentation line to the method, it still remains <param name="productTagId">Product tag identifier</param>

atiq-bs23:
@skoshelev updated, please review now.

Issue #7495, opened 2025-01-14
Don't display the "Purchased with orders" block (the product edit page) when the current user does not have a "View orders" permission


AndreiMaz:
nopCommerce version: 4.80.1

We should not display the "Purchased with orders" block (the product edit page) when the current user does not have a "View orders" permission. Let's just hide it

PR #7581, opened 2025-02-14
Hide "Purchased with orders" block for users without "View orders" permission

atiq-bs23:
The "Purchased with orders" block on the product edit page is hidden when the current user does not have the "View orders" permission.


atiq-bs23:
#7495 related

atiq-bs23:
@skoshelev based on your below commit, I have updated my PR as well.

https://github.com/nopSolutions/nopCommerce/commit/7398bb278482d74b270671372c1a09f9cf349f68

skoshelev:
@atiq-bs23 Thank you for your contribution

PR #7583, opened 2025-02-14
Optimize DeleteProductAttributesAsync by using bulk delete

atiq-bs23:
Replaced the deletion in loop with a bulk delete operation to improve performance. This reduces the number of database calls.

Other changes is for improve code readability.


skoshelev:
@atiq-bs23 Thank you very much for your help

Issue #7580, opened 2025-02-14
Don't display the "Shipment List" block (Oder edit page) when the current user does not have a "View Shipments" permission


atiq-bs23:
![Image](https://github.com/user-attachments/assets/f6dbc285-f63b-4da5-8641-7bf88479069e)

![Image](https://github.com/user-attachments/assets/f84c09ef-ea85-4b48-a6ec-d7ec4d753ec2)

atiq-bs23:
#7495 issue is similar

PR #7585, opened 2025-02-17
Hide "Shipments" block for users without "View Shipments" permission on the order edit page

atiq-bs23:
The "Shipments" block on the order edit page is hidden when the current user does not have the "View Shipments" permission.


atiq-bs23:
#7580

skoshelev:
Hi, @atiq-bs23.
Good job, thank you.

PR #7591, opened 2025-02-18
Fix: correct index existence check for database migration in v4.9.0

damiantasevski:
This bug caused a database exception because the code attempted to create an index that already existed.


exileDev:
@damiantasevski Thanks!

Issue #7422, opened 2024-11-20
Add new PostCopyProductEvent event


skoshelev:
We can add a new event PostCopyProductEvent. This will allow plugins to handle the creation of a product based on copying. 
Let the event contain two products: the original product, and the created copy

PR #7601, opened 2025-02-23
Add new PostCopyProductEvent event

atiq-bs23:
new post copy product event added with the new product and the original product.


atiq-bs23:
@skoshelev  issue #7422

skoshelev:
Hi, @atiq-bs23!
Thank you for your help

PR #7631, opened 2025-03-14
fix: use proper search model

dotnetdreamer:


RomanovM:
@dotnetdreamer Thanks for your help

Issue #7654, opened 2025-04-01
Change the ProductPriceModel.PriceWithDiscount property to PriceWithoutDiscount property


skoshelev:
The Price property itself already assumes that it reflects the lowest possible price, that is, already with all discounts, so there is no need for a separate PriceWithDiscount property. It is better to create PriceWithoutDiscount to store the price without a discount.

Also, do not forget to change the use of these properties on the views accordingly.

KaranChadha10:
I've created a PR (https://github.com/nopSolutions/nopCommerce/pull/7656) to address this issue. Please review when possible. Thanks!

PR #7656, opened 2025-04-02
refactor: renamed PriceWithDiscount to PriceWithoutDiscount

KaranChadha10:
### Description

- Renamed the `PriceWithDiscount` to `PriceWithoutDiscount` in the `ProductPriceModel.cs` class.
-  Updated the property name in `_ProductPrice.cshtml` view.
- Updated the model property in `ProductModelFactoryTests.cs`.

Issue Link: https://github.com/nopSolutions/nopCommerce/issues/7654

### DIFF
--- a/src/Presentation/Nop.Web/Factories/ProductModelFactory.cs
+++ b/src/Presentation/Nop.Web/Factories/ProductModelFactory.cs
@@ -488,7 +488,7 @@ public partial class ProductModelFactory : IProductModelFactory
 
         if (finalPriceWithoutDiscountBase != finalPriceWithDiscountBase)
         {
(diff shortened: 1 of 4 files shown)


skoshelev:
Hi @KaranChadha10. 

Thank you very much for your willingness to help. Sorry for not looking at your code for a long time. 

Unfortunately, the essence of the ticket is a bit more complicated than just renaming one field. At a minimum, you also need to rename the PriceWithDiscountValue field. But even that's not all, the value in the new field also needs to be set with another variable. Also, the location of the field display on the view should change, as well as the locale. 

Unfortunately, I can't accept your PR in this form. If you want to continue working on the task, then let me know

PR #7660, opened 2025-04-07
Unnecessary duplicate code remove

atiq-bs23:
Removed unnecessary duplicate code


RomanovM:
@atiq-bs23 Thanks for your help

Issue #7678, opened 2025-04-15
Incorrect file size for Category image, Download file and Sample, additional query request overhead.


jafin:
# ⚠️ Issue: Incorrect File Size for Category Image, Download File, and Sample File, Plus Unnecessary Query Overhead

The Category image, Download file, and Sample file components each attempt to retrieve file information by making a request to their respective controller endpoints. However, these requests return HTML rather than the actual file, since the controller is not designed to serve raw file metadata.

Despite this, the FilePond control uses the HTML response size as the file size, which leads to an incorrect display of file size in the UI.

You can observe this behavior by inspecting network traffic in the Admin panel. Notice that FilePond makes GET requests to controller URLs, and the size shown corresponds to the size of the HTML page, not the actual uploaded file.

This file size is actually the HTML content size, NOT the uploaded file's size.
![Image](https://github.com/user-attachments/assets/15a71fcc-1fe6-422a-8b37-e102a13a34f0)

Capture of FilePond query to controller, with size of HTML returned.
![Image](https://github.com/user-attachments/assets/abc05da2-9a55-40d8-9870-7ae217d841a3)

## 💡 Proposal: Populate File Size on Initial Server Render

FilePond supports pre-populating its files array with metadata, including the file size. By providing this metadata during server-side rendering, we can:

- Avoid redundant controller requests
- Ensure correct file size is displayed
- Reduce overall server traffic (performance benefit)
- Eliminate misleading GET requests for HTML content

This approach works well if the file is stored in a database (e.g., as a Binary), where its size is readily available.

However, if a third-party file provider (e.g., AWS S3) is used, the Binary property may be null, and fetching the file just to determine size could be expensive — especially for large files (e.g., 100MB+).

To solve this efficiently, introduce a way to cache the file size at the time of upload. Two possible approaches:

- Add a FileSize column to the Download and Picture tables.
- Add a serialized Attributes JSON field that stores extensible metadata (e.g., FileSize, MimeType, etc.).

The latter provides future flexibility for 3rd parties and avoids needing schema changes for every new attribute. It can be deserialized into a dictionary as needed, without requiring native JSON SQL support.
The former is simple and easy to implement.  The only complexity may be populating it for existing files, if they could be ignored the change would be trivial, and show 0bytes for sizes of existing (which is no worse that what is occurring now with incorrect size being shown.)

## No decimal in the Category Picture control.

There is no decimal between filename and extension as seen below.

[...]

PR #7679, opened 2025-04-15
Improvement for Incorrect file size for Category image, Download file and Sample, additional query request overhead. #7678

jafin:
Proposal for #7678 

### Improvement to FileSize for Category, Download and Sample

Populate the FilePond schema for 
- Category Image
- Download file
- Sample File

This avoids additional queries against the respective controllers returning the HTML size to the control (which is incorrect), since the files array for FilePond is already popualted with this content.

### Refactoring front end JS
- Improved separation of backend variables and clientside JS by introducing js viewModel, and using JSON Serializing to ensure correct response.
- Move `GetAcceptedMimeTypes` to `MediaSettings` to avoid duplicate code (in future this could be added to settings to allow user to override I guess, for now readonly.
- Minor refactorings to JS,  
  - Jquery `.click` -> `on('click',fn())`,  
  -  Create jquery Selectors once, and reuse.

__NOTE:__ Depending on the configuration of the `PictureService` or `DownloadService` (via 3rd parties) the fileSize of the attachments may NOT be populated.  It is not guaranteed this value is set, nor for performance should it be.  See proposal in #7678 for ideas to introduce a size attribute to db to cache this information (avoiding the need to pull down large objects just to obtain filesize) 

This should be a non breaking change.


exileDev:
Thank you! I've squashed and merged this pull request via 9a8f1ab.

Issue #7659, opened 2025-04-07
Notification use popup Setting is not work properly


pddobariya:
Hello,

I see the notification use popup setting is not working correctly.

**messagessettings.usepopupnotifications** 
This is the setting that I can set to true and false from the admin, but it's not working as expected.

The reason is that the setting is not used in the place.
 
<img width="1195" alt="Image" src="https://github.com/user-attachments/assets/fb571406-d28c-446b-8b46-9589df1ecf71" />

Thanks.

exileDev:
Closed #7659

PR #7680, opened 2025-04-17
Fix popup notification setting usage

atiq-bs23:
`Messagesettings.Usepopupnotifications` was not used — a hardcoded false was passed to JavaScript. This PR removes the hardcoded value and uses the correct setting value.


atiq-bs23:
Issue #7659

exileDev:
@atiq-bs23 Thanks for your suggestion, but we decided to solve this issue more thoroughly.

PR #7684, opened 2025-04-22
Prevent Overwriting of Configuration File by Adding Existence Check

Shubhk209:
Fix: Check if configuration file exists before creating it

Prevent overwriting existing configuration (e.g., appsettings.json). Only create the file if it doesn't already exist.
Ensure default values are used only when the file is missing, keeping user settings intact.


RomanovM:
No need to check this, it's inside INopFileProvider.CreateFile() method

Issue #7687, opened 2025-04-24
Add an event invoked when the application starts


RomanovM:
Sometimes we need to perform some actions when the application starts, for example, install and update plugins, start the task scheduler, etc. At moment we perform all this in the core, but this is not a very flexible solution. 
We could add an event on startup that can be handled anywhere including in plugins.

Make sure the dependency injection will work correctly in this case.

See PR: S

atiq-bs23:
@RomanovM @skoshelev If we publish the event before app.RunAsync, and a consumer (e.g., from a plugin) takes time to complete its work, the application will be delayed in starting to listen for HTTP requests. We should consider this scenario and ensure long-running event handlers do not block app startup.

PR #7690, opened 2025-04-25
Add AppStartedEvent and publish it on application startup using background thread

atiq-bs23:
Adds a new `AppStartedEvent` that is published after the application completes its startup process (`StartEngineAsync`).

- The event is published after key startup steps like plugin installation, DB migrations, and task scheduler initialization.
- Uses `Task.Run` to publish the event in a background thread, ensuring that long-running consumers (e.g., plugins) do not block the app from handling HTTP requests.
- Ensures dependency injection works correctly.

### DIFF
--- a/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
+++ b/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
@@ -19,6 +19,7 @@ using Nop.Core.Configuration;
 using Nop.Core.Domain.Common;
 using Nop.Core.Domain.Customers;
 using Nop.Core.Domain.Localization;
+using Nop.Core.Events;
 using Nop.Core.Http;
 using Nop.Core.Infrastructure;
 using Nop.Data;
@@ -56,7 +57,7 @@ public static class ApplicationBuilderExtensions
         EngineContext.Current.ConfigureRequestPipeline(application);
     }
 
-    public static async Task StartEngineAsync(this IApplicationBuilder _)
+    public static async Task StartEngineAsync(this IApplicationBuilder app)
     {
         var engine = EngineContext.Current;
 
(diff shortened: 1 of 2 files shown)


atiq-bs23:
@skoshelev issue #7687

src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -89,6 +90,21 @@ public static async Task StartEngineAsync(this IApplicationBuilder _)
             //clear payment info requests
             var genericAttributeService = engine.Resolve<IGenericAttributeService>();
             await genericAttributeService.DeleteAttributesAsync<Customer>(NopCustomerDefaults.ProcessPaymentRequestAttribute);
+
+            // Publish AppStartedEvent in background to avoid delaying HTTP request handling
```
skoshelev, line 94: There is one small but very significant problem here. Since you do not restart the server very often, and usually do it during the lowest load, it is better to neglect lost HTTP requests than to face a situation where some data is not initialized at startup, since it will go in parallel.

src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -89,6 +90,21 @@ public static async Task StartEngineAsync(this IApplicationBuilder _)
...
+            var eventPublisher = engine.Resolve<IEventPublisher>();
+            _ = Task.Run(async () =>
+            {
+                try
+                {
+                    await eventPublisher.PublishAsync(new AppStartedEvent());
+                }
+                catch (Exception ex)
+                {
+                    var logger = engine.Resolve<ILogger>();
```
skoshelev, line 104: I think simply logging problems with processing the application start event at this level is also bad, for the same reason as processing the event in parallel

src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
atiq-bs23: @skoshelev Thanks for explaining it in detail.
I agree. 
I have updated the code to publish the event synchronously.

skoshelev:
Hi @atiq-bs23!
Thank you for your help

PR #7718, opened 2025-05-27
We moved the PluginUpdatedEvent trigger to the very end of the POST;

wakku9780:
🛠️ What did we do to fix it?
We moved the PluginUpdatedEvent trigger to the very end of the POST method — after:

The plugin's enabled/disabled logic runs,

All related settings are saved,

Activity log is recorded.

This ensures that the plugin is in its final and consistent state when the event is published, and all event subscribers get accurate information.

- await _eventPublisher.PublishAsync(new PluginUpdatedEvent(pluginDescriptor));
  // (Previously triggered too early)

...

+ await _customerActivityService.InsertActivityAsync(...);
+ await _eventPublisher.PublishAsync(new PluginUpdatedEvent(pluginDescriptor));
  // (Now triggered at the correct place, after all updates)

### DIFF
--- a/src/Presentation/Nop.Web/Areas/Admin/Controllers/PluginController.cs
+++ b/src/Presentation/Nop.Web/Areas/Admin/Controllers/PluginController.cs
@@ -396,7 +396,7 @@ public partial class PluginController : BaseAdminController
             pluginDescriptor.Save();
 
             //raise event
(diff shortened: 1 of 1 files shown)


wakku9780:
Hi @skoshelev skoshelev 
You're welcome! 😊

Happy to contribute and glad the fix was helpful. Let me know if there's anything else I can help with in the future. 

Thanks for maintaining such a great open-source project! 🚀

PR #7722, opened 2025-05-30
To fix this, we should update the LINQ query to preserve the order of Products.

wakku9780:
❓ Problem:
When a custom SearchProvider is used (like Azure Search, ElasticSearch, or any plugin), it may return a list of product IDs in a specific order, representing a ranked result set. However, SearchProductsAsync() currently does not maintain that custom order after filtering.

✅ Proposed Solution
The issue occurs because the SearchProductsAsync method does not apply the order defined by the external SearchProvider when it is available and returns product IDs.

To fix this, we should update the LINQ query to preserve the order of providerResults after retrieving the matching products.

Here’s how we can do it:

path of this code:
[nopCommerce/src/Libraries/Nop.Services/Catalog/ProductService.cs] :

public virtual async Task<IPagedList> SearchProductsAsync():

if (providerResults?.Any() == true)
{
    // Filter products based on SearchProvider results and preserve order
    productsQuery =
        from p in productsQuery
        where providerResults.Contains(p.Id)
        orderby providerResults.IndexOf(p.Id)
        select p;
}

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/ProductService.cs
+++ b/src/Libraries/Nop.Services/Catalog/ProductService.cs
@@ -1073,14 +1073,13 @@ public partial class ProductService : IProductService
                 }
             }
 
-            if (!runStandardSearch && providerResults?.Any() == true)
-            {
-                productsQuery =
-                    from p in productsQuery
-                    join id in providerResults.Select((id, index) => new { id, index }) on p.Id equals id.id
-                    orderby id.index
-                    select p;
-            }
+
+            productsQuery =
+                from p in productsQuery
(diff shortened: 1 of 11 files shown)


danFbach:
sorry, but your solution will not fix the issue.
you also are changing an event publish which is unrelated to the issue.
I have proposed a working solution that I have been using since I opened the issue for the nop team to review.

wakku9780:
Hi     @danFbach
Thanks for the clarification!
I’ll recheck everything once again from my side. 🙏
As you mentioned, the core issue is that the product is not appearing in the order search — I’ll focus on reproducing and verifying exactly that scenario.

Appreciate your input — I’ll update my PR if needed. 👍

PR #7724, opened 2025-05-30
fixes #7705

### LINKED ISSUE #7705: SearchProductsAsync does not respect SearchProvider ordering
If you review SearchProductsAsync in ProductService:
 https://github.com/nopSolutions/nopCommerce/blob/402ed11ee4dea8d9d2904ddb7a0463f24d06f216/src/Libraries/Nop.Services/Catalog/ProductService.cs#L1167-L1180
You can see that the Pagination is applied _before_ the providers ordering/sorting is applied. If you apply a large page size (such that all items are on one page) you will see that sorting is applied correctly, but on smaller page sizes, where results are truncated prior to ordering, the sort order is incorrect.

Something like the following would fix it and maintains ToPageListAsync optimization and function surface.
```csharp

var products = productsQuery.OrderBy(_localizedPropertyRepository, await _workContext.GetWorkingLanguageAsync(), orderBy);

if (providerResults.Any() && orderBy == ProductSortingEnum.Position && !showHidden)
{
    var orderDict = providerResults.Select((value, index) => new { value, index }).ToDictionary(x => x.value, x => x.index);
    products = from item in products
               join order in orderDict on item.Id equals order.Key
               orderby order.Value
               select item;
}

return await products.ToPagedListAsync(pageIndex, pageSize);
```

danFbach:
fixes #7705 with solution proposed in issue.
been running on my own installation since first solution proposed, works as intended.
the Sorting class could be moved somewhere else or perhaps an alternative to using the class is possible.?

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/ProductService.cs
+++ b/src/Libraries/Nop.Services/Catalog/ProductService.cs
@@ -1164,20 +1164,26 @@ public partial class ProductService : IProductService
             }
         }
 
-        var products = await productsQuery.OrderBy(_localizedPropertyRepository, await _workContext.GetWorkingLanguageAsync(), orderBy).ToPagedListAsync(pageIndex, pageSize);
+        var products = productsQuery.OrderBy(_localizedPropertyRepository, await _workContext.GetWorkingLanguageAsync(), orderBy);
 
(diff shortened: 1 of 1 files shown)


exileDev:
@danFbach Thanks, we made some similar changes but with `IQueryable` (see related issue)

PR #7778, opened 2025-07-16
Add support for language exclusion in sitemap.xml

### LINKED ISSUE #7777: Support per-language exclusions in sitemap.xml generation
### Summary

Currently, the `RobotsTxtSettings.DisallowLanguages` setting allows administrators to exclude specific languages from being indexed by robots. However, this exclusion is **only applied in `robots.txt`**, not in `sitemap.xml`.

As a result, languages disallowed in `robots.txt` can **still appear in sitemap entries** (including `<xhtml:link rel="alternate" hreflang="...">`), which may send **conflicting signals to crawlers**.

---

### 💡 Real-world use case

Imagine an e-commerce store that:

- Has a language version `/de` for testing or development purposes
- Blocks `/de` in `robots.txt` to prevent indexing
- Still sees `/de` alternate URLs in `sitemap.xml`

This can cause:
- Google Search Console warnings
- Indexing of alternate versions that should not be public
- Duplicate content penalties across language versions

---

### ✅ Recommendation

Introduce a new setting:

```csharp
SitemapXmlSettings.DisallowLanguages
```

This would allow fine-grained control over which language versions are included in the sitemap, independently of what is excluded from `robots.txt`.

We should **not reuse `RobotsTxtSettings.DisallowLanguages`**, because:
- It may be desirable to **exclude a language from sitemap** but still allow crawlers to access it (e.g., noindex SEO experiments)
- The two systems have separate behavior and timing

---

### 📦 Affected areas

- `SitemapModelFactory`
- `SitemapXmlSettings`

---

### 💪 Next step

I'd be happy to prepare a PR that introduces:

- A new setting `DisallowLanguages` in `SitemapXmlSettings`
- Updates to `SitemapModelFactory` to respect this list when generating entries and alternate `<xhtml:link>` references

bambuca:
This PR adds the ability to exclude specific languages from sitemap.xml generation.

### What’s new:
- New property `DisallowLanguages` in `SitemapXmlSettings`
- `SitemapModelFactory` and alternate URL logic now respects this setting
- Default behavior remains unchanged (no exclusions)

### Why:
In some scenarios, site owners may want to:
- Prevent certain language versions from being listed in sitemaps (e.g. test/staging content)
- Avoid conflicting SEO signals when the same languages are excluded in `robots.txt`

This setting provides fine-grained control and separates concerns between `robots.txt` and sitemap configuration.

Closes #7777

### DIFF
--- a/src/Presentation/Nop.Web/Factories/SitemapModelFactory.cs
+++ b/src/Presentation/Nop.Web/Factories/SitemapModelFactory.cs
@@ -13,10 +13,8 @@ using Nop.Core.Domain.Forums;
 using Nop.Core.Domain.Localization;
 using Nop.Core.Domain.News;
 using Nop.Core.Domain.Seo;
-using Nop.Core.Domain.Stores;
 using Nop.Core.Domain.Topics;
 using Nop.Core.Events;
-using Nop.Core.Http;
 using Nop.Core.Infrastructure;
 using Nop.Services.Blogs;
 using Nop.Services.Catalog;
@@ -524,7 +522,7 @@ public partial class SitemapModelFactory : ISitemapModelFactory
     /// <param name="fullPath">The path and name of the sitemap file</param>
     /// <param name="id">Sitemap identifier</param>
     /// <returns>A task that represents the asynchronous operation</returns>
-    protected virtual async Task GenerateAsync(string fullPath, int id = 0) 
+    protected virtual async Task GenerateAsync(string fullPath, int id = 0)
     {
         //generate all URLs for the sitemap
         var sitemapUrls = await GenerateUrlsAsync();
@@ -605,22 +603,6 @@ public partial class SitemapModelFactory : ISitemapModelFactory
         return new Uri(new Uri(scheme), localizedPath).ToString();
     }
 
-    /// <summary>
-    /// Retrieves the list of languages for the given store that are not excluded in the sitemap XML settings,
-    /// if SEO-friendly URLs for languages are enabled.
-    /// </summary>
-    /// <param name="store">The store to retrieve allowed languages for.</param>
-    /// <returns>
-    /// A list of <see cref="Language"/> if SEO-friendly URLs are enabled; otherwise, <c>null</c>.
-    /// </returns>
-    protected async Task<List<Language>> GetAllowedLanguagesAsync(Store store)
-    {
-        return _localizationSettings.SeoFriendlyUrlsForLanguagesEnabled
-            ? (await _languageService.GetAllLanguagesAsync(storeId: store.Id))
-            .Where(lang => !_sitemapXmlSettings.DisallowLanguages.Contains(lang.Id))
-            .ToList()
-            : null;
-    }
     #endregion
 
     #region Methods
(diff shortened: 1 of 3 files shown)


src/Presentation/Nop.Web/Factories/SitemapModelFactory.cs
```diff
@@ -912,16 +878,26 @@ var name when name.Equals(nameof(ProductTag), StringComparison.InvariantCultureI
             _ => GetUrlHelper().RouteUrl(routeName, values, protocol)
         };
 
-        //url for current language
-        var url = await routeUrlAsync(routeName,
```
skoshelev, line 882: This code does not need to be removed. By default, the main address is the address without localization

skoshelev:
I think [this logic](https://github.com/nopSolutions/nopCommerce/pull/7778/commits/052034e9a37d2d68f363546712f3b9df06cce10c#diff-cfe252349708b9b74f3a8e423d6e1bb524f9d32e15f0c0df24d2afa086a25333R884-R900) is unnecessary in this context, it will be enough to simply filter languages using the DisallowLanguages setting

skoshelev:
Hi @bambuca.

Thank you very much for your help. I left some comments on your code.

src/Libraries/Nop.Core/Domain/Common/SitemapXmlSettings.cs
```diff
@@ -71,4 +71,9 @@ public SitemapXmlSettings()
     /// Gets or sets the wait time (in seconds) before the operation can be started again
     /// </summary>
     public int SitemapBuildOperationDelay { get; set; }
+
+    /// <summary>
+    /// Disallow languages
+    /// </summary>
+    public List<int> DisallowLanguages { get; set; } = new();
```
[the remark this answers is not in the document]
bambuca: Ok, now there is no UI for administrating the SitemapXmlSettings so I will add new section for all settings, not just for this new property.

src/Presentation/Nop.Web/Factories/SitemapModelFactory.cs
bambuca: Thanks for the feedback! However, I'd like to clarify why we replaced this part:

```csharp
var url = await routeUrlAsync(routeName,
    getRouteParamsAwait != null ? await getRouteParamsAwait(null) : null,
    await GetHttpProtocolAsync());
```

with this one:

```csharp
var languages = _localizationSettings.SeoFriendlyUrlsForLanguagesEnabled
    ? (await _languageService.GetAllLanguagesAsync(storeId: store.Id))
        .Where(lang => !_sitemapXmlSettings.DisallowLanguages.Contains(lang.Id))
        .ToList()
    : null;

var workingLanguage = await _workContext.GetWorkingLanguageAsync();
var language = languages?.FirstOrDefault(lang => lang.Id == store.DefaultLanguageId)
    ?? languages?.FirstOrDefault()
    ?? workingLanguage;

var url = await routeUrlAsync(
    routeName,
    getRouteParamsAwait != null ? await getRouteParamsAwait(language.Id) : null,
    await GetHttpProtocolAsync());

if (language.Id != workingLanguage.Id)
    url = GetLocalizedUrl(url, language);
```

### Why this change?

Even though the original version uses `getRouteParamsAwait(null)`, that `null` does **not eliminate language context**. In fact:

- For some entities (like News), `LanguageId` is passed explicitly (e.g., `news.LanguageId`).
- For others, like in `GetSeoRouteParamsAwait(...)`, the `LanguageId` is **not provided**, and the system falls back to the **current language from `IWorkContext`**, resolved in `_urlRecordService.GetSeNameAsync(...)`.

This makes the result dependent on the active user/session, which is undesirable for sitemap generation.

In addition to that, the default `IOutboundParameterTransformer` implementation in `LanguageParameterTransformer.cs` also:
- First checks for the language in `RouteValues`
- Then falls back to `IWorkContext`

This behavior is not fully deterministic and leads to inconsistencies in the resulting URLs.

That’s why we explicitly select a `language` for sitemap URL generation:
- Prefer the **default store language** (if allowed)
- Fallback to the first allowed language
- Fallback to the current one (last resort)

And finally, we use this condition:

```csharp
if (language.Id != workingLanguage.Id)
    url = GetLocalizedUrl(url, language);
```

to ensure the generated URL reflects the explicitly chosen language, overriding whatever may have been implicitly applied via `IOutboundParameterTransformer`.

This guarantees that all sitemap entries are generated consistently and intentionally, regardless of the current user context or route state.

bambuca:
> I think [this logic](https://github.com/nopSolutions/nopCommerce/pull/7778/commits/052034e9a37d2d68f363546712f3b9df06cce10c#diff-cfe252349708b9b74f3a8e423d6e1bb524f9d32e15f0c0df24d2afa086a25333R884-R900) is unnecessary in this context, it will be enough to simply filter languages using the DisallowLanguages setting

Thanks! Just to clarify — this PR actually addresses two related aspects:

1. It ensures that **alternate URLs are generated only for allowed languages** (those not listed in `DisallowLanguages`), which is especially important when `SeoFriendlyUrlsForLanguagesEnabled` is `true`.

2. It also makes sure that the **main language of the sitemap** is consistent by using `store.DefaultLanguageId`, instead of relying on the current session/context (`IWorkContext`), which could vary per request.  
   This avoids situations where, for example, the admin user has a language selected that is **not allowed** in the sitemap (e.g. listed in `DisallowLanguages`) — and yet the whole sitemap gets generated in that disallowed language.

This way, the sitemap is generated in a **deterministic** and **controlled** manner, aligned with store and SEO settings.

(Reference: [earlier comment](https://github.com/nopSolutions/nopCommerce/pull/7778#discussion_r2228604119))

src/Presentation/Nop.Web/Factories/SitemapModelFactory.cs
skoshelev: Okay, I understand your idea and the problem you are solving, but it is not related to the ticket itself. Let's create a separate ticket and a separate pull request for this task. I will review everything again. As for the task #7777 itself, we decided not to implement it in the core for now

Issue #7789, opened 2025-07-24
Use ISO 8601 with timezone for DateTime properties in the JSON-LD models


skoshelev:
All DateTime fields in JSON-LD models must be serialized in ISO 8601 format with timezone

skoshelev:
Closed #7789

PR #7791, opened 2025-07-24
Use ISO 8601 with timezone for DateTime properties in the JSON-LD models

wakku9780:
🔧 Fix: Serialize priceValidUntil using ISO 8601 with timezone
Changed DateTime? to DateTimeOffset? in JsonLdOfferModel to ensure proper timezone-aware JSON serialization for structured data as per issue #7789.

This helps Google and other consumers correctly interpret the expiry date of an offer with its associated timezone.


skoshelev:
Hi @wakku9780.

Thank you very much for your suggestion, but we solved this issue a little easier, and only for JSON-Ld models. You can see the details in [this commit](https://github.com/nopSolutions/nopCommerce/commit/68faf37204011f9e1abae1bb2e2f145d5521af46)

Issue #7787, opened 2025-07-22
Property MessageTokensAddedEvent.Tokens does not require a typed parameter U


skoshelev:
The Tokens property from the MessageTokensAddedEvent class contains a redundant typed parameter U, this property always receives IList<Token> as a value.

skoshelev:
Closed #7787

PR #7792, opened 2025-07-24
Removed unnecessary generic parameter from MessageTokensAddedEvent.Tokens property

wakku9780:
This PR fixes the issue where the Tokens property in the MessageTokensAddedEvent class was using a redundant generic parameter U. Since the property always accepts an IList<Token>, the type parameter was unnecessary and could lead to confusion or misuse.

Removed type parameter U from Tokens property

Updated all references accordingly

Fixes: #7787

🔗 Related Issue:
[Property MessageTokensAddedEvent.Tokens does not require a typed parameter U #7787](https://github.com/nopSolutions/nopCommerce/issues/7787)


skoshelev:
Hi @wakku9780.

Thank you very much for your desire to help, but unfortunately, your PR contains many changes that are not related to the main task, which makes it difficult to check the correctness of the changes.

Issue #7673, opened 2025-04-09
AutoMapper is going commercial


uttamughareja:
Automapper is going commercial 

Pricing model is not yet finalized, More details can be found here.
https://github.com/AutoMapper/AutoMapper/discussions/4536

Possible High Performance Alternatives
Mapster:
https://github.com/MapsterMapper/Mapster

Mapperly:
https://github.com/riok/mapperly

Or simple extension methods with handwritten code.

KaranChadha10:
@AndreiMaz @skoshelev I've created a Draft PR (https://github.com/nopSolutions/nopCommerce/pull/7829) to address this issue. Please review when possible. Thanks!

skoshelev:
Closed #7673

PR #7829, opened 2025-08-18
refactor [issue-7673] replaced automapper with mapster

KaranChadha10:
**Description

This PR contains the following points:**

- Introduced MapsterConfiguration (Nop.Core.Infrastructure.Mapper) to hold the global TypeAdapterConfig and IMapper.
- Removed AutoMapper package and usages:
- Deleted <PackageReference Include="AutoMapper" Version="14.0.0" /> from Nop.Core.csproj.
- Removed AutoMapperConfiguration class and its test.
- Cleared AutoMapper using directives where applicable.

**Why Mapster**

- No commercial licensing concerns.
- Faster mapping with fewer allocations; supports compile-time generation.
- Simple, explicit configuration via TypeAdapterConfig.

Issue Link: https://github.com/nopSolutions/nopCommerce/issues/7673


skoshelev:
Hi @KaranChadha10.

Hi, I looked at the changes, superficially for now, since this task is scheduled for a future release.

Overall, everything is fine, I just don't agree with the need to create the GetMapper method. I would suggest using the static MapsterConfiguration class.

But we need to take into account the main point, we have not yet finally decided which mapper to switch to and whether to switch at all. Mapster looks good, but has not been updated for a long time. We would not like to face a situation where we switch to a deliberately unsupported library.

PR #7831, opened 2025-08-19
refactor: cache parent→child lookup with full Category objects

KaranChadha10:
Description:

- This PR updates CategoryService to cache a parent→children lookup of full Category objects instead of just IDs.
- Added GetCachedCategoryLookupAsync() method that creates a cached dictionary lookup storing full Category objects instead of just IDs
- Optimized GetChildCategoryIdsAsync and GetAllCategoriesByParentCategoryIdAsync methods

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/CategoryService.cs
+++ b/src/Libraries/Nop.Services/Catalog/CategoryService.cs
@@ -155,6 +155,43 @@ public partial class CategoryService : ICategoryService
             yield return orphan;
     }
 
+    /// <summary>
+    /// Gets cached category lookup by parent category ID containing full category objects
+    /// </summary>
+    /// <param name="storeId">Store identifier</param>
+    /// <param name="showHidden">A value indicating whether to show hidden records</param>
+    /// <param name="customerRoleIds">Customer role identifiers</param>
+    /// <returns>
+    /// A task that represents the asynchronous operation
+    /// The task result contains the dictionary lookup of categories by parent ID
+    /// </returns>
+    protected virtual async Task<IReadOnlyDictionary<int, IReadOnlyList<Category>>> GetCachedCategoryLookupAsync(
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi @KaranChadha10.

I looked at the pull request. I want to point out a few things. The task is quite complex with a lot of pitfalls. You can look at the entire correspondence in the previous [pull request](https://github.com/nopSolutions/nopCommerce/pull/6931) to see some of them. Also note that this task is not marked for 4.90, since, due to the criticality of this functionality, we will not be able to include it in this release, as we simply do not have enough time for full testing.

I hope you understand correctly. I will have to close your pull request, since we will spend too much effort and time on detailing it.

However, as I said, I looked at your code, and here are a couple of points for the future:

- We do not use IReadOnly collections in our code, since they do not make sense and have no advantages, and because of them, your GetCachedCategoryLookupAsync method looks very cumbersome and can be significantly simplified by calling the ToGroupedDictionary method
- You added a new caching key, but you do not clear it anywhere. That is, with any changes in categories or mappings, the data in the cache will become irrelevant, and this will not be tracked in any way, given the repeated caching in the GetChildCategoryIdsAsync method
- The recursive CollectChildIds method looks strange, especially since it changes data that does not belong to its scope, which is not obvious. It would be better to design it something like this https://github.com/nopSolutions/nopCommerce/pull/6931/commits/c1957f5a77251c5c5210371664b61dc97367144f#diff-b11976958a0c67419eb7a5dd835294e9b84dceb2ead72bf96cf5a175e725f8a7R152, but evaluate the feasibility of using IEnumerable
- You removed the only use of the CategoriesByParentCategoryCacheKey key, but did not remove it from the NopCatalogDefaults class, nor did you remove the CategoriesByParentCategoryPrefix prefix

Thank you very much for your willingness to help us with development. I will be glad to see your future PRs.

Issue #7756, opened 2025-06-23
ModelPreparedEvent should contain a model class instance (not a general BaseNopModel instance)


skoshelev:
Currently, when publishing the ModelPreparedAsync event, an object of the BaseNopModel base class is passed to it instead of an instance of the final class. This leads to the fact that all events in a row are intercepted in the handler, and it is necessary to check the required class through code similar to next one: 
```cs
eventMessage.Model is ShoppingCartModel model
```

KaranChadha10:
@AndreiMaz @skoshelev I've created a PR (https://github.com/nopSolutions/nopCommerce/pull/7833) to address this issue. Please review when possible. Thanks!

PR #7833, opened 2025-08-20
refactor: Replace generic BaseNopModel event consumers with type-specific handlers

KaranChadha10:
**Description**

- Remove manual type checking and casting (is/as operators) from event handlers
- Separate mixed model logic into focused, single-responsibility methods
- Improve type safety and eliminate runtime type checking overhead

Issue link: https://github.com/nopSolutions/nopCommerce/issues/7756


skoshelev:
Thank you very much for your help. However, this task is not as simple as it seems at first glance, and has pulled along a large layer of changes. You can see all the changes in [this branch ](https://github.com/nopSolutions/nopCommerce/compare/develop...issue-7756-ModelPreparedEvent-should-contains-exectly-model-classs-instance)

PR #7846, opened 2025-09-15
ignore its mistake (i was trying to update automapper to riok.mapperly fully by ai and create webapi plugin by ai too - There are probably a lot of bugs in it.)

amrmostafa800:


amrmostafa800:
wait what happen here
i think i open pull request in my respository not here
all edits here is by ai and maybe all worng

Issue #7867, opened 2025-09-19
Discount optimizations


danFbach:
Another frequently complained about bottleneck.
This can be optimized with parallelization. 

I have dozens of complicated discount requirements - i have >8k products. Some pages load > 100 products at a time.

Some basic benchmarks I've run, before and after parallelizing tasks in the PriceCalculationService and DiscountService.

```txt
| PROC          | Load w/o Prl | Load w/ Prl |
| 3.5Ghz 8-Core |       1300ms |       700ms |
| 2.4Ghz 8-Core |       3300ms |      1650ms |
```

Will submit a pull request with proposed changes.

skoshelev:
We have decided to leave this for customization.

PR #7868, opened 2025-09-19
#7867 discounts optimizations

danFbach:
optimizes discount retrieval by parallelizing requests, validations calculations.
Uses ConcurrentDictionary with Semaphores to prevent race condition issues.

Try this out in your own test stores. I am running it in my stores and have seen a significant reduction in load and response time. See issue for benchmark table.

It is possible that there is overhead introduced in smaller/less complex stores, so a setting to turn it on and off should maybe be considered.

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/PriceCalculationService.cs
+++ b/src/Libraries/Nop.Services/Catalog/PriceCalculationService.cs
@@ -75,11 +75,23 @@ public partial class PriceCalculationService : IPriceCalculationService
             return allowedDiscounts;
 
         var couponCodesToValidate = await _customerService.ParseAppliedDiscountCouponCodesAsync(customer);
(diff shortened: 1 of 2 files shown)


skoshelev:
Hi, @danFbach!

I’ve analyzed your code and want to thank you very much. It’s a good solution to the performance problem in the face of numerous discounts. However, we cannot accept your code into the core, as such optimizations are not always necessary, and the code becomes more difficult to read and maintain.

In our project, we strive to strike a balance between performance and a low barrier to entry for new developers. It is also important to keep in mind that parallelization works well on multi-core or multi-processor systems, but, unfortunately, there are servers where websites run on a single core. Parallelization on such a system will only lead to a performance penalty.

I am very sorry we cannot accept your code. I hope you understand me correctly. We have decided to leave this for customization.

Issue #7847, opened 2025-09-15
DeleteGuests task optimization (batch)


danFbach:
I know that the DeleteGuests task failing has been brought up before - i have myself brought it up before here and on the forums.
This is still frequently posted about in the forums and people have devised all kinds of crazy work-arounds simply to cleanup their guest accounts. Personally my "DeleteGuests" task hasn't run successfully in many months.
I had chatgpt refactor it to a batched query instead of one mega query which is known to time out, lock up the database, leaves behind artifact tmp tables, etc...
The batch size could probably be tweaked, but the batch size of 500 runs pretty quickly and without error.
I strongly suggest you implement this, or something similar and put this issue to rest.

```c#
public override async Task<int> DeleteGuestCustomersAsync(DateTime? createdFromUtc, DateTime? createdToUtc, bool onlyWithoutShoppingCart)
{
    var guestRole = await GetCustomerRoleBySystemNameAsync(NopCustomerDefaults.GuestsRoleName);
    var batchSize = 500;
    var totalDeleted = 0;

    while (true)
    {
        // Step 1: Find a batch of guest customer IDs to delete
        var guestIdsQuery =
            from guest in _customerRepository.Table
            join ccm in _customerCustomerRoleMappingRepository.Table on guest.Id equals ccm.CustomerId
            where ccm.CustomerRoleId == guestRole.Id
                  && !guest.IsSystemAccount
                  && (createdFromUtc == null || guest.CreatedOnUtc > createdFromUtc)
                  && (createdToUtc == null || guest.CreatedOnUtc < createdToUtc)
            select guest.Id;

        // Exclude guests with related data if needed
        if (onlyWithoutShoppingCart)
        {
            guestIdsQuery = from id in guestIdsQuery
                            where !_shoppingCartRepository.Table.Any(sci => sci.CustomerId == id)
                            select id;
        }

        guestIdsQuery = from id in guestIdsQuery
                        where !_orderRepository.Table.Any(o => o.CustomerId == id)
                              && !_blogCommentRepository.Table.Any(bc => bc.CustomerId == id)
                              && !_newsCommentRepository.Table.Any(nc => nc.CustomerId == id)
                              && !_productReviewRepository.Table.Any(pr => pr.CustomerId == id)
                              && !_productReviewHelpfulnessRepository.Table.Any(prh => prh.CustomerId == id)
                              && !_pollVotingRecordRepository.Table.Any(pvr => pvr.CustomerId == id)
[...]

PR #7869, opened 2025-09-19
#7847 optimize delete guests async

danFbach:
optimizes delete guests async by breaking queries into batches.
significantly reduces chances of long running task that locks the database and blocks store main thread waiting on db.
reduces chances of failed task which results in remnant tmp_ tables.
runs faster overall since queries are entirely smaller.


Parthiban6316:
Hii, I am Pathiban Sivaraj I would like to work on this issue. I am  good in EF core and MSSQL

Issue #7878, opened 2025-09-29
Delete plugins runtimes folders after build


skoshelev:
Currently, when building plugins or the entire project, the plugin output includes a runtimes directory containing specific library versions. However, since we're building the project for a specific operating system, we have no need for these builds. 
We've long ignored them when publishing the project. Let's delete this directory after each build to save disk space.

KaranChadha10:
@skoshelev I have raised the PR https://github.com/nopSolutions/nopCommerce/pull/7880 to address this issue, kindly take time to review the PR, thanks.

PR #7880, opened 2025-10-01
refactor: [issue-7878] delete plugins runtime folders after build

KaranChadha10:
**Description**

- This PR implements automatic deletion of runtimes folders from plugin output directories after each build, saving disk space and keeping the build output clean.

Issue link: https://github.com/nopSolutions/nopCommerce/issues/7878


skoshelev:
Hi @KaranChadha10
Many thanks for your code

Issue #7967, opened 2025-11-20
Refactoring of migration for Alter, Delete, and Schema operations using NameCompatibilityManager


atiq-bs23:
We currently have `TableFor<TEntity>` extension support only for `ICreateExpressionRoot`, where it correctly uses `NameCompatibilityManager` for resolving table names.

But we also need similar support for `IDeleteExpressionRoot`, `IAlterExpressionRoot` and `ISchemaExpressionRoot`

And additionally, we need `ColumnFor` as well to support for schema checks, so `NameCompatibilityManager` is also used for column name mapping.

Otherwise migrations can break when database table/column names differ from entity class/property names.

Current implementation 
```
var productTableName = nameof(Product);
if (!Schema.Table(productTableName).Column(ageVerificationColumnName).Exists())
{
    Alter.Table(productTableName)....
}
```

Expected code like below

```
if (!Schema.TableFor<Topic>().ColumnFor<Topic>(t => t.AvailableEndDateTimeUtc).Exists())
 {
     Alter.TableFor<Topic>()
         .AddColumnFor<Topic>(t => t.AvailableEndDateTimeUtc)
         .AsDateTime()
         .Nullable();
 }
 
var footerColumn1ColumnName = "IncludeInFooterColumn1";
if (Schema.TableFor<Topic>().Column(footerColumn1ColumnName).Exists())
    Delete.Column(footerColumn1ColumnName).FromTable<Topic>();
```

atiq-bs23:
#7952 related

skoshelev:
Closed #7967

skoshelev:
Closed #7967

Issue #7952, opened 2025-11-11
Add TableFor extension method to IDeleteExpressionRoot interface


skoshelev:
We have TableFor extension method for the ICreateExpressionRoot interface, which responds to uses NameCompatibilityManager for creating a new table. But we need the same method for IDeleteExpressionRoot  to use NameCompatibilityManager  during deleting the table

skoshelev:
Absorbed by #7967

PR #7968, opened 2025-11-20
Add TableFor and ColumnFor extensions with NameCompatibilityManager support

atiq-bs23:
This PR adds `TableFor`, `ColumnFor` and `FromTable` extension methods for `IDeleteExpressionRoot`, `IAlterExpressionRoot`, and `ISchemaExpressionRoot `to ensure table and column names are resolved via `NameCompatibilityManager`.

Supports schema checks, altering, adding, and deleting columns in a type-safe manner.

Includes strongly-typed column selector overloads for safer refactoring.

Updated XML summaries for all new methods for consistency with existing extension methods.

**Note:** Changes in `SchemaMigration` are only for reference purposes.
These changes will not affect any existing migrations since the `NopSchemaMigration` attribute date has not been modified, and the migration will not execute.

Now we can write like below
```
 if (!Schema.TableFor<Topic>().ColumnFor<Topic>(t => t.AvailableEndDateTimeUtc).Exists())
 {
     Alter.TableFor<Topic>()
         .AddColumnFor<Topic>(t => t.AvailableEndDateTimeUtc)
         .AsDateTime()
         .Nullable();
 }
 
var footerColumn1ColumnName = "IncludeInFooterColumn1";
if (Schema.TableFor<Topic>().Column(footerColumn1ColumnName).Exists())
    Delete.Column(footerColumn1ColumnName).FromTable<Topic>();
```

### DIFF
--- a/src/Libraries/Nop.Data/Extensions/FluentMigratorExtensions.cs
+++ b/src/Libraries/Nop.Data/Extensions/FluentMigratorExtensions.cs
@@ -8,6 +8,7 @@ using FluentMigrator.Builders.Create;
 using FluentMigrator.Builders.Create.Table;
 using FluentMigrator.Builders.Delete;
 using FluentMigrator.Builders.Schema;
+using FluentMigrator.Builders.Schema.Table;
 using FluentMigrator.Infrastructure.Extensions;
 using FluentMigrator.Model;
 using FluentMigrator.Runner;
@@ -228,6 +229,21 @@ public static class FluentMigratorExtensions
     {
         var tableName = NameCompatibilityManager.GetTableName(typeof(TEntity));
         return expressionRoot.Table(tableName).Column(columnName).Exists();
+    }
+
+    /// <summary>
+    /// Targets the entity's mapped table for schema-related operations.
+    /// </summary>
+    /// <param name="expressionRoot">The root expression for schema inspection</param>
+    /// <typeparam name="TEntity">The entity type mapped to the database table</typeparam>
+    /// <returns>
+    /// A fluent syntax interface for performing schema operations 
+    /// such as checking table or column existence.
+    /// </returns>
+    public static ISchemaTableSyntax TableFor<TEntity>(this ISchemaExpressionRoot expressionRoot) where TEntity : BaseEntity
+    {
+        var tableName = NameCompatibilityManager.GetTableName(typeof(TEntity));
+        return expressionRoot.Table(tableName);
     }
 
     /// <summary>


atiq-bs23:
#7967 #7952 related

atiq-bs23:
@skoshelev Thanks for your feedback, I'll mention you once all comments resolved.

src/Libraries/Nop.Data/Extensions/FluentMigratorExtensions.cs
```diff
@@ -145,6 +153,101 @@ public static void TableFor<TEntity>(this ICreateExpressionRoot expressionRoot)
...
+    /// <param name="tableSchema">The schema table expression</param>
+    /// <param name="selector">An expression selecting the entity property</param>
+    /// <returns>
+    /// A fluent syntax interface for performing schema operations 
+    /// such as checking column existence.
+    /// </returns>
+    public static ISchemaColumnSyntax ColumnFor<TEntity>(
+    this ISchemaTableSyntax tableSchema, Expression<Func<TEntity, object>> selector) where TEntity : BaseEntity
+    {
+        var property = ((MemberExpression)selector.Body)?.Member?.Name;
```
skoshelev: The static analyzer says that the null checks in this code are redundant

skoshelev:
Hi @atiq-bs23!

Thank you very much for your help. 

I looked at the code and left a couple of suggestions. Please make the necessary changes. Also, modify older migrations so that the code seems consistent across the entire codebase.

atiq-bs23:
@skoshelev I resolved all comments, please merge this one so that I can create a new PR for updating all existing migrations

atiq-bs23:
There are a lot of changes in existing migrations and also I found some code where NameCompatibility is not considered. So I think it would be good to have it in a separate PR.

@skoshelev

skoshelev:
Hi @atiq-bs23.

I've looked at the code again. I suggest closing this PR and moving the changes to a new one, as this one has too many commits. But ultimately, there aren't that many changes. 

Technically, this shouldn't be a problem for you; copy the code from two files separately from the current code and replace them in the new branch. You'll get all the changes in a single commit.

But there's one more thing I'd like to add before you start updating older migrations.

In another task, we're working on simplifying migrations that insert locales and settings. There, we came to the decision to combine the Exist methods with action methods. I suggest adding such methods for tables with columns as well. This way, you can call methods like this: `Create.TableForIfNotExist<Menu>(); ` instead of
```
if (!Schema.TableExist<Menu>())
Create.TableFor<Menu>();
```

Thanks again for making our code more compact and readable.

atiq-bs23:
@skoshelev Thanks for taking the time to review the PR and for your suggestions. I’ll create a new PR incorporating your recommended changes. I’m currently traveling, so I will open the PR the day after tomorrow.

atiq-bs23:
@skoshelev Created a new PR so closing this one

atiq-bs23:
@skoshelev Maybe Create.TableForIfNotExist<Menu>(); will not support because the method is extension with ICreateExpressionRoot, but to check we need ISchemaExpressionRoot

skoshelev:
Hi @atiq-bs23.

I didn't go into the implementation details, but perhaps it's worth making the extension method not on a specific action interface, but using IMigration as a base, and then calling the methods CreateTableIfNotExist

atiq-bs23:
@skoshelev In that we case we need to use `this` keyword when calling inside SchemaMigration
Like below
`this.CreateTableIfNotExist<Menu>();`

skoshelev:
@atiq-bs23 Yes, we know about this feature, but considering the overall simplification of the code, this is not a high price to pay.

atiq-bs23:
@skoshelev I've updated the code, you can review

PR #7974, opened 2025-11-26
Check Deleted flag when retrieving ProductManufacturers by product

atiq-bs23:
This fixes an issue where deleted manufacturers could still appear in the product manufacturers list. Added a check to ensure only non-deleted manufacturers are matched when showHidden flag is false.

### DIFF
--- a/src/Libraries/Nop.Services/Catalog/ManufacturerService.cs
+++ b/src/Libraries/Nop.Services/Catalog/ManufacturerService.cs
@@ -213,9 +213,9 @@ public partial class ManufacturerService : IManufacturerService
 
         if (discountId.HasValue)
             manufacturers = from manufacturer in manufacturers
-                join dmm in _discountManufacturerMappingRepository.Table on manufacturer.Id equals dmm.EntityId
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi @atiq-bs23.
Thank you very much for your help.

I left a comment on the code. This check is already in place, just above the corrected code, so the situation you described can't happen.

For future reference, it's useful to know that the showHidden parameter is never used in our code to display deleted posts.

And one more thing, please keep an eye on the automatic edits in the code, as you've essentially changed one line, but we've received a large number of changes with changed indentations. This makes it very difficult to understand the edits.

atiq-bs23:
@skoshelev then should we not fix for the showHidden?

skoshelev:
@atiq-bs23 I don't quite understand what exactly you're suggesting to fix.

atiq-bs23:
@skoshelev  if the showHidden is true, it still filtering Deleted

skoshelev:
@atiq-bs23 This is how it should be, records marked as deleted for nopCommerce are considered permanently deleted and are not loaded in most cases, regardless of the showHidden flag.

PR #8030, opened 2025-12-11
Refactor: simplify null and published check for language

RyanDDDDDD:
### Summary
Simplified the conditional check for language retrieval to improve readability.

### Changes
- Replaced the expression `if (!language?.Published ?? false)` with a clearer equivalent:
  ```csharp
  if (language == null || !language.Published)
      language = await _workContext.GetWorkingLanguageAsync();


RomanovM:
@RyanDDDDDD Why? There are tons of similar checks within the solution

RyanDDDDDD:
The reason I changed this specific check is readability.  

`if (language == null || !language.Published)` is easier to parse at a glance than  
`if (!language?.Published ?? false)`

RomanovM:
Thanks for the suggestion, but we leave it as-is.

PR #8054, opened 2025-12-24
Optimize some methods by batching SQL and reducing the number of DB calls

atiq-bs23:
This PR optimizes several methods by batching SQL operations and reducing the number of database calls.
The goal is to improve performance and efficiency while keeping the existing behavior unchanged


atiq-bs23:
@skoshelev

skoshelev:
Hi @atiq-bs23.

Thank you for your work, but I can't accept this PR. Numerous markup changes unrelated to the main goal greatly complicate the analysis. It would be better to create separate PRs for each method, or at least for each service.

atiq-bs23:
@skoshelev thank you, I'll create PRs separately

Issue #8076, opened 2026-01-15
Async blocking in FluentValidation extension methods used across validators


Navidliverpool:
Async blocking in FluentValidation extension methods used across validators

Description

While reviewing the validation infrastructure, I found a widespread async issue caused by custom FluentValidation extension methods that synchronously block on async code. The issue exists across the solution and affects a large number of validators.

Root Cause

`RuleBuilderOptionsExtension.cs`

This file defines async extension methods that block on tasks using `.Result`:

* `WithMessageAwait()` (3 overloads)
* `MustAwait()`
* `WhenAwait()`

These methods are reused extensively by validators throughout the solution.

Scope

Verified usages affect 100+ validator files, including:

* Admin validators (catalog, customer, message/email, settings, shipping, tax, store)
* Public validators (`RegisterValidator`, `LoginValidator`, `CustomerInfoValidator`, `ContactUsValidator`, `AddressValidator`, etc.)
* Plugin validators (`Tax.Avalara`, `Payments.PayPalCommerce`, `Widgets.FacebookPixel`, and others)

Impact

* Blocking async code with `.Result` introduces a potential deadlock risk in ASP.NET Core
* Causes thread blocking, reducing scalability under load
* Affects the core validation layer, not isolated functionality

Suggested Direction

FluentValidation v10+ provides native async validation support. A possible approach would be to remove these custom extension methods and migrate validators to the built-in async APIs:

* `MustAwait()` → `MustAsync()`
* `WhenAwait()` → `WhenAsync()`
* Refactor `WithMessageAwait()` usages to proper async validation patterns

This would be a larger refactor, but it would align the codebase with FluentValidation’s supported async model and avoid synchronous blocking in validation logic.

Should this approach be considered acceptable, I would be glad to contribute an implementation.

skoshelev:
Closed #8076

PR #8079, opened 2026-01-17
Refactor FluentValidation validators to remove synchronous blocking #8076

Navidliverpool:
## Description
This PR refactors FluentValidation validators throughout the nopCommerce solution to remove synchronous blocking on async operations.

Problem
The custom `RuleBuilderOptionsExtension.cs` methods used `.Result` on `Task<string>` objects, which:
- Blocks threads unnecessarily
- Impacts performance under load
- Affects all user registrations, logins, and form validations

Solution
- **Deleted** `RuleBuilderOptionsExtension.cs` (source of the bug)
- **Updated** all validators to use FluentValidation's native async methods:
  - `MustAwait` → `MustAsync` (with CancellationToken)
  - `WhenAwait` → `WhenAsync` (with CancellationToken)
  - `WithMessageAwait` → `WithMessage` (with resource keys)
- **Updated** affected tests to use `TestValidateAsync`

Benefits
✅ Removes synchronous blocking in validators  
✅ Follows FluentValidation best practices  
✅ Improves performance and scalability  
✅ No behavioral changes to validation logic  

Testing
- ✅ All existing tests pass  
- ✅ Build successful  
- ✅ No breaking changes  

Files Changed
- **Deleted:** 1 file (`RuleBuilderOptionsExtension.cs`)  
- **Modified:** 102 files (101 validators + 1 test)


skoshelev:
Hi @Navidliverpool!

Hi. Thank you very much for your help. However, this isn't a solution, as instead of an error message, the user will see the locale name.

image

It's also important to note that this code works without any issues and doesn't cause any deadlocks or thread exhaustion, so we need to find a good alternative to calling .Result or .GetAwaiter().GetResult(). But there is no rush in this.

Navidliverpool:
@skoshelev 
Hi, thank you for the clarification. That makes sense — I missed the localization impact. I apologise for the mistake. I’ll keep this in mind and follow up if I find a suitable approach. Thanks again for the feedback.

Navidliverpool:
Closing this PR for now. I'll submit a new one once I have a proper solution. Thank you @skoshelev again.

Issue #8000, opened 2025-12-02
More user-friendly display of quantity validation messagen when submitting return request


anikdasbs23:
nopCommerce 4.90.2

When the submit button is clicked without selecting a quantity, the return request form should remain visible. Instead of disappearing, it should show a validation message at the top of the page

**Steps to Reproduce:**

1. Log in with valid customers credentials.
2. Complete checkout for a product 
3. Log in with valid admin credentials and complete shipment
4. Log in with valid customer credentials again and navigate to orders
5. Click on the return request
6. Submit the forms without selecting qty

**Solution:**
Let's refactor the "ReturnRequestSubmit" method logic (and its view) so when we have the "ReturnRequests.NoItemsSubmitted" error after "_returnRequestModelFactory.PrepareSubmitReturnRequestModelAsync" execution, then this error is displayed as a validation message (not like a result)

Video:
https://jam.dev/c/9fac833a-37f8-4d61-a5aa-63269b09037d

AndreiMaz:
Thanks a lot! Although quantity could be 0 in some cases (when you have multiple products), I'll update the issue description accordingly

jat-zare:
I have read the relevant code and I would like to create a PR.
We can show an error barNotification to the user if they didn't select any item at all.
Are you open to PR? @AndreiMaz , @DmitriyKulagin

DmitriyKulagin:
@jat-zare Yes you can do PR

DmitriyKulagin:
Closed #8000

PR #8082, opened 2026-01-21
Add client-side validation to require at least one item in return request (#8000)

jat-zare:
Hi @DmitriyKulagin 

This PR addresses issue #8000
**Changes:**
- Added client-side validation on return request form submission
- Prevent submission if no product has quantity > 0
   - Show localized error notification: `displayBarNotification('@T("ReturnRequests.Submit.Errors.NothingSelected")', 'error', 4000);`


DmitriyKulagin:
Hi @jat-zare,

Thank you for your contribution, but we decided not to implement this check on the client side to avoid duplicating logic. You can see our implementation in this commit - https://github.com/nopSolutions/nopCommerce/commit/ecccfd88db433af80a146ade4ac47b5a4f4c4c97

Issue #8083, opened 2026-01-21
"Continue Shopping" button redirects to incorrect page after adding product from Home page


Kibria-BS1954:
**Title**

Continue Shopping button redirects to incorrect page after adding product from Home page

**Environment**

Application: nopCommerce

Browser: Edge (Windows 11)

Build/Version: 4.9

**Preconditions**

User is on the Home page

Product is available to add to cart

**Steps to Reproduce**

Open the Home page

Add any product to the cart directly from the Home page

After the success popup appears, click on Continue Shopping

**Expected Result**

User should be redirected back to the Home page

**Actual Result**

User is redirected to a different/incorrect page instead of the Home page

**Impact**

Causes confusion for users

Breaks expected shopping flow

Affects user experience during browsing

**Severity**

Medium

**Priority**

High

Notes / Attachments:
[https://jam.dev/c/46741f44-aeb9-44d2-9272-4d0c317ec7a6](url)

RomanovM:
@Kibria-BS1954 Can you reproduce this on our [demo site](https://demo.nopcommerce.com/)?

Kibria-BS1954:
@RomanovM Hello, here is the link you can follow these step: [https://jam.dev/c/9a78d5d0-911a-4d5c-a745-f80242beafe4](url)

atiq-bs23:
Just an addition, if we add to cart from home page it redirects homepage which if fine.
I checked the code, and found the problem is when add to cart from a category, manufacturer, vendor or search page it stored the generic attribute but later when add to cart from home page it is not storing the last visited page

@RomanovM @skoshelev

Kibria-BS1954:
@RomanovM When a product is added to the cart from any page, clicking Continue Shopping on the cart page redirects the user back to the same page from where the product was added.

However, when a product is added from the Home page, clicking Continue Shopping does not redirect back to the Home page.

Demo site Steps video: [https://jam.dev/c/9a78d5d0-911a-4d5c-a745-f80242beafe4](url)

skoshelev:
Closed #8083

skoshelev:
Closed #8083

PR #8084, opened 2026-01-24
Bug Fix: "Continue Shopping" button redirects to incorrect page after adding product from Home page

atiq-bs23:
Previously, the last visited page URL was stored only on specific pages (`Category`, `Manufacturer`, `Vendor`, `Search`, and `SearchByFilterLevelValues`).
However, `ShoppingCartController.AddToCartCatalog` can be called from any page (`Category`, `Manufacturer`, `Vendor`, `Search`, `SearchByFilterLevelValues `or others).

This PR updates `AddToCartCatalog `to store the referrer page URL (using `_webHelper.GetUrlReferrer()`), ensuring the Continue Shopping button always redirects users back to the exact page where the item was added to the cart.

`AddToCartDetails `keeps resetting the value to an empty string to preserve existing redirect behavior.

### DIFF
--- a/src/Presentation/Nop.Web/Controllers/ShoppingCartController.cs
+++ b/src/Presentation/Nop.Web/Controllers/ShoppingCartController.cs
@@ -403,11 +403,6 @@ public partial class ShoppingCartController : BasePublicController
             case ShoppingCartType.ShoppingCart:
             default:
             {
-                //'Continue shopping' URL
-                await _genericAttributeService.SaveAttributeAsync(await _workContext.GetCurrentCustomerAsync(),
-                    NopCustomerDefaults.LastContinueShoppingPageAttribute,
-                    string.Empty,
-                    store.Id);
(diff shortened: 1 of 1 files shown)


atiq-bs23:
@skoshelev fix for #8083,
Please let me know if you have any suggestions or would prefer a different approach. Thanks

skoshelev:
Hi @atiq-bs23.

This is a good idea. Unfortunately, there are several inaccuracies in the implementation. The code that deletes the saved page is unnecessary, as it leads to problems when a product can't be added from the catalog page, for example, when attributes need to be selected. However, the code that saves the desired page is located too low in the code, and in some situations, the correct page may not be saved. The entire new block needs to be moved higher and re-placed immediately after checking that the product exists.

atiq-bs23:
@skoshelev  thanks a lot for you suggestion, I have updated the PR. Please review again

Issue #8090, opened 2026-01-29
Show default value for product review rating


jat-zare:
When a user submits a product review, the rating field is currently not required. This allows the review to be submitted without explicitly selecting a rating.
However, if no rating is selected, the system implicitly treats the review as Excellent (5 stars) after submission. This behavior can be misleading, as it may not reflect the user’s actual intention and can negatively affect review accuracy and overall UX.
Steps to reproduce:
1.	Log in.
2.	Open a product page.
3.	Submit the review without selecting a rating.

If the maintainers agree with this improvement, I would be happy to prepare and submit a pull request. I can implement the following solutions or any other you prefer:
1.	Make the rating field required before allowing submission.
2.	Set a clear default value for the rating radio buttons so the user is always aware of the selected rating.

exileDev:
Closed #8090

PR #8091, opened 2026-01-30
#8090 Show default value for product review rating

jat-zare:
Hi @exileDev 
this references issue #8090 ,
The default rating value is now set based on `CatalogSetting.DefaultProductRatingValue`
let me know if there is any suggestions


jat-zare:
Got it, thanks.
I originally created the branch based on develop, which was not the correct target for merging.
I have now created a new PR with a branch based on 4.90-bug-fixes.
@exileDev

PR #8143, opened 2026-03-18
Add null check for current vendor in Create action

noorzaka:
Adds a null check for the current vendor in the Create action of ProductController to prevent errors if no vendor is found.


AndreiMaz:
Vendors are not reqired for products.

Issue #8136, opened 2026-03-13
Add new social media links - TikTok, Snapchat, Pinterest, and Tumblr


DmitriyKulagin:
Currently, the social model only includes four social links (Facebook, X, YouTube, Instagram). 
Add a few more: TikTok, Snapchat, Pinterest, and Tumblr.

i-dentify:
Discord as well, please!

compito:
Or just make it easy to add new ones by just adding a new row, put in name and url, and it's done forever? 
It's generally preferred method in many systems.

DmitriyKulagin:
Closed #8136

PR #8148, opened 2026-03-26
Feature/append new social links

0xfelaback:
This PR addresses Issue #8136


AndreiMaz:
Thanks a lot! But we've already implemented it

PR #8150, opened 2026-03-27
Upgrade FluentMigrator from MySQL 5 to MySQL 8

### LINKED ISSUE #8145: Configure FluentMigrator to use MySQL 8-specific services
We currently use FluentMigrator with a configuration for [MySQL version 5](https://github.com/nopSolutions/nopCommerce/blob/release-4.90.4/src/Libraries/Nop.Data/Extensions/FluentMigratorExtensions.cs#L74). However, official support for this version ended a long time ago, so we should switch to a configuration specific to version 8 and above.

MazenSghaier:
I've successfully upgraded your FluentMigrator configuration from MySQL 5 to MySQL 8+. Here's what was changed:
Changes Made:
1.	Created new MySQL 8 TypeMap (NopMySql8TypeMap.cs):
•	New class NopMySql8TypeMap that extends MySql8TypeMap instead of MySql5TypeMap
•	Maintains the same custom type mapping logic for character sets and collations
2.	Updated FluentMigrator Extensions (FluentMigratorExtensions.cs):
•	Changed AddMySql5() to AddMySql8() in the AddNopDbEngines() method
•	This applies to both the installed database scenario and the initial setup scenario
3.	Updated Database Startup Configuration (NopDbStartup.cs):
•	Changed the registered IMySqlTypeMap from NopMySql5TypeMap to NopMySql8TypeMap


skoshelev:
Hi, @MazenSghaier! 
Thank you very much for your time. I looked at the code, and there are a couple of overlooked points:

- The MySql5TypeMap class needs to be removed, as it's no longer needed.
- The target MySql version shuld by updated for Linq2db too
- The MySqlConnector assembly needs to be updated
- We need to ensure that the changes work correctly for both the existing database and the new installation

MazenSghaier:
Thank you for pointing the issues I should check and fix, the PR will be ready soon.

src/Libraries/Nop.Data/DataProviders/MySqlDataProvider.cs
```diff
@@ -353,7 +353,7 @@ public virtual Task<string> GetDataBaseCollationAsync()
     /// <summary>
     /// MySql data provider
     /// </summary>
-    protected override IDataProvider LinqToDbDataProvider => MySqlTools.GetDataProvider();
+    protected override IDataProvider LinqToDbDataProvider => MySqlTools.GetDataProvider("8.0");
```
skoshelev, line 356: It doesn't work

Image

skoshelev:
Hi, @MazenSghaier.

I'm very sorry, your changes don't work. It looks like you neglected to check if the database was being installed from scratch. Please pay attention to the NopProcessorAccessor and NopGeneratorAccessor classes, you haven't changed the version in them either, which is why errors occur when working.. Before a new commit, please make sure that the project starts and works both with an existing database and when installed from scratch.

MazenSghaier:
Hi @skoshelev, 
My apologies I'll make as much tests as possible to make sure that it is migrated to mySQL 8.

skoshelev:
Hi @MazenSghaier. 
Thanks for your help, I've modified your code myself and closed the ticket with this [commit](https://github.com/nopSolutions/nopCommerce/commit/582a33e15ac33794be1e722ab2668b016860b478).

PR #8152, opened 2026-04-01
Simplified DataSettingsManager class

### LINKED ISSUE #8141: Simplify the DataSettingsManager class
Currently, the DataSettingsManager class contains a number of methods that simply forward calls to the DataConfig object, such as the GetSqlCommandTimeout method. We can simplify this class and leave it with only three methods: LoadSettings, SaveSettings, and IsDatabaseInstalled.

phantom930:
Closes #8141 

## Summary

Simplifies `DataSettingsManager` so it only handles persistence and installation state, not per-call reads from `DataConfig`.

## What changed

- Removed forwarding helpers from `DataSettingsManager` (`GetSqlCommandTimeout`, `UseNoLock`, `GetBulkCopyOptions`, `GetCloseDataContextAfterUse`).
- Call sites use the existing `DataSettings` / `DataConfig` instance instead (`BaseDataProvider`, `MsSqlDataProvider`).

### DIFF
--- a/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
+++ b/src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
@@ -1,4 +1,4 @@
-﻿using System.Data.Common;
+using System.Data.Common;
 using System.Linq.Expressions;
 using System.Reflection;
 using FluentMigrator;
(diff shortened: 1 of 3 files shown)


phantom930:
@skoshelev Could you kindly review my PR?

src/Libraries/Nop.Data/DataProviders/BaseDataProvider.cs
```diff
@@ -18,6 +18,18 @@ public abstract partial class BaseDataProvider
 {
     #region Utilities
 
+    private static BulkCopyOptions CreateBulkCopyOptions(DataConfig settings)
```
skoshelev: This code highlights several points, mostly related to the stylistics of our code:

- We do not use private methods in classes implementing service interfaces.
- All protected methods must be virtual.
- All methods, except private ones, must be accompanied by documentation comments.
- This method must not be static.
- This method does not require a parameter; the class already has access to the DataConfig object.

phantom930: @skoshelev I addressed this comment

PR #8168, opened 2026-04-17
Deletes temporary tables before re-creating it

markusschnepf:
"tmp_guestsToDelete and"
and
"tmp_guestsAddressesToDelete"

threw an "table already exsists" error when the shop restarted before the "deleteGuests" task was finished and you ran the job again.


exileDev:
Thanks!

Issue #8159, opened 2026-04-13
Add exception handling when invoking EncryptionService methods


RomanovM:
`System.Security.Cryptography.CryptographicException` is possible when using EncryptText() or DecryptText() methods. So, it should be wrapped in a try-catch block.
Is seems the problem currently exists only in `OrderModelFactory.PrepareOrderModelPaymentInfoAsync()` method, the other places are already have exception handling. Let's check.

skoshelev:
Closed #8159

skoshelev:
Closed #8159

PR #8179, opened 2026-04-30
#8159 Add exception handling for EncryptionService methods

wakku9780:
- Added exception handling when invoking EncryptionService methods
- Prevents crashes due to encryption failures
- Improves system stability and error handling

### DIFF
--- a/src/Libraries/Nop.Core/BaseEntity.cs
+++ b/src/Libraries/Nop.Core/BaseEntity.cs
@@ -9,4 +9,5 @@ public abstract partial class BaseEntity
     /// Gets or sets the entity identifier
     /// </summary>
     public int Id { get; set; }
+
 }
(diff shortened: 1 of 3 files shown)


skoshelev:
Hi, @wakku9780!

Thank you very much for your help, we've closed the ticket with [this commit](https://github.com/nopSolutions/nopCommerce/commit/b3f7da72041ef7ec41c48c84f62fb57d30f8b59b).

But I've also reviewed your changes. You may have noticed that you haven't added exception handling in all places. Also, it's not very correct to set empty values ​​if an error occurs. It's better to do nothing and notify the user about the problem than to hide it.

I also noticed some fixes in the tests. I'm not sure if the logic is broken, but even if there really is a problem, it's better to file it in another PR, since this test doesn't address the original issue.

PR #8187, opened 2026-05-08
ci: add NuGet package caching to speed up builds

Omnivex3:
## Description
Adds NuGet package caching to the CI build workflow. The cache key is based on a hash of all `.csproj` files and `src/Directory.Build.props` — it automatically invalidates when dependencies change.

## Impact
With caching:
- **Cache hit:** Restore uses cached packages (saves 1-3 min per build)
- **Cache miss:** Falls back to downloading packages (no regression)
- `restore-keys` fallback ensures it never blocks the build

## Testing
No functional change — the `dotnet restore`, `dotnet build`, and `dotnet test` commands remain identical. Caching is transparent.

## Checklist
- [x] Cache key covers dependency files (all `.csproj` files + `Directory.Build.props`)
- [x] Restore key provides fallback for partial cache hits
- [x] Cache path is `~/.nuget/packages` (Windows-standard for `windows-latest` runner)

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -1,6 +1,3 @@
-# This workflow will build a .NET project
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net
-
 name: .NET
 
 on:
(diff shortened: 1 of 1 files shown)


Omnivex3:
Closing this PR. Thanks for your time.

PR #8211, opened 2026-05-27
#8210 Fix file descriptor leak in PictureService.GetDefaultPictureUrl…

### LINKED ISSUE #8210: File descriptor leak in PictureService.GetDefaultPictureUrlAsync — SKCodec.Create not disposed
nopCommerce version: 4.90.4 (also reproducible on `develop`)

  1. Deploy nopCommerce 4.90 to a Linux container with default `ulimit -n` (e.g. 1024).
  2. Generate many distinct default-image thumbnail sizes — easiest reproduction is a catalog browse over many categories/products that lack their own picture, OR a picture import that deletes and regenerates thumbs (`picture.IsNew` path).
  3. Watch open file descriptors of the process grow:
     ```
     ls /proc/$(pidof dotnet)/fd | wc -l
     ```
  4. After hours of sustained traffic the count approaches `ulimit -n`. Eventually cascading `IOException` is logged ("ErrorPage.htm cannot be opened" — the default error page itself cannot be opened to render the 500 response), the healthcheck
  fails, and the orchestrator restarts the container.

  ---
  
  ## Root cause

  In `Nop.Services/Media/PictureService.cs`, inside `GetDefaultPictureUrlAsync` (~line 468), `SKCodec.Create(filePath)` is **not** wrapped in `using`:

  ```csharp
  using var image = SKBitmap.Decode(filePath);   // ✓ disposed
  var codec = SKCodec.Create(filePath);          // ✗ leaks — IDisposable, holds open file handle
  var format = codec.EncodedFormat;
  ```

  `SKCodec` is `IDisposable` and keeps an open file handle to the default image (`wwwroot/images/default-image.png` etc.) until the GC finalizer runs. Under sustained load this accumulates file descriptors faster than Gen2 GC reclaims them,
  eventually exhausting `ulimit -n`.

  The two adjacent `SKCodec.Create()` calls in the same file dispose correctly:
  - `PictureService.cs:609` — `using var codec = SKCodec.Create(inputStream);`
  - `PictureService.cs:1014` — `using var codec = SKCodec.Create(input);`

  This one branch was the outlier.

  ## Where the path fires

  First-time thumb generation for default images (categories / products / customers / avatars with no own picture). Particularly:
  - After deploy (no cached thumbs in `wwwroot/images/thumbs/`)
  - After thumb directory cleanup
  - During picture imports — `picture.IsNew` triggers `DeletePictureThumbsAsync` → next request regenerates → leak path fires

  ## Production symptom we observed

  | Time | Event | Effect |
  |---|---|---|
  | 06:27 → 13:44 | Steady traffic | Container memory grew 3420 → 4308 MB (~12 MB/h leak) |
  | 14:09 | Process hit `ulimit -n` | Burst `IOException: ErrorPage.htm cannot be opened` |
  | 14:10 | Healthcheck fail | Swarm spawned replacement; new container leaked too |
  | 14:25 | Manual restart | Stable at ~1.2 GB |

  ## Environment

  - nopCommerce 4.90.4
  - .NET 9
  - SkiaSharp 3.119
  - Docker Swarm / Linux

  ## Fix
  
  Single `using` keyword. PR with regression test (100 iterations → before fix: ~48 leaked fds; after fix: 0): #8211

bambuca:
Closes #8210

  ## Problem

  `SKCodec.Create(filePath)` in `PictureService.GetDefaultPictureUrlAsync` (~line 468) is not wrapped in `using`:

  ```csharp
  using var image = SKBitmap.Decode(filePath);   // ✓ disposed
  var codec = SKCodec.Create(filePath);          // ✗ leaked — IDisposable
  var format = codec.EncodedFormat;
  ```

  `SKCodec` is `IDisposable` and holds an open file handle to the default image until the GC finalizer runs. Under sustained load this accumulates file descriptors and eventually exhausts `ulimit -n`, causing cascading `IOException` and container
  restarts. See #8210 for full reproduction and production symptom.

  ## Why it slipped

  The two adjacent `SKCodec.Create()` calls in the same file already use `using` correctly:
  - `PictureService.cs:609` — `using var codec = SKCodec.Create(inputStream);`
  - `PictureService.cs:1014` — `using var codec = SKCodec.Create(input);`

  This branch was the outlier.

  ## Fix

  Single keyword:

  ```diff
                   using var image = SKBitmap.Decode(filePath);
  -                var codec = SKCodec.Create(filePath);
  +                using var codec = SKCodec.Create(filePath);
  ```

  ## Regression test
  
  Added `Nop.Services.Tests/Media/PictureServiceFileDescriptorTests`:
  - Runs 100 iterations of `GetDefaultPictureUrlAsync` with unique `targetSize` values (each iteration triggers the leak code path because the thumb file doesn't exist yet)
  - Snapshots open fd count via `/proc/self/fd` (Linux) or `/dev/fd` (macOS) **before** forcing GC, so finalizers don't mask the live leak
  - Cleans up generated thumb files in `[OneTimeTearDown]`
  - Skips on Windows (no `/proc` or `/dev/fd`)

  | | Before fix | After fix |
  |---|---|---|
  | `leakedLive` after 100 iter | **48** | **0** (deterministic across 3 runs) |
  | `leakedAfterGc` | 0 (GC reaped) | 0 |

  ## Out of scope (worth a separate look)

  `PictureService.cs:1014` and `:1017` also have `SKBitmap.Decode(...)` not wrapped in `using`. These leak Skia native heap (not file descriptors) and the test doesn't cover them — happy to file a separate PR if maintainers want.

### DIFF
--- a/src/Libraries/Nop.Services/Media/PictureService.cs
+++ b/src/Libraries/Nop.Services/Media/PictureService.cs
@@ -465,7 +465,7 @@ public partial class PictureService : IPictureService
             try
             {
                 using var image = SKBitmap.Decode(filePath);
-                var codec = SKCodec.Create(filePath);
+                using var codec = SKCodec.Create(filePath);
                 var format = codec.EncodedFormat;
                 var pictureBinary = ImageResize(image, format, targetSize);
                 var mimeType = GetMimeTypeFromFileName(thumbFileName);
--- a/src/Tests/Nop.Tests/Nop.Services.Tests/Media/PictureServiceFileDescriptorTests.cs
+++ b/src/Tests/Nop.Tests/Nop.Services.Tests/Media/PictureServiceFileDescriptorTests.cs
@@ -1,0 +1,99 @@
+using FluentAssertions;
+using Nop.Core.Domain.Media;
+using Nop.Services.Media;
+using NUnit.Framework;
+
+namespace Nop.Tests.Nop.Services.Tests.Media;
+
+[TestFixture]
+public class PictureServiceFileDescriptorTests : ServiceTest
+{
+    private IPictureService _pictureService;
+    private IThumbService _thumbService;
+    private readonly List<string> _generatedThumbPaths = [];
+
+    [OneTimeSetUp]
+    public void SetUp()
+    {
+        _pictureService = GetService<IPictureService>();
+        _thumbService = GetService<IThumbService>();
+    }
+
+    [OneTimeTearDown]
+    public void TearDown()
+    {
+        foreach (var path in _generatedThumbPaths)
+        {
+            try
+            {
+                if (File.Exists(path))
+                    File.Delete(path);
+            }
+            catch
+            {
+                // ignored — best-effort cleanup
+            }
+        }
+    }
+
+    [Test]
+    public async Task GetDefaultPictureUrlAsync_DoesNotLeakFileDescriptors()
+    {
+        if (!OperatingSystem.IsLinux() && !OperatingSystem.IsMacOS())
+            Assert.Ignore("File descriptor counting requires /proc/self/fd (Linux) or /dev/fd (macOS).");
+
+        const int warmupIterations = 10;
+        const int testIterations = 100;
+        const int baseSize = 990_000;
+
+        for (var i = 0; i < warmupIterations; i++)
+        {
+            var size = baseSize + i;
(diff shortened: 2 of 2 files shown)


skoshelev:
Hi, @bambuca!

Thank you very much for the detailed description of the problem and its solution.

Issue #6681, opened 2023-04-21
DeadLock in ConfirmOrder


MehdiNaz:
nopCommerce version:4.6

Steps to reproduce the problem: 

When placing an order, the following error is given and some shopping cart items are not registered in the order

Error while placing order. Error 1: Transaction (Process ID 251) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction..

https://mysite.com/RealOnePageCheckout/ConfirmOrder

MehdiNaz:
image error in sql profiler  :
https://ibb.co/vPDKkPQ

Deadlock graph	1328351	1	sa	0X01	17	.\MSSQLSERVER,1433		2023-04-22 03:02:41.483	<deadlock-list>
 <deadlock victim="process23cbf9e8108">
  <process-list>
   <process id="process23cbf9e8108" taskpriority="0" logused="232" waitresource="KEY: 8:72057597023551488 (f69b74d115f6)" waittime="726" ownerId="5389480823" transactionname="MERGE" lasttranstarted="2023-04-22T03:02:40.737" XDES="0x23c4dac8470" lockMode="U" schedulerid="18" kpid="43588" status="suspended" spid="370" sbid="0" ecid="0" priority="0" trancount="3" lastbatchstarted="2023-04-22T03:02:40.737" lastbatchcompleted="2023-04-22T03:02:40.743" lastattention="1900-01-01T00:00:00.743" clientapp="Core Microsoft SqlClient Data Provider" hostname="WIN-G6CNI51E14D" hostpid="42200" loginname="....." isolationlevel="read committed (2)" xactid="5389480823" currentdb="8" currentdbname="....." lockTimeout="4294967295" clientoption1="671219744" clientoption2="128056">
    <stackFrames>
[...]

Issue #6482, opened 2022-12-22
SQL Deadlocks during order creation


skoshelev:
nopCommerce version: 4.50+

On a relatively busy site (60-100 orders/day), we are getting 2-3 deadlocks per day, all on URL https://...../checkout/confirm and the result is that the payment has been taken on the gateway, but the order receipt not emailed and the order left in "Pending" state and payment state is "Paid" with one of the payment methods, or "Pending" with the other two - not the expected "Processing" (NB. with one of the payment methods the payment is not taken so we sometimes lose a sale if the customer doesn't try again).

The NOP admin log has the following each time the issue occurs:

```
Microsoft.Data.SqlClient.SqlException (0x80131904): Transaction (Process ID 146) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted)
   at Microsoft.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader(Boolean isInternal, Boolean forDescribeParameterEncryption)
   at Microsoft.Data.SqlClient.SqlCommand.InternalEndExecuteNonQuery(IAsyncResult asyncResult, Boolean isInternal, String endMethod)
   at Microsoft.Data.SqlClient.SqlCommand.EndExecuteNonQueryInternal(IAsyncResult asyncResult)
   at Microsoft.Data.SqlClient.SqlCommand.EndExecuteNonQueryAsync(IAsyncResult asyncResult)
[...]

PR #8213, opened 2026-05-27
#8212 Fix SQL deadlock in EntityRepository bulk Insert/Update/Delete

### LINKED ISSUE #8212: SQL deadlocks in EntityRepository bulk methods — default Serializable isolation in TransactionScope
nopCommerce version: 4.90.x (also reproducible on `develop`)

  1. Deploy nopCommerce 4.90 to SQL Server with concurrent checkout traffic (60+ orders/hour).
  2. Watch SQL Server's `system_health` extended events session or `sys.dm_tran_locks` during checkout confirm.
  3. Within hours, deadlock graphs appear with `isolationlevel="serializable"` and `RangeS-S` / `RangeI-N` lock types on `Product`, `Product_SpecificationAttribute_Mapping`, or `GenericAttribute` tables.
  4. The affected request typically completes with `SqlException: Transaction (Process ID N) was deadlocked on lock resources` — the checkout fails after payment was processed by the gateway, leaving inconsistent order state (paid but pending, no
  confirmation email).

  ---
  
  ## Root cause

  `Nop.Data/EntityRepository.cs` has 6 occurrences of:

  ```csharp
  using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
  ```

  That constructor uses **default `TransactionOptions`**, which sets `IsolationLevel = Serializable`. Microsoft engineer David Browne addresses this exact pattern in his blog post **["using new TransactionScope() Considered 
  Harmful"](https://learn.microsoft.com/en-us/archive/blogs/dbrowne/using-new-transactionscope-considered-harmful)**:

  > *"TransactionScope's default constructor is, for the purposes of SQL Server database programming, **broken**. (...) in SQL Server SERIALIZABLE transactions are rarely useful and extremely deadlock-prone. (...) since the TransactionScope is the 
  recommended way to manage transactions in .NET, its default constructor is **setting up SQL Server applications to be deadlock-prone**."*

  On SQL Server, Serializable isolation acquires **range locks** (`RangeS-S` for reads, `RangeI-N` for inserts) and holds them until commit. For nopCommerce's bulk Insert/Update/Delete paths this means range locks on entire index ranges of
  `Product`, `GenericAttribute`, and related tables — even though the operations only need row-level atomicity (already guaranteed by the transaction itself).

  Under concurrent load this produces the deadlock pattern observed.

  ## Affected hot paths
  
  The 6 bulk-method `TransactionScope` instances are reached from:
  - `OrderProcessingService.ResetCheckoutDataAsync` — **every checkout confirm** (root cause of #6482)
  - `CustomerActivityService` — activity log writes
  - `QueuedEmailService` — email queue processing
  - `LocalizationService` bulk save — plugin startup resource load, admin save localized fields
  - `UrlRecordService` cascade delete
  - `ProductService.UpdateProductsAsync(products)` — Solr reindex, stock sync
  - `ProductWarehouseInventoryService` stock recalc — after every placed order

  ## Production data — before/after applying the fix
  
  Equivalent fix deployed to one production store (Koscom e-shop fork) on **2026-05-20 ~09:00**:

  | 24-hour window | Deadlocks |
  |---|---|
  | 16.5. 9:00 → 17.5. 9:00 | 51 |
  | 17.5. 9:00 → 18.5. 9:00 | 9 |
  | 18.5. 9:00 → 19.5. 9:00 | 24 |
  | 19.5. 9:00 → 20.5. 9:00 | **162** |
  | **20.5. 9:00 → 21.5. 9:00** | **0** ← fix deployed |
  | 21.5. 9:00 → 22.5. 9:00 | 0 |
  | 22.5. 9:00 → 23.5. 9:00 | 0 |
  | 23.5. 9:00 → 24.5. 9:00 | 0 |
  | 24.5. 9:00 → 25.5. 9:00 | 0 |
  | 25.5. 9:00 → 26.5. 9:00 | 0 |
  | 26.5. 9:00 → 27.5. 9:00 | 0 |

  **246 deadlocks in 4 days → 0 deadlocks in 7+ days**, with the inflection point exactly at the deploy time.

  ## Relation to prior issues
  
  This addresses #6482 and #6681 (both closed). #6482 was closed with *"I could not even find a theoretical possibility to repeat this problem"*, but the comment thread shows **7 independent users confirming the same deadlock** across versions 4.0,
   4.5, 4.6, and 4.60.3 — including a report of customers being double-charged when retrying after the deadlock-aborted order. #6681 was marked as duplicate of #6482.

  This issue provides what was missing:
  - **theoretical mechanism** (default Serializable + range locks)
  - **empirical evidence** (deadlock graph with explicit `isolationlevel="serializable"`)
  - **production verification** (before/after table above)

  ## Environment
  
  - nopCommerce 4.90.x (confirmed); also affects 4.50–4.80 and current `develop` (3 occurrences after refactor consolidation)
  - .NET 9
  - SQL Server (any version supporting range locks — all)
  - `READ_COMMITTED_SNAPSHOT = ON` (typical for nopCommerce deployments)

  ## Proposed fix

  Add a small helper that creates a `TransactionScope` with explicit `IsolationLevel.ReadCommitted` and `Timeout = TransactionManager.MaximumTimeout`, matching the MS-recommended pattern from the blog post above. Apply it to all 6 bulk paths.

  PR: #8213

bambuca:
Closes #8212
  Refs #6482 #6681

  ## Problem

  All 6 `TransactionScope` instances in `EntityRepository.cs` use the default constructor:

  ```csharp
  using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
  ```

  This defaults `IsolationLevel` to `Serializable` — a known footgun on SQL Server, addressed in detail by Microsoft engineer David Browne's blog post **["using new TransactionScope() Considered 
  Harmful"](https://learn.microsoft.com/en-us/archive/blogs/dbrowne/using-new-transactionscope-considered-harmful)**:

  > *"in SQL Server SERIALIZABLE transactions are rarely useful and extremely deadlock-prone (...) its default constructor is setting up SQL Server applications to be deadlock-prone."*

  On SQL Server this produces range locks (RangeS-S / RangeI-N) held for the duration of bulk Insert/Update/Delete operations, leading to deadlocks under concurrent load (most visibly `ResetCheckoutDataAsync` during checkout confirm).

  See #8212 for the full reproduction, hot paths, relation to #6482/#6681 (both previously closed for lack of reproduction — we now provide both the theoretical mechanism and empirical evidence), and production before/after data showing **246 
  deadlocks → 0** at deploy time.

  ## Fix

  Adds a private helper `CreateReadCommittedScope()` matching the MS-recommended pattern from the blog post above:

  ```csharp
  private static TransactionScope CreateReadCommittedScope()
  {
      return new TransactionScope(TransactionScopeOption.Required,
          new TransactionOptions
          {
              IsolationLevel = IsolationLevel.ReadCommitted,
              Timeout = TransactionManager.MaximumTimeout
          },
          TransactionScopeAsyncFlowOption.Enabled);
  }
  ```

  Applied to all 6 bulk paths:
  - `InsertAsync(IList<TEntity>)` / `Insert(IList<TEntity>)` — bulk insert
  - `DeleteAsync(IList<TEntity>)` / `Delete(IList<TEntity>)` — bulk delete (entities)
  - `DeleteAsync(Expression<Func<TEntity, bool>>)` / `Delete(Expression<Func<TEntity, bool>>)` — bulk delete (predicate)

  ## Why ReadCommitted is safe here

  The bulk operation is itself a single LinqToDB call wrapped in the transaction. Row-level atomicity (each row's INSERT/UPDATE/DELETE is atomic) is guaranteed by the transaction itself. Serializable's range-lock semantics — preventing phantom
  reads across the affected index range — is not needed for these write-only paths, because there is no application-level read to re-execute.

  Under `READ_COMMITTED_SNAPSHOT = ON` (typical for nopCommerce deployments) readers don't block writers at all, and writers among themselves operate with row-level locks instead of range locks.

  ## Timeout — transparent disclosure of behavioral change
  
  The previous single-argument `TransactionScope(TransactionScopeAsyncFlowOption)` constructor **does not set a scope-level timeout** (no `_scopeTimer` is created). The 3-argument constructor we use here passes `TransactionOptions` through
  `TransactionManager.ValidateTimeout`, which maps the default `TimeSpan.Zero` to `MaximumTimeout` (10 minutes).

  To make this explicit and match the MS-recommended pattern, the helper sets `Timeout = TransactionManager.MaximumTimeout` directly. In practice this scope-level value is masked by `SqlCommand.CommandTimeout` (30s default), which fires first —
  verified in production without any observable behavioral change.

  ## Production verification

  See #8212 for the full before/after table. Summary: an equivalent fix deployed to one production store on **2026-05-20 ~09:00** dropped the deadlock rate from **246 deadlocks in 4 days before** to **0 deadlocks in 7+ days after**, with the
  inflection point exactly at the deploy time.

  ## Risk

  **Minimal.** ReadCommitted was the SQL Server default for nopCommerce code running outside `TransactionScope` wrappers — this fix only restores that default for the 6 bulk paths. Existing CRUD tests (`EntityRepositoryTests`) pass unchanged.

  ## Why no new test added
  
  A behavioral test would require mocking `INopDataProvider` to capture `Transaction.Current.IsolationLevel` at runtime — adding more code than the fix itself, and asserting on TransactionScope internals that aren't part of any public contract.
  Happy to add one if the maintainers prefer; the change is otherwise small enough to verify by inspection.

### DIFF
--- a/src/Libraries/Nop.Data/EntityRepository.cs
+++ b/src/Libraries/Nop.Data/EntityRepository.cs
@@ -123,6 +123,35 @@ public partial class EntityRepository<TEntity> : IRepository<TEntity> where TEnt
         return query.OfType<ISoftDeletedEntity>().Where(entry => !entry.Deleted).OfType<TEntity>();
     }
 
+    /// <summary>
+    /// Returns a <see cref="TransactionScope"/> with explicit <see cref="IsolationLevel.ReadCommitted"/>.
+    /// The default <c>new TransactionScope(...)</c> constructor uses <see cref="IsolationLevel.Serializable"/>,
+    /// which holds range locks (RangeS-S / RangeI-N on SQL Server) for the duration of bulk Insert/Update/Delete
+    /// and is the root cause of the deadlocks reported in #6482 and #6681. Row-level atomicity is guaranteed
+    /// by the transaction itself; range locks are not needed for these write paths.
+    /// <para>
+    /// Timeout is set explicitly to <see cref="TransactionManager.MaximumTimeout"/> per the MS-recommended pattern
+    /// — this matches the effective behavior of the previous single-argument <c>TransactionScope</c> constructor
+    /// (which left the scope timer unset). In practice the per-command <c>SqlCommand.CommandTimeout</c> (30s default)
+    /// fires first, so this scope-level value is a generous upper bound.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// See David Browne (Microsoft), <see href="https://learn.microsoft.com/en-us/archive/blogs/dbrowne/using-new-transactionscope-considered-harmful">
+    /// "using new TransactionScope() Considered Harmful"</see>:
+    /// <em>"in SQL Server SERIALIZABLE transactions are rarely useful and extremely deadlock-prone … its default constructor is setting up SQL Server applications to be deadlock-prone."</em>
+    /// </remarks>
+    private static TransactionScope CreateReadCommittedScope()
+    {
+        return new TransactionScope(TransactionScopeOption.Required,
+            new TransactionOptions
+            {
+                IsolationLevel = IsolationLevel.ReadCommitted,
+                Timeout = TransactionManager.MaximumTimeout
+            },
+            TransactionScopeAsyncFlowOption.Enabled);
+    }
+
     #endregion
 
     #region Methods
@@ -441,7 +470,7 @@ public partial class EntityRepository<TEntity> : IRepository<TEntity> where TEnt
     {
         ArgumentNullException.ThrowIfNull(entities);
 
-        using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+        using var transaction = CreateReadCommittedScope();
         await _dataProvider.BulkInsertEntitiesAsync(entities);
         transaction.Complete();
 
@@ -462,7 +491,7 @@ public partial class EntityRepository<TEntity> : IRepository<TEntity> where TEnt
     {
         ArgumentNullException.ThrowIfNull(entities);
 
-        using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+        using var transaction = CreateReadCommittedScope();
         _dataProvider.BulkInsertEntities(entities);
         transaction.Complete();
 
@@ -632,7 +661,7 @@ public partial class EntityRepository<TEntity> : IRepository<TEntity> where TEnt
         if (!entities.Any())
             return;
 
-        using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+        using var transaction = CreateReadCommittedScope();
         
         if (typeof(TEntity).GetInterface(nameof(ISoftDeletedEntity)) == null)
             await _dataProvider.BulkDeleteEntitiesAsync(entities);
@@ -666,7 +695,7 @@ public partial class EntityRepository<TEntity> : IRepository<TEntity> where TEnt
         if (!entities.Any())
             return;
 
-        using var transaction = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
+        using var transaction = CreateReadCommittedScope();
 
         if (typeof(TEntity).GetInterface(nameof(ISoftDeletedEntity)) == null)
             _dataProvider.BulkDeleteEntities(entities);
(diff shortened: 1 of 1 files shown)


skoshelev:
see details on [this commit](https://github.com/nopSolutions/nopCommerce/commit/3a1db40a68d06afa177bf2649dec4841fa79d56b)

PR #8215, opened 2026-05-28
#8214 Cache compiled property selectors in LocalizationService

### LINKED ISSUE #8214: DynamicResolver accumulation in LocalizationService.GetLocalizedAsync — keySelector.Compile() on every fallback
**nopCommerce version:** 4.90.4 (also reproducible on `develop`)
  
  **Steps to reproduce:**
  1. Deploy nopCommerce 4.90 to a Linux container with moderate memory limit (e.g. 2 GB)
  2. Populate a store where many localized entities (e.g. `SpecificationAttributeOption.Name`, `Category.Name`) have a value in the entity's own column but no
  matching `LocalizedProperty` row for the current working language — i.e. the default-value fallback path inside `GetLocalizedAsync` is hit often
  3. Send sustained traffic to a page that enumerates such entities per request (product detail with spec attribute options is a typical hot path)
  4. Observe a steadily growing count of `System.Reflection.Emit.DynamicResolver` (and `DynamicResolver+DestroyScout`) objects in the heap and finalizer queue;
   under burst load the runtime can throw a preallocated `OutOfMemoryException` (zero-IP, empty message, `HResult 0x8007000E`) even while the container memory
  metric stays well below the limit
  
  **Root cause:**
  `LocalizationService.GetLocalizedAsync<TEntity, TPropType>` calls `keySelector.Compile()` on the default-value fallback path (`LocalizationService.cs` around
   line 604):
  
  ```csharp
  //set default value if required
  if (!string.IsNullOrEmpty(resultStr) || !returnDefaultValue)
      return result;
  var localizer = keySelector.Compile();
  result = localizer(entity);
  ```
  
  Every call to `LambdaExpression.Compile()` emits a new `DynamicMethod` backed by a `DynamicResolver` (and a finalizable `DestroyScout`) holding native IL
  code. Because the method is on the fallback path of a very hot read API, and the same `e => e.PropName` selectors are called over and over from
  factories/event consumers, the heap and finalizer queue grow under load faster than GC and the finalizer thread can reclaim them. Eventually the runtime hits
   an internal allocation failure and throws the preallocated `OutOfMemoryException`.

  Production heap snapshot (a single ASP.NET worker just before SIGABRT):
  - 33 456 `System.Reflection.Emit.DynamicResolver` instances
  - 23 099 `DynamicResolver+DestroyScout` ready for finalization
  - 22 508 objects ready for finalization total
  - 40 756 finalizable objects in Gen1
  - 14 of 74 threads stuck in `DestroyScout.Finalize()`; 23 in `Expression.Compile()` / `LambdaCompiler..ctor` / `DynamicMethod.GetMethodDescriptor`
  **Similar pattern elsewhere:**
  `ArtificialIntelligenceService.GetTitleAndTextAsync` (around `ArtificialIntelligenceService.cs:101–102`) has the same `titleSelector.Compile()` /
  `textSelector.Compile()` shape and is affected by the same problem, although on a much colder path.
  
  **Why it's safe to cache:**
  `GetLocalizedAsync` already requires the selector to be a `MemberExpression` referring to a `PropertyInfo` (`LocalizationService.cs:570–574`). That means
  `(typeof(TEntity), propInfo.Name)` is a deterministic, semantically-correct cache key for the compiled delegate — two callers passing different `Expression`
  instances of `e => e.Name` would compile to identical IL.
  
  **Fix:** see accompanying PR #8215— introduces a `PropertySelectorCache` helper keyed by `(EntityType, PropertyName)`, replaces both call sites, falls back
  to a plain compile for non-property selectors (no behavioural change).

bambuca:
Fixes #8214 

  `LocalizationService.GetLocalizedAsync` calls `keySelector.Compile()` on the default-value fallback path on every invocation, emitting a new `DynamicMethod`
  each time and creating significant GC + finalizer pressure under load (see linked issue for production heap details).

  ## Change

  Introduce `PropertySelectorCache` in `Nop.Services.Localization` — a process-wide `ConcurrentDictionary` of compiled delegates keyed by `(EntityType,
  PropertyName)`. The key is semantically correct because `GetLocalizedAsync` already validates that the selector is a `MemberExpression` on a `PropertyInfo`
  (`LocalizationService.cs:570–574`), so two callers passing different `Expression` instances of `e => e.PropName` will produce identical IL.
  
  Applied to:
  - `LocalizationService.cs` (primary hot path)
  - `ArtificialIntelligenceService.GetTitleAndTextAsync` (same pattern, much colder path — included for consistency)

  Selectors that aren't a simple property access (e.g. method calls, casts) fall back to a plain `selector.Compile()` — no behavioural change for that edge
  case.

  ## Safety

  - Zero changes at call sites; ~100 existing `GetLocalizedAsync(entity, x => x.Prop)` invocations across the solution are unaffected
  - Dictionary growth is bounded by the number of distinct `(EntityType, PropertyName)` pairs reachable from the codebase (tens of entries in practice)
  - `PropertyInfo` is stable for the lifetime of the process, so the cached delegate stays valid

  ## Test
  
  Added `Tests/Nop.Tests/Nop.Services.Tests/Localization/PropertySelectorCacheTests.cs` — 7 correctness tests + 1 reporting benchmark (no perf assertion,
  
  - Same selector for the same property returns the same cached delegate
  - Different property / different entity type / different property type → different delegate
  - Non-property selector (e.g. `x => x.Name.ToUpperInvariant()`) falls through to plain compile and still works
  - `null` selector throws `ArgumentNullException`
  - Cached delegate returns the same value as a fresh compile
  
  Benchmark output (10 000 iterations, net9.0 Release, M-class Apple Silicon):
  
  ```
  Plain Compile():      ~270 ms   ~46.5 MB alloc
  Cached GetCompiled():  ~13 ms    ~5.0 MB alloc
  Speedup: ~20x, allocation reduction: ~9x 
  ```
  
  The allocation figure for the cached path is dominated by per-iteration `Expression` allocations in the test loop itself; **no new `DynamicResolver` is 
  created on cache hit**, which is the property that matters for the production scenario.

### DIFF
--- a/src/Libraries/Nop.Services/ArtificialIntelligence/ArtificialIntelligenceService.cs
+++ b/src/Libraries/Nop.Services/ArtificialIntelligence/ArtificialIntelligenceService.cs
@@ -98,8 +98,8 @@ public partial class ArtificialIntelligenceService : IArtificialIntelligenceServ
         string textRequiredLocale, string titleRequiredLocale)
         where TEntity : BaseEntity, ILocalizedEntity, IMetaTagsSupported
     {
-        var getTitle = titleSelector.Compile();
-        var getText = textSelector.Compile();
+        var getTitle = PropertySelectorCache.GetCompiled(titleSelector);
+        var getText = PropertySelectorCache.GetCompiled(textSelector);
 
         if (languageId == 0)
             return await GetTitleAndTextAsync(entity, languageName, getTitle, getText, textRequiredLocale, titleRequiredLocale);
--- a/src/Libraries/Nop.Services/Localization/LocalizationService.cs
+++ b/src/Libraries/Nop.Services/Localization/LocalizationService.cs
@@ -601,7 +601,7 @@ public partial class LocalizationService : ILocalizationService
         //set default value if required
         if (!string.IsNullOrEmpty(resultStr) || !returnDefaultValue)
             return result;
-        var localizer = keySelector.Compile();
+        var localizer = PropertySelectorCache.GetCompiled(keySelector);
         result = localizer(entity);
 
         return result;
--- a/src/Libraries/Nop.Services/Localization/PropertySelectorCache.cs
+++ b/src/Libraries/Nop.Services/Localization/PropertySelectorCache.cs
@@ -1,0 +1,49 @@
+﻿using System.Collections.Concurrent;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace Nop.Services.Localization;
+
+/// <summary>
+/// Caches delegates compiled from property selector expressions (<c>e =&gt; e.PropertyName</c>)
+/// </summary>
+/// <remarks>
+/// Avoids repeated calls to <see cref="LambdaExpression.Compile"/> for the same property selector,
+/// which would otherwise emit a new <c>DynamicMethod</c> on every invocation and create significant
+/// GC and finalizer pressure under load. Cache key is <c>(EntityType, PropertyName)</c>; selectors
+/// that do not target a property fall back to a plain compile.
+/// </remarks>
+public static class PropertySelectorCache
+{
+    #region Fields
+
+    private static readonly ConcurrentDictionary<(Type EntityType, string PropertyName), Delegate> _cache = new();
+
+    #endregion
+
+    #region Methods
+
+    /// <summary>
+    /// Get a compiled delegate for the given property selector, reusing a previously cached one
+    /// when the selector targets the same property on the same entity type
+    /// </summary>
+    /// <typeparam name="TEntity">Entity type</typeparam>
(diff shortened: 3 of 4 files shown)


skoshelev:
Hi, @bambuca.

I merged your code with [this commit](https://github.com/nopSolutions/nopCommerce/commit/15ca7d375d67184e441cda577acb5b81e092ee5a) and added my [refactoring](https://github.com/nopSolutions/nopCommerce/commit/caad5327ac5f8ccda09be9c93272323573e2d6d3) on top of it. You can examine it if you'd like. Essentially, I moved the method from a separate class to an existing one and formatted it as an extension method. Thank you very much

bambuca:
Thank you. I like your refactor.

Issue #8093, opened 2026-01-30
Advanced phone number validation


AndreiMaz:
Consider phone validation with https://github.com/twcclegg/libphonenumber-csharp

Offical C# port of https://github.com/googlei18n/libphonenumber

It could be helpful with https://github.com/nopSolutions/nopCommerce/issues/2430

mykh-hailo:
@AndreiMaz As my first contribution for nopCommerce, can I work on this issue please?

DmitriyKulagin:
Closed #8093

PR #8216, opened 2026-05-29
#8093 Use libphonenumber-csharp for validation

Zalhera:
Replace regex validation with libphonenumber validation.


DmitriyKulagin:
Your implementation based on obtaining the region code from the customer's settings wasn't part of our plans.

`var regionCode = _customerSettings.DefaultCountryId.HasValue`

It would have been more accurate to calculate the region code based on the phone number.

Perhaps this was due to an inaccurate task description, but we have already implemented this task ourselves and are grateful for your willingness to contribute.

PR #8221, opened 2026-06-02
Use UseAuthentication() extension to prevent duplicate auth middleware under minimal hosting

### LINKED ISSUE #8220: Duplicate authentication middleware under minimal hosting breaks redirect_uri scheme behind a reverse proxy
### Description

`UseNopAuthentication` registers authentication with `application.UseMiddleware<AuthenticationMiddleware>()` instead of the `application.UseAuthentication()` extension.

Under minimal hosting (`WebApplication`), only `UseAuthentication()` sets the `__AuthenticationMiddlewareSet` marker in `app.Properties` that suppresses the framework's automatic insertion of authentication middleware. Because the marker is never set, `WebApplication` concludes authentication was not configured and **auto-inserts a second `AuthenticationMiddleware` ahead of nopCommerce's own pipeline** — notably before `UseForwardedHeaders`.

### Impact

Remote-authentication callbacks (any `OAuthHandler`/OpenID provider redirect) are handled by that front instance **before `UseForwardedHeaders` runs**, so `Request.Scheme` reflects the raw backend connection scheme rather than `X-Forwarded-Proto`. When nopCommerce runs behind a TLS-terminating reverse proxy that forwards to an HTTP backend, the remote handler builds the token-exchange `redirect_uri` as `http://…`, while the authorize `redirect_uri` (built later in the pipeline, after forwarded headers) is `https://…`. The provider rejects the mismatch — Google `redirect_uri_mismatch`, Facebook error 191 — and external/social login fails.

The bug is latent in the common single-host HTTPS setup (the front instance happens to observe `https`) and only surfaces behind a reverse proxy with an HTTP backend.

### Steps to reproduce

1. Host nopCommerce behind a TLS-terminating reverse proxy (IIS ARR, nginx, …) forwarding to an HTTP backend; set `Hosting.UseProxy = true` and forward `X-Forwarded-Proto`.
2. Configure an OAuth external-authentication provider whose redirect URI is the site's `https` URL.
3. Attempt to log in via the provider → token exchange fails with `redirect_uri_mismatch` (Google) / error 191 (Facebook).

### Root cause

`UseMiddleware<AuthenticationMiddleware>()` does not set the marker that `UseAuthentication()` sets, so minimal hosting's auto-middleware injects a duplicate `AuthenticationMiddleware` at the front of the pipeline, ahead of `UseForwardedHeaders`. The provider callback is therefore processed with the unproxied scheme, and the resulting token-exchange `redirect_uri` does not match the authorize one.

### Affected versions

Introduced in **4.50.0** (the switch to minimal hosting / `WebApplication`) and present in every release since, including current `develop`. 4.40.x and earlier (Generic Host) are unaffected.

### Suggested fix

Register the middleware via `application.UseAuthentication()` so the marker is set and the duplicate front instance is suppressed, leaving only the intended instance (between `UseRouting` and `UseEndpoints`), which runs after forwarded headers. PR to follow.

professor-k:
Fixes #8220

### Problem

`UseNopAuthentication` registers authentication with `application.UseMiddleware<AuthenticationMiddleware>()`. Under minimal hosting (`WebApplication`), the `UseAuthentication()` extension is what sets the `__AuthenticationMiddlewareSet` marker that prevents the framework from auto-adding authentication middleware. Without it, `WebApplication` auto-inserts a **second** `AuthenticationMiddleware` at the front of the pipeline, before `UseForwardedHeaders`.

### Effect

Remote-auth provider callbacks are handled by that front instance before forwarded headers run, so the token-exchange `redirect_uri` is built from the raw backend scheme (`http://`) while the authorize `redirect_uri` is `https://`. Behind a reverse proxy with an HTTP backend this breaks OAuth/social login (Google `redirect_uri_mismatch`, Facebook error 191). Full details and reproduction in #8220.

### Fix

Register via the `application.UseAuthentication()` extension. One-line change; the duplicate front instance is suppressed and the remaining instance (between `UseRouting` and `UseEndpoints`) runs after forwarded headers.

### Affected versions

Introduced in 4.50.0 (the switch to minimal hosting / `WebApplication`); present in every release since, including current `develop`. 4.40.x and earlier (Generic Host) are unaffected.

### Testing

Verified on an IIS ARR blue/green deployment (TLS-terminating proxy in front of HTTP loopback backends): external + social login (Google, Facebook) fail before the change and succeed after it; ordinary email/password login is unaffected.

### Note

The same pattern may apply to `UseAuthorization()` vs `UseMiddleware<AuthorizationMiddleware>()` if registered the same way; left out of scope here.

### DIFF
--- a/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
+++ b/src/Presentation/Nop.Web.Framework/Infrastructure/Extensions/ApplicationBuilderExtensions.cs
@@ -401,7 +401,7 @@ public static class ApplicationBuilderExtensions
         if (!DataSettingsManager.IsDatabaseInstalled())
             return;
 
-        application.UseMiddleware<AuthenticationMiddleware>();
+        application.UseAuthentication();
     }
 
     /// <summary>


exileDev:
Thanks!

PR #8242, opened 2026-07-01
Fix N+1 query in product overview models (load customer wishlists once)

### LINKED ISSUE #8241: N+1 query: product overview models re-query customer wishlists once per product
### Description

`ProductModelFactory.PrepareProductToWishlistModelAsync(Product)` calls `ICustomWishlistService.GetAllCustomWishlistsAsync(customerId)`, which hits the database. In `PrepareProductOverviewModelsAsync` this method is invoked **inside the per-product loop**, so the current customer's wishlists are re-queried once for every product in the list — even though the result depends only on the customer and is identical for every product in the request.

On a catalog/category page rendering N products this produces N identical queries:

```sql
SELECT ... FROM [CustomWishlist] WHERE [CustomerId] = @customerId ORDER BY [CreatedOnUtc] DESC
```

### Impact

Observed on a live store under crawler load: ~56,000 of these queries across ~2,000 pages (~28 per page = the number of products per listing page). Each query is individually cheap, but the volume of app↔DB round-trips adds measurable per-page overhead. `GetAllCustomWishlistsAsync` is not cached, so nothing absorbs the repetition.

### Steps to reproduce

1. Enable multiple wishlists (`ShoppingCartSettings.AllowMultipleWishlist = true`).
2. Open any catalog page listing several products (category, search, manufacturer).
3. In the SQL/query log, observe one `SELECT ... FROM CustomWishlist` per product on the page, all with the same `@customerId`.

### Expected behavior

The customer's wishlists should be loaded once per request and reused for every product — one query per page instead of one per product.

### Affected version

`develop` (and 4.90.x, where the multiple-wishlists feature was introduced).

A fix is ready (load the wishlists once before the loop and reuse them via a synchronous model-builder overload); PR to follow.

professor-k:
Fixes #8241

### Problem

`PrepareProductOverviewModelsAsync` calls `PrepareProductToWishlistModelAsync(product)` inside its per-product loop, and that method queries the current customer's custom wishlists via `ICustomWishlistService.GetAllCustomWishlistsAsync`. The wishlist set depends only on the customer, so it is identical for every product in the request — a listing page of N products issues N identical `SELECT ... FROM CustomWishlist WHERE CustomerId = @id` queries (observed ~28/page, ~56k over a 2,000-page crawl).

### Fix

Load the customer's wishlists **once** before the loop and reuse them via a new synchronous `PrepareProductToWishlistModel(product, currentWishlists)` overload. The existing async `PrepareProductToWishlistModelAsync(product)` delegates to it, so the single-product path (product details page) is unchanged.

On an N-product listing page this reduces `CustomWishlist` queries from N to 1. Behaviour is otherwise identical — no functional or schema change.

### DIFF
--- a/src/Presentation/Nop.Web/Factories/ProductModelFactory.cs
+++ b/src/Presentation/Nop.Web/Factories/ProductModelFactory.cs
@@ -789,13 +789,29 @@ public partial class ProductModelFactory : IProductModelFactory
     {
         ArgumentNullException.ThrowIfNull(product);
 
+        //custom wishlists
+        var currentCustomer = await _workContext.GetCurrentCustomerAsync();
+        var currentWishlists = await _customWishlistService.GetAllCustomWishlistsAsync(currentCustomer.Id);
+
+        return PrepareProductToWishlistModel(product, currentWishlists);
+    }
+
+    /// <summary>
+    /// Prepare the product to wishlist model from an already-loaded set of the customer's wishlists.
+    /// Callers that render many products in one request (e.g. catalog listings) should load the
+    /// wishlists once and reuse them here, rather than querying once per product.
+    /// </summary>
+    /// <param name="product">Product</param>
+    /// <param name="currentWishlists">The current customer's custom wishlists</param>
+    /// <returns>The product add to wishlist model</returns>
+    protected virtual ProductToWishlistModel PrepareProductToWishlistModel(Product product, IList<CustomWishlist> currentWishlists)
+    {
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi, @professor-k!

Thank you for your help. We used your code as a basis, but made some minor changes and improvements. If you'd like, you can see the details in [this commit](https://github.com/nopSolutions/nopCommerce/commit/862960e04f6063a9cf3b8f6be170e7620918967b).

Issue #8255, opened 2026-07-09
The NopUrlHelper.RouteUrl method raises an exception if the fragment parameter is provided without the lead '#' symbol


skoshelev:
This issue was introduced in the #7859 ticket.

RinkeshKumarSinha:
Open for Review PR Link: - https://github.com/nopSolutions/nopCommerce/pull/8256

@skoshelev @AndreiMaz

skoshelev:
Closed #8255

PR #8256, opened 2026-07-11
Issue 8255: Fix NopUrlHelper.RouteUrl exception for fragment parameter

RinkeshKumarSinha:
Reference to the bug ticket: - https://github.com/nopSolutions/nopCommerce/issues/8255

Problem:
When generating URLs, the fragment value could be passed without a leading “#”. This caused URL generation to fail or behave incorrectly, especially in the .NET 10 upgrade path.

Solution:
The code now normalizes the fragment before building the URL by adding a leading “#” when it is missing. This makes URL generation more robust and ensures fragments are formatted correctly for the underlying ASP.NET routing helper.

**Dev Sanity: -**

Fragment with string "abc" without "#" - no code changes - reproduced issue screenshot: -
image

Fragment string being "abc" - replaced to #abc to avoid the exception with a check if the string starts whether # or not.

image

### DIFF
--- a/src/Presentation/Nop.Web.Framework/Mvc/Routing/NopUrlHelper.cs
+++ b/src/Presentation/Nop.Web.Framework/Mvc/Routing/NopUrlHelper.cs
@@ -208,6 +208,9 @@ public partial class NopUrlHelper : INopUrlHelper
         if (httpContext is null)
             return string.Empty;
 
+        if (!string.IsNullOrWhiteSpace(fragment) && !fragment.StartsWith('#'))
+            fragment = String.Concat("#", fragment);
+        
         if (!string.IsNullOrEmpty(protocol) || !string.IsNullOrEmpty(host))
         {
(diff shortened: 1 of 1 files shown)


RinkeshKumarSinha:
This change makes URL fragments safer by adding a leading “#” when it is missing. That prevents URL generation from breaking when a fragment like “reviews” is passed instead of “#reviews”.

@skoshelev  @AndreiMaz

skoshelev:
Hi, @RinkeshKumarSinha!

Thank you very much for your help. We've already [prepared](https://github.com/nopSolutions/nopCommerce/compare/develop...issue-8255-NopUrlHelper.RouteUrl-raises-exception) the similar solution and are currently testing it. 

Regarding your solution, although it's safe because a string is an immutable object, it's still best to avoid overriding input parameters unless they are explicitly declared as mutable (ref, out).

Issue #8261, opened 2026-07-24
PrepareProductPriceModelAsync: BasePricePAngVValue returns incorrect decimal value while formatted BasePricePAngV is correct


andrejfilipi:
Hi,

I found what appears to be an inconsistency in `ProductModelFactory.PrepareProductPriceModelAsync`.

The formatted base price (`BasePricePAngV`) is calculated correctly, but the corresponding decimal value (`BasePricePAngVValue`) is not.

Current code:

```csharp
model.BasePricePAngV = await _priceFormatter.FormatBasePriceAsync(product, finalPriceWithDiscountBase);
model.BasePricePAngVValue = finalPriceWithDiscountBase;

skoshelev:
The problem was introduced in #5913

skoshelev:
Closed #8261

PR #8282, opened 2026-08-18
#8283 Update the reference implementation in ProductModelFactoryTests

### LINKED ISSUE #8283: CanPreparePriceModel unit test fails
nopCommerce version: 4.90 (4.90-bug-fixes, c4d283390b)                                                                                                                                                                                           

Running Nop.Tests on 4.90-bug-fixes gives one failure, in ProductModelFactoryTests.CanPreparePriceModel:                                                                                                                                                                                                   

Expected object1PropertyValue to be 1200M because The property "ProductPriceModel.BasePricePAngVValue" of these objects is not equal, but found `<null>`.

That test doesn't assert fixed values, it compares ProductModelFactory against the reference copy in ProductModelFactoryForTest (ProductModelFactoryTests.cs:241). #8261 moved production to GetBaseProductPriceAsync, but the reference copy still assigns the raw price, so the two disagree for a product with no base price configured.

 PR: #8282

bambuca:
Fixes: #8283
  `ProductModelFactoryTests` compares production `ProductModelFactory` against a reference                                                                                                                                                         
  implementation kept in the test fixture (`ProductModelFactoryForTest`). #8261 moved production                                                                                                                                                   
  to `GetBaseProductPriceAsync`, but left the reference assigning the raw price, so                                                                                                                                                                
  `CanPreparePriceModel` fails on `4.90-bug-fixes`:                                                                                                                                                                                                
                                                                                                                                                                                                                                                   
  Expected object1PropertyValue to be 1200M because The property                                                                                                                                                                                   
  "ProductPriceModel.BasePricePAngVValue" of these objects is not equal, but found <null>.                                                                                                                                                         
                                                                                                                                                                                                                                                   
  The reference returns the raw 1200; production correctly returns `null` for a product with no                                                                                                                                                    
  base price configured.                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                   
  ## Change                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                   
  The three assignments in the reference implementation (`ProductModelFactoryTests.cs:372`, `:456`,                                                                                                                                                
  `:568`) now call `GetBaseProductPriceAsync`, matching production. Test code only — no production                                                                                                                                                 
  file is touched.

### DIFF
--- a/src/Tests/Nop.Tests/Nop.Web.Tests/Public/Factories/ProductModelFactoryTests.cs
+++ b/src/Tests/Nop.Tests/Nop.Web.Tests/Public/Factories/ProductModelFactoryTests.cs
@@ -369,7 +369,7 @@ public class ProductModelFactoryTests : WebTest
 
                         //PAngV default baseprice (used in Germany)
                         priceModel.BasePricePAngV = await _priceFormatter.FormatBasePriceAsync(product, finalPriceWithDiscount);
-                        priceModel.BasePricePAngVValue = finalPriceWithDiscount;
+                        priceModel.BasePricePAngVValue = await GetBaseProductPriceAsync(product, finalPriceWithDiscount);
                     }
                 }
                 else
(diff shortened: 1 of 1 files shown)


skoshelev:
Hi, @bambuca!

Thank you for your help

Issue #8315, opened 2026-09-22
N+1 queries in ProductModelFactory.PrepareProductOverviewModelsAsync — applied discounts and tier prices are loaded per product


bambuca:
`ProductModelFactory.PrepareProductOverviewModelsAsync` ([ProductModelFactory.cs:1361-1369](https://github.com/nopSolutions/nopCommerce/blob/4.90-bug-fixes/src/Presentation/Nop.Web/Factories/ProductModelFactory.cs#L1361-L1369)) prepares the price model **inside the per-product loop**. For every product, `PriceCalculationService.GetFinalPriceAsync` loads its data one entity at a time:

- discounts — [PriceCalculationService.cs:79](https://github.com/nopSolutions/nopCommerce/blob/4.90-bug-fixes/src/Libraries/Nop.Services/Catalog/PriceCalculationService.cs#L79) → `DiscountService.GetAppliedDiscountsAsync(entity)` ([DiscountService.cs:262-270](https://github.com/nopSolutions/nopCommerce/blob/4.90-bug-fixes/src/Libraries/Nop.Services/Discounts/DiscountService.cs#L262-L270)); the cache behind it is per request, so this repeats on every request, products without discounts included;
- tier prices — [PriceCalculationService.cs:365](https://github.com/nopSolutions/nopCommerce/blob/4.90-bug-fixes/src/Libraries/Nop.Services/Catalog/PriceCalculationService.cs#L365) → `ProductService.GetTierPricesByProductAsync(productId)` ([ProductService.cs:2082](https://github.com/nopSolutions/nopCommerce/blob/4.90-bug-fixes/src/Libraries/Nop.Services/Catalog/ProductService.cs#L2082)); static cache, so once per product until the entry expires.

Neither service exposes a multi-entity API, so the factory cannot load the page in one round trip even though it holds the whole list.

```sql
SELECT ... FROM [Discount] [d] INNER JOIN [Discount_AppliedToProducts] [ad] ... WHERE [ad].[EntityId] = @entityId   -- N times per request
SELECT ... FROM [TierPrice] [tp] WHERE [tp].[ProductId] = @productId                                              -- N times on a cold cache
```

### Impact

For a page with N products: N discount queries on every request → 1, and N tier-price queries on a cold cache → 1 (none once cached). With N = 32 that is 32 + 32 statements replaced by 2 (the per-request cache does not help discounts — it is filled inside the same loop).

### Steps to reproduce

[...]

PR #8316, opened 2026-09-22
#8315 Load applied discounts and tier prices of a product list in batch

bambuca:
Fixes #8315 .

`PrepareProductOverviewModelsAsync` prepares the price model per product, and `GetFinalPriceAsync` loads each product's applied discounts (per-request cache) and tier prices (static cache) with a query of its own — N discount queries per request and N tier-price queries on a cold cache.

## Change

- `IDiscountService.GetAppliedDiscountsAsync<T>(IEnumerable<IDiscountSupported<T>> entities)` — every entity is read through the same `_shortTermCacheManager.GetAsync(acquire, AppliedDiscountsCacheKey, entity.GetType().Name, entity)` as the single-entity method; the acquires share one lazily executed `EntityId IN (...)` query, so cached entities cost nothing and the missing ones one query in total. Empty results are cached too. The single-entity method delegates to the batch.
- `IProductService.GetTierPricesByProductsAsync(int[] productIds)` — same shape over `_staticCacheManager.GetAsync(PrepareKeyForDefaultCache(TierPricesByProductCacheKey, id), acquire)`.
- `ProductModelFactory.PrepareProductOverviewModelsAsync` calls both before the loop when `preparePriceModel` is set and more than one product is prepared. The loop, `PriceCalculationService` and existing signatures are unchanged; `ProductModelFactory` gets an `IDiscountService` constructor dependency (last parameter).

## Safety

Same keys, value types (`List<Discount>`, `List<TierPrice>`), cache managers and lifetimes as the per-entity methods, through the same `GetAsync(key, acquire)` path — nothing is cached longer or shared more widely, and concurrent acquires are still de-duplicated by the `Lazy<Task<T>>` entry. Same approach as #8242 took for the wishlists in this loop.

## Test

`DiscountServiceTests` and `ProductServiceTests` run the batches against the real `MemoryCacheManager` / `PerRequestCacheManager` and assert that a following single-entity call returns the very same cached instance (a value stored under another generic type would make `GetTierPricesByProductAsync` throw `InvalidCastException`). Empty input and `null` entities covered; `ProductModelFactoryTests` updated for the constructor.

For a page with N products: N discount queries per request → 1, N tier-price queries on a cold cache → 1 (none once cached).


AndreiMaz:
See my comment at https://github.com/nopSolutions/nopCommerce/issues/8315#issuecomment-5788829768

PR #8320, opened 2026-09-23
#8319 Fix the casing of the Nop.Plugin.Feed.ChatGptShopping path in NopCommerce.sln

### LINKED ISSUE #8319: Docker build fails: NopCommerce.sln references Nop.Plugin.Feed.ChatGPTShopping with the wrong casing
`src/NopCommerce.sln` references the ChatGPT Shopping plugin directory as `Nop.Plugin.Feed.ChatGPTShopping`, but the actual directory on disk is `Nop.Plugin.Feed.ChatGptShopping`.

On case-sensitive filesystems, this causes MSBuild to fail because it cannot find the project:

`error MSB3202: The project file "/src/Plugins/Nop.Plugin.Feed.ChatGPTShopping/Nop.Plugin.Feed.ChatGptShopping.csproj" was not found. [/src/NopCommerce.sln]`

This also breaks the repository’s Dockerfile, since it runs `dotnet build NopCommerce.sln` on `mcr.microsoft.com/dotnet/sdk:10.0-alpine`.

To reproduce on any Linux host:

`docker compose up -d --build`

The issue was introduced in `2839459fb3` (#8198). It is currently present on `develop`, but has not been included in a release tag yet. The .NET workflow does not catch it because it runs on `windows-latest`, where file paths are case-insensitive.

The fix is a one line change in `src/NopCommerce.sln`: `ChatGPTShopping` → `ChatGptShopping`. The project name and `.csproj` filename on the same line already use `ChatGpt`, and this is the only one of the solution’s 39 project references whose path does not match the actual directory name on disk.

I have the fix verified locally and can open a PR.

AntonioVCruz:
Fixes #8319.

`src/NopCommerce.sln` referenced the ChatGPT Shopping plugin directory as `Plugins\Nop.Plugin.Feed.ChatGPTShopping\`, but the actual directory on disk is `Nop.Plugin.Feed.ChatGptShopping`. On case-sensitive filesystems, MSBuild cannot resolve the project path, causing `dotnet build NopCommerce.sln` to fail with `MSB3202`. This also breaks the in-repo `Dockerfile` and, as a result, `docker compose up --build`.

Verified on `develop` at `60da289f2b`: `docker compose up -d --build` fails with `MSB3202` before this change and completes successfully afterward. The `.NET` workflow is unaffected because it runs on `windows-latest`, where paths are case-insensitive.

This is a one-line fix. The project name and `.csproj` filename on the same line already use `ChatGpt`, and this was the only one of the solution’s 39 project references whose path did not match the directory name on disk.


skoshelev:
Hi, @AntonioVCruz!

Thank you for your help

